PR #50, opened 2025-08-06
Implementing valkey cluster commands for ruby

pratheep-kumar:


lib/valkey.rb
```diff
@@ -182,15 +182,32 @@ def convert_response(res, &block)
...
 
     route_buf = FFI::MemoryPointer.from_string(route)
 
-    arg_ptrs, arg_lens = build_command_args(command_args)
+    # Handle empty command_args case
+    if command_args.empty?
+      arg_ptrs = FFI::MemoryPointer.new(:pointer, 1)
+      arg_lens = FFI::MemoryPointer.new(:ulong, 1)
+      arg_ptrs.put_pointer(0, FFI::MemoryPointer.new(1))
+      arg_lens.put_ulong(0, 0)
```
mohsen-alizadeh, line 204: What this section fixes? the `build_command_args` has to make proper arg_ptrs and arg_lens already

lib/valkey.rb
```diff
@@ -182,15 +182,32 @@ def convert_response(res, &block)
...
 
   def send_command(command_type, command_args = [], &block)
+    # Validate connection
+    if @connection.nil?
+      raise "Connection is nil"
+    elsif @connection.null?
+      raise "Connection pointer is null"
+    elsif @connection.address.zero?
+      raise "Connection address is 0"
+    end
```
mohsen-alizadeh, line 192: This block runs on eacy command, so when/how those methods can turn to invalid after creating the connection?

PR #87, opened 2026-03-19
Removing protobuf layer in connection creation

pratheep-kumar:
# Removing protobuf layer in connection creation

## Summary

Removes the protobuf dependency from the connection creation path and replaces it with
URI-based client creation using the `create_client_from_uri` FFI function, introduced
in [valkey-glide#5620](https://github.com/valkey-io/valkey-glide/pull/5620).

### Core connection (`lib/valkey.rb`)
- `initialize` now builds a `redis://` / `rediss://` URI from connection options and
  passes it to `create_client_from_uri` along with an extra JSON options hash
- Removed the `addresses` key from the JSON options — it is not a valid key for
  `create_client_from_uri` (that field belongs to the protobuf `ConnectionRequest`).
  For cluster mode, GLIDE discovers the full topology automatically via
  `CLUSTER SLOTS`/`CLUSTER NODES` from the seed URI with `cluster_mode_enabled: true`

### Protobuf removed
- Deleted `lib/valkey/protobuf/` (`connection_request_pb.rb`, `command_request_pb.rb`,
  `response_pb.rb`)
- Removed `google-protobuf` gem dependency from `valkey.gemspec`

### Bindings (`lib/valkey/bindings.rb`)
- Added `create_client_from_uri` FFI function signature

### Scripting (`lib/valkey/commands/scripting_commands.rb`)
- Fixed null pointer issue in `invoke_script`

### Updated libglide
- `libglide_ffi.dylib` and `libglide_ffi.so` updated to the version that ships
  `create_client_from_uri`

### Test coverage
- Added `test/valkey/uri_connection_test.rb` covering standalone connections, cluster
  mode, TLS/SSL, authentication, timeouts, reconnect strategy, URI parsing, and
  protocol options

## How cluster mode works now

```
redis://127.0.0.1:7000  +  { "cluster_mode_enabled": true, ... }
         |
         v
GLIDE seed connection
         |
         v
CLUSTER SLOTS / CLUSTER NODES  ->  full topology discovered automatically
```

## Testing

```sh
bundle exec rake test:valkey    # standalone suite
bundle exec rake test:cluster   # cluster suite (requires cluster on ports 7000-7005)
bundle exec rubocop             # no offenses in lib/ or test/
```

PR #88, opened 2026-03-27
Reorder hash command args

yoka:
Using protobuf gem version 3 became painful at least on arch linux I'm running so decided to bump the version to newest one. Almost butter smooth but some tests failed and I assume they were broken already before the protobuf update. Now they're fixd.
Thanks!

Here's a summary:

Move FIELDS and field list to the end of hash command arguments so optional flags are placed between key and fields. Adjust hsetex argument order to include EX and FIELDS count. Update tests to use httl(...).first to match reply shape. Bump google-protobuf dependency to ~> 4.31


Sasidharan3094:
@yoka We are working on to remove the protobuf layer completely and it is pending on the [rust core ](https://github.com/valkey-io/valkey-glide/pull/5620) and once that is merged, this [PR](https://github.com/valkey-io/valkey-glide-ruby/pull/87) will be merged. Please hold on.

Sasidharan3094:
@yoka Can we have reorder hash command args alone in this PR?

Sasidharan3094 (approved):
LGTM! Verified the argument ordering against the official Valkey docs:

HEXPIRE / HPEXPIRE / HEXPIREAT / HPEXPIREAT — confirmed the spec requires [NX|XX|GT|LT] before FIELDS, the old code had them appended after which would cause a syntax error from the server
HGETEX — same pattern, expiry option must precede FIELDS
HSETEX — old code was missing both the EX keyword and FIELDS entirely, new arg order matches the spec
.first on httl calls in tests is correct since HTTL always returns an array
Good catch, the fixes are accurate.

pratheep-kumar (approved):
LGTM

Sasidharan3094:
Since it is failing due to commit having veririfed signatures, closing this PR and taking the changes of hash command args reoder in this PR - https://github.com/valkey-io/valkey-glide-ruby/pull/91

PR #91, opened 2026-05-19
Fixing ruby standalone tests and hash command reorder

Sasidharan3094:
# Fix CI test failures and warnings

## Summary

Fixes 2 test failures and 2 categories of warnings surfaced in the GitHub Actions test run.

---

### `test/lint/generic_commands.rb`

**`test_move`** — Previously used `r.select` to switch databases to verify the move, which is unreliable with GLIDE's connection-pooled client. Replaced with two dedicated clients, each pinned to their target database via the `db:` connection option.

```ruby
# Before: used r.select 14 / r.select 15
# After:
db14 = _new_client(db: 14)
db15 = _new_client(db: 15)
```

**`test_copy`** — Same issue for the cross-database path (`copy` with `db:` option). The same-database assertions are unchanged. The cross-database section now uses dedicated clients per database instead of `with_db` / `r.select`.

---

### `test/valkey/eval_evalsha_property_test.rb`

Two property tests (`test_script_execution_consistency` and `test_type_conversion_correctness`) used `assert_equal eval_result, evalsha_result` where `eval_result` can be `nil` at runtime (e.g. when the sampled script is `"return nil"` or `"return false"`). Minitest warns against `assert_equal nil, x` and will make it a hard failure in Minitest 6.

Replaced with a nil-aware check:

```ruby
# Before:
assert_equal eval_result, evalsha_result

# After:
if eval_result.nil?
  assert_nil evalsha_result
else
  assert_equal eval_result, evalsha_result
end
```

---

### `test/lint/json_commands.rb`

The `setup` method used a `@json_not_available` flag to skip JSON tests when the module couldn't be loaded. Because the module always loads successfully in CI, the `rescue` blocks never ran and the flag was never assigned — causing Ruby to emit warnings on every test method that read it.

Since the JSON module is always present in CI (copied from `.github/files/` and mounted into the Valkey container), the flag and all associated `skip` guards have been removed. The `setup` method now loads the module directly and will fail loudly if it is unavailable, which is the correct behaviour for a required dependency.

---

### Problem

`build_command_args` allocates `FFI::MemoryPointer` buffers for each command argument
and stores them in a local `buffers` array to prevent garbage collection. However, only
`arg_ptrs` and `arg_lens` were returned — `buffers` went out of scope immediately after
the method returned, making those pointers eligible for GC before the FFI layer finished
reading them.

Under GC pressure (e.g. property-based tests running 100 iterations in a loop), the
memory could be reclaimed and reused mid-call, causing arguments to be read back as
garbage bytes instead of the original strings.

This was observed as a flaky failure in `test_parameter_round_trip_preservation` (2 out
of 30 CI runs), where ARGV values came back as raw pointer-sized byte sequences:

Expected: ["arg0_12"] Actual: ["A\xA3\x01Y\x05\x00\x00"]

### Fix

Return `buffers` as a third value from `build_command_args` and hold a reference to it
in both `send_command` and `send_batch_commands` until after `Bindings.command` returns.

```ruby
# before
arg_ptrs, arg_lens = build_command_args(command_args)

# after
arg_ptrs, arg_lens, _buffers = build_command_args(command_args)
#                   ^^^^^^^^ keeps FFI::MemoryPointer objects alive until FFI call completes
No functional change — commands sent to Valkey are identical. This only affects memory lifetime of the argument buffers.

### DIFF
--- a/lib/valkey.rb
+++ b/lib/valkey.rb
@@ -55,7 +55,7 @@ class Valkey
     buffers = [] # Keep references to prevent GC
 
     commands.each do |command_type, command_args, block|
-      arg_ptrs, arg_lens = build_command_args(command_args)
+      arg_ptrs, arg_lens, arg_bufs = build_command_args(command_args)
 
       cmd = Bindings::CmdInfo.new
       cmd[:request_type] = command_type
@@ -65,7 +65,7 @@ class Valkey
 
       cmds << cmd
       blocks << block
-      buffers << [arg_ptrs, arg_lens] # Prevent GC
+      buffers << [arg_ptrs, arg_lens, arg_bufs] # Prevent GC
     end
 
     # Create array of pointers to CmdInfo structs
@@ -131,7 +131,7 @@ class Valkey
   def build_command_args(command_args)
     # For empty arrays, pass NULL pointers as per Rust FFI contract
     # This matches Go's approach which successfully uses nil pointers
-    return [FFI::Pointer::NULL, FFI::Pointer::NULL] if command_args.empty?
+    return [FFI::Pointer::NULL, FFI::Pointer::NULL, []] if command_args.empty?
 
     arg_ptrs = FFI::MemoryPointer.new(:pointer, command_args.size)
     arg_lens = FFI::MemoryPointer.new(:ulong, command_args.size)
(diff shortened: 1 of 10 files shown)


pratheep-kumar (approved):
LGTM

PR #95, opened 2026-05-27
Modify Repo to use submodule + Add CD pipeline

alexr-bq:
This PR:
- Replaces hardcoded build of native library with a build of glide core from submodule
- Modifies CI job to support this
- Adds CD job which creates, deploys and tests a gem


Sasidharan3094 (approved):
LGTM, pls resolve the conflict for readme.

DEVELOPER.md
```diff
@@ -102,36 +113,57 @@ bundle exec bin/console
...
+- **Rust toolchain**: Install via [rustup](https://rustup.rs/)
+- **Protobuf compiler**: `protoc` (required for glide-core)
 
-Copy the built library into the Ruby gem:
+### Build with Rake
 
 ```bash
-# Linux
-cp target/release/libglide_ffi.so /path/to/valkey-glide-ruby/lib/valkey/
+# Build the native FFI library (release mode)
```
jduo, line 126: Is there a way to automatically build the native lib then build the packaged gem? Let's document if so.

valkey.gemspec
```diff
@@ -19,13 +19,14 @@ Gem::Specification.new do |spec|
 
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
+  # Native libraries (lib/valkey/native/**) are added separately since they're built during CD.
```
jduo: What happens if a dev wants to build the gem without deploying CD? What do they do.

DEVELOPER.md
jduo: This isn't accurate anymore right? Please update.

DEVELOPER.md
alexr-bq: Added instructions

valkey.gemspec
alexr-bq: Updated comment and added instructions in DEVELOPER.md

DEVELOPER.md
alexr-bq: rebuild -> build

jduo (approved):

PR #111, opened 2026-06-24
Add Full Matrix Tests

alexr-bq:
## Summary

Adds full matrix test setup, similar to other GLIDE projects.

Some matrix elements are flagged as "run-always", which will run on each PR, others will be run in nightly full matrix test run.

For now, only x86 and ARM linux platforms are supported. Mac support will be worked on next and come in a seperate PR.

Also added some additional version guards for tests found in testing under FMT.

### DIFF
--- a/.github/json_matrices/build-matrix.json
+++ b/.github/json_matrices/build-matrix.json
@@ -14,6 +14,7 @@
         "RUNNER": "ubuntu-24.04-arm",
         "ARCH": "arm64",
         "TARGET": "aarch64-unknown-linux-gnu",
+        "run": "always",
         "languages": ["ruby"]
     }
 ]
--- a/.github/json_matrices/build-matrix.json
+++ b/.github/json_matrices/build-matrix.json
@@ -5,7 +5,6 @@
         "RUNNER": "ubuntu-24.04",
         "ARCH": "x64",
         "TARGET": "x86_64-unknown-linux-gnu",
-        "PACKAGE_MANAGERS": ["rubygems"],
         "run": "always",
         "languages": ["ruby"]
     },
@@ -15,7 +14,6 @@
         "RUNNER": "ubuntu-24.04-arm",
         "ARCH": "arm64",
         "TARGET": "aarch64-unknown-linux-gnu",
-        "PACKAGE_MANAGERS": ["rubygems"],
         "languages": ["ruby"]
     }
 ]
--- a/.github/json_matrices/build-matrix.json
+++ b/.github/json_matrices/build-matrix.json
@@ -5,6 +5,7 @@
         "RUNNER": "ubuntu-latest",
         "ARCH": "x64",
         "TARGET": "x86_64-unknown-linux-gnu",
+        "languages": ["ruby"],
         "run": "always"
     },
     {
@@ -13,6 +14,7 @@
         "RUNNER": "ubuntu-24.04-arm",
         "ARCH": "arm64",
         "TARGET": "aarch64-unknown-linux-gnu",
+        "languages": ["ruby"],
         "run": "always"
     }
 ]
--- a/.github/json_matrices/build-matrix.json
+++ b/.github/json_matrices/build-matrix.json
@@ -2,11 +2,12 @@
     {
         "OS": "ubuntu",
         "NAMED_OS": "linux",
-        "RUNNER": "ubuntu-latest",
+        "RUNNER": "ubuntu-24.04",
         "ARCH": "x64",
         "TARGET": "x86_64-unknown-linux-gnu",
-        "languages": ["ruby"],
-        "run": "always"
+        "PACKAGE_MANAGERS": ["rubygems"],
+        "run": "always",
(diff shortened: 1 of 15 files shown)


.github/workflows/CI.yml
```diff
@@ -70,223 +149,203 @@ jobs:
...
           for i in {1..30}; do
-            if docker exec valkey-test valkey-cli ping > /dev/null 2>&1; then
-              echo "Valkey is ready!"
+            if redis-cli -p 6379 ping > /dev/null 2>&1 || valkey-cli -p 6379 ping > /dev/null 2>&1; then
+              echo "Standalone server is ready on port 6379!"
               break
             fi
-            echo "Waiting... ($i/30)"
             sleep 1
           done
```
Aryex: The main repo uses a python script to prepare the Valkey servers. Could we also do the same here?

.github/json_matrices/build-matrix.json
```diff
@@ -2,17 +2,20 @@
     {
         "OS": "ubuntu",
         "NAMED_OS": "linux",
-        "RUNNER": "ubuntu-latest",
+        "RUNNER": "ubuntu-24.04",
         "ARCH": "x64",
         "TARGET": "x86_64-unknown-linux-gnu",
-        "run": "always"
+        "PACKAGE_MANAGERS": ["rubygems"],
+        "run": "always",
```
Aryex, line 8: Is this always being used?

alexr-bq: Removed

.github/workflows/CI.yml
alexr-bq: Yes, we were using this for cluster but just modified to use for standalone

.github/json_matrices/build-matrix.json
```diff
@@ -2,17 +2,18 @@
...
+        "languages": ["ruby"]
     },
     {
         "OS": "ubuntu",
         "NAMED_OS": "linux",
         "RUNNER": "ubuntu-24.04-arm",
         "ARCH": "arm64",
         "TARGET": "aarch64-unknown-linux-gnu",
-        "run": "always"
+        "languages": ["ruby"]
```
xShinnRyuu, line 18: Are we missing a  `"run": "always",` here? Its not clear in your PR description what is suppose to run as part of CI for PRs or purely for the nightly run

.github/workflows/CI.yml
```diff
@@ -70,223 +152,191 @@ jobs:
...
-
-          echo "Verifying SSL cert files are accessible inside container..."
-          docker exec valkey-ssl ls -lh /ssl/ || echo "Cannot access /ssl/ directory"
-          docker exec valkey-ssl cat /etc/valkey-ssl.conf || echo "Cannot read config file"
-
-          echo "Waiting for SSL Valkey to be ready..."
           for i in {1..30}; do
-            if docker exec valkey-ssl valkey-cli --tls --insecure -p 6380 ping > /dev/null 2>&1; then
-              echo "SSL Valkey is ready!"
+            if docker exec valkey-ssl ${{ matrix.engine.type }}-cli --tls --insecure -p 6380 ping > /dev/null 2>&1; then
```
xShinnRyuu: Is it possible to create or use something similar to the [cluster_manager.py that the main glide repo](https://github.com/valkey-io/valkey-glide/blob/main/utils/cluster_manager.py) uses instead of having to depend on docker?

.github/json_matrices/build-matrix.json
xShinnRyuu: It does not look like this was removed

test/lint/pubsub_commands.rb
```diff
@@ -41,6 +41,7 @@ def test_pubsub_numsub_no_channels
     end
 
     def test_pubsub_shardchannels
+      omit_version("7.0")
```
xShinnRyuu, line 46: Can we add comments as to why these are omitted? I assume its becasue its not supported on redis/valkey 7.0.

Also from **.github/json_matrices/engine-matrix.json** it looks like we also run against Redis 6.2. So that should also be omitted

Please address this issue where applicable in this PR

.github/json_matrices/build-matrix.json
Aryex: It was removed. "always" wasn't being checked.

.github/workflows/CI.yml
alexr-bq: Was using for other cases but this one got left behind for TLS tests, fixed now.

.github/json_matrices/build-matrix.json
alexr-bq: Set this to run always

test/lint/pubsub_commands.rb
alexr-bq: Added comments. This does also omit redis 6.2

xShinnRyuu (approved):

Aryex (approved):

PR #112, opened 2026-06-25
test(ruby): privatize CLIENT TRACKING, CLIENT CACHING, and CLIENT TRACKINGINFO

### LINKED ISSUE #105: Enable/Improve client-side caching tests
## Description

Client-side caching tests are disabled. These should be enabled and fixed if needed.

Aryex:
## Overview

Valkey Client Side Caching requires implementation of GLIDE's dedicated CSC interface which has not been done. Public interfaces for `CLIENT TRACKING`, `CLIENT CACHING`, and `CLIENT TRACKINGINFO` commands has been made private and their tests are removed. 

## Other changes
- Fixes a bug in tracking argument construction surfaced by the new tests.
- Updated `reddis-rb` send command to only access public methods.
 
## Related Issue
Closes #105 

## Changes
- `connection_commands.rb`: privatized `client_tracking`, `client_caching`, `client_tracking_info`, and `client_getredir`
- `connection_commands.rb`: Removed tests related to the privatized CSC methods.
- `server_commands.rb`: Removed tests related to the privatized CSC methods.
- `connection_commands.rb`: fixed `build_client_tracking_args` to emit `PREFIX` before each prefix value (`PREFIX foo PREFIX bar`).
- `connection_commands.rb`: Updated send to use `public_send`, restricting it to public methods only.

### DIFF
--- a/test/lint/server_commands.rb
+++ b/test/lint/server_commands.rb
@@ -158,29 +158,6 @@ module Lint
       assert [0, 1].include?(result), "Expected unblock to return 0 or 1"
     end
 
-    def test_redis_rb_client_caching
-      # CLIENT CACHING requires shared tracking state, which is not reliable in
-      # cluster mode because TRACKING and CACHING may hit different nodes.
-      skip("CLIENT CACHING requires tracking state to be shared, not reliable in cluster mode") if cluster_mode?
-
-      # The redis-rb dispatcher (r.client(:caching, mode)) forwards to client_caching.
-      # CLIENT CACHING YES works with OPTIN mode.
-      r.client(:tracking, "ON", "OPTIN")
-      assert_equal "OK", r.client(:caching, "YES")
-      r.client(:tracking, "OFF")
-      # CLIENT CACHING NO requires OPTOUT mode.
-      r.client(:tracking, "ON", "OPTOUT")
-      assert_equal "OK", r.client(:caching, "NO")
-      r.client(:tracking, "OFF")
-    end
-
-    def test_redis_rb_client_tracking
-      # The redis-rb dispatcher (r.client(:tracking, status)) forwards to
-      # client_tracking, which requires a status argument.
-      assert_equal "OK", r.client(:tracking, "ON")
-      assert_equal "OK", r.client(:tracking, "OFF")
-    end
-
     def test_client_reply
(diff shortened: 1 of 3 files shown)


test/lint/server_commands.rb
```diff
@@ -158,20 +158,27 @@ def test_client_unblock
...
+      r.client(:tracking, "OFF")
     end
 
-    def test_client_tracking
-      skip("CLIENT TRACKING command not implemented in backend yet")
-
-      # Assuming tracking is enabled by default, this should return true
-      response = r.client(:tracking)
-      assert_equal true, response
+    def test_redis_rb_client_tracking
```
affonsov: we don't want to expose client tracking through api, using like this wont work we need to use clientSideCaching config through the connection configuration fro client side caching work. It need the server assisted from the core to work.
Remove this from the API and implement the connection configuration

test/lint/server_commands.rb
```diff
@@ -158,20 +158,27 @@ def test_client_unblock
       assert [0, 1].include?(result), "Expected unblock to return 0 or 1"
     end
 
-    def test_client_caching
-      skip("CLIENT CACHING command not implemented in backend yet")
+    def test_redis_rb_client_caching
```
affonsov: client caching should not be exposed as well, because we are able to only expose one way to enable client tracking and that way doesn't require client tracking.

Please follow the same pattern from the main repo how the client side caching work

alexr-bq (approved):

test/lint/server_commands.rb
Aryex: CSC has not been implemented so I made the decision to privatize these commands API. Related tests are removed.

addressed in [23d71ed](https://github.com/valkey-io/valkey-glide-ruby/pull/112/commits/23d71eda15ec00edea83a5421c7133252d6cbf8b)

test/lint/server_commands.rb
Aryex: CSC has not been implemented so I made the decision to privatize these commands API. Related tests are removed.

addressed in [23d71ed](https://github.com/valkey-io/valkey-glide-ruby/pull/112/commits/23d71eda15ec00edea83a5421c7133252d6cbf8b)

affonsov:
LGTM

PR #131, opened 2026-07-07
feat: Add missing core options

### LINKED ISSUE #127: Connection config parity: read_from, client_az, inflight_requests_limit, lazy_connect, periodic_checks
## Overview

GLIDE's native core already supports several connection settings — which replica to read from,
availability-zone affinity, in-flight request limits, delayed connecting, and periodic
cluster-topology health checks — but `Valkey.new` doesn't expose any of them.

## Description

All 5 are already-supported `extra_options_json` keys on the existing `create_client_from_uri`
path — no new FFI surface, no protobuf. Only `periodic_checks` is cluster-only in effect (its
topology-refresh behavior only applies inside cluster mode); it's still accepted, harmlessly, on
standalone. The other 4 apply to both modes.

## Proposal

```ruby
client = Valkey.new(
  host: "localhost",
  port: 6379,
  read_from: "PreferReplica",
  client_az: "us-west-2a",
  inflight_requests_limit: 1000,
  lazy_connect: true,
  periodic_checks: { manual_interval: { duration_in_sec: 30 } }
)
```

Add each as a top-level `Valkey.new` keyword option, validated in Ruby before being placed on
`json_options` — matching the existing pattern for `protocol`/`client_name`. Reject unknown
`read_from` strings, negative `inflight_requests_limit`, etc. in Ruby rather than letting the FFI
layer's JSON parser reject them with a less-clear error.

## Other

Split from `pubsub_subscriptions` and `iam_credentials` (different test shapes) — tracked as
separate issues.

Aryex:
## Overview

Adds 5 new `Valkey.new` constructor options — `read_from`, `client_az`,
`inflight_requests_limit`, `lazy_connect`, `periodic_checks` — as passthrough
`extra_options_json` keys on the existing `create_client_from_uri` FFI path.
No new FFI surface, no protobuf changes. This is PR 1 of 3 in the
connection-config-parity plan (`docs/cluster-connection-config-parity-implementation-plan.md`);
`pubsub_subscriptions` and `iam_credentials` are split into follow-up PRs.

Each option is validated in Ruby before serialization (raising `ArgumentError`
on invalid input), matching the existing pattern already used for
`protocol:`/`client_name:`/`connection_retry_strategy:` in `lib/valkey.rb`.

## Related Issue

Closes #127


lib/valkey.rb
```diff
@@ -128,6 +128,22 @@ def send_batch_commands(commands, exception: true)
     results
   end
 
+  # Builds the `periodic_checks` extra_options_json value from the `periodic_checks:`
+  # constructor option. Accepts either `{ manual_interval: { duration_in_sec: N } }` or
+  # `{ disabled: true/false }` (symbol or string keys), converting Ruby symbol keys to
+  # the string keys GLIDE's native core expects.
+  def build_periodic_checks(periodic_checks)
```
prateek-kumar-improving: Should this method be private? Will this be called by a user?
Please check this for other methods in the class like `build_command_args` and `convert_response`.

lib/valkey.rb
```diff
@@ -128,6 +128,22 @@ def send_batch_commands(commands, exception: true)
...
+    if periodic_checks.key?(:disabled) || periodic_checks.key?("disabled")
```
prateek-kumar-improving: If `periodic_checks` is passed as `empty {}` this would throw NoMethodError instead of ArgumentError.
What is expected here? It would be best to confirm from other clients and see if the error is thrown at the FFI layer or the client layer?

lib/valkey.rb
```diff
@@ -414,6 +430,41 @@ def initialize(options = {})
     # Client name (user-configurable)
     json_options["client_name"] = options[:client_name] if options[:client_name]
 
+    # Read routing preference (which node(s) to read from). Ruby-friendly symbols are mapped to the canonical GLIDE
+    # strings; anything else (including typos) is forwarded as-is and rejected by the
+    # core with its own error, rather than duplicating that validation here.
+    if options.key?(:read_from)
```
prateek-kumar-improving:
```
READ_FROM_MAP = {
  primary: "Primary",
  prefer_replica: "PreferReplica",
  lowest_latency: "LowestLatency",
  az_affinity: "AZAffinity",
  az_affinity_replicas_and_primary: "AZAffinityReplicasAndPrimary"
}.freeze

Define this map at the class level, then Valkey::READ_FROM_MAP can be accessed in the tests.
```

```
# Then in initialize:
if options.key?(:read_from)
  json_options["read_from"] = READ_FROM_MAP[options[:read_from]] || options[:read_from].to_s
end
```
Would it be better to have a map instead of switch case? A map can be

test/valkey/connection_config_test.rb
```diff
@@ -0,0 +1,232 @@
...
+      assert_equal "Bogus", json_options["read_from"]
+    end
+
+    def test_read_from_omitted_when_not_provided
+      json_options = captured_json_options
+      refute json_options.key?("read_from")
+    end
+
+    # ====================
+    # client_az
```
prateek-kumar-improving: The test name already has the name of the field, we can remove the field name comment for all tests if it does not add any value.

lib/valkey.rb
Aryex: Addressed in [10ccdc4](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/10ccdc49e3c3f744b174cd91b876164d8a923dbb)

Created https://github.com/valkey-io/valkey-glide-ruby/issues/146 to later privatize `send_command()`

test/valkey/connection_config_test.rb
Aryex: Removed in [515a32e](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/515a32eff5fd8643385d030a21074c50d379fbdf)

lib/valkey.rb
Aryex: Updated in [09a59e6](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/09a59e651a244542c53ca8a265c2b0abb7a151c4)

I decided to use an Enum instead to match the patterns in other clients.

lib/valkey.rb
Aryex: updated in [85b7ea0](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/85b7ea0ff74fa01effa51324c23e794df735d521)

lib/valkey.rb
```diff
@@ -31,399 +40,133 @@ def pipelined(exception: true)
...
+    az_affinity_values = %w[AZAffinity AZAffinityReplicasAndPrimary].freeze
+    if az_affinity_values.include?(json_options["read_from"]) && !options[:client_az]
+      raise ArgumentError, "client_az must be set when read_from is AZAffinity or AZAffinityReplicasAndPrimary"
+    end
 
-        Array.new(count) do |i|
-          item = Bindings::CommandResponse.new(ptr + (i * Bindings::CommandResponse.size))
+    if options.key?(:inflight_requests_limit)
+      json_options["inflight_requests_limit"] = options[:inflight_requests_limit]
+    end
```
xShinnRyuu, line 149: I see `timeout` and `connect_timeout` being type validated above, we should do the same for all options to be consistent

lib/valkey.rb
```diff
@@ -31,399 +40,133 @@ def pipelined(exception: true)
...
 
-          map_key = convert_response.call(Bindings::CommandResponse.new(item[:map_key]))
-          map_value = convert_response.call(Bindings::CommandResponse.new(item[:map_value]))
+    json_options["lazy_connect"] = true if options[:lazy_connect]
```
xShinnRyuu: When `lazy_connect` is false it gets ignored here. Do we have referencing documentation that lazy_connect is default to false?

lib/valkey.rb
```diff
@@ -31,399 +40,133 @@ def pipelined(exception: true)
...
-          convert_response.call(item)
-        end
-      when ResponseType::MAP
-        return nil if response_item[:array_value].null?
+    json_options["client_az"] = options[:client_az] if options[:client_az]
 
-        ptr = response_item[:array_value]
-        count = response_item[:array_value_len].to_i
-        map = {}
+    az_affinity_values = %w[AZAffinity AZAffinityReplicasAndPrimary].freeze
```
xShinnRyuu: This `az_affinity_values` array is created on every single call. We can probably abstract it to be a class-level constant similar to `READ_FROM_MAP`

lib/valkey.rb
```diff
@@ -21,6 +22,14 @@ class Valkey
   include Commands
   include PubSubCallback
 
+  # read_from values map.
+  READ_FROM_MAP = {
```
xShinnRyuu: This is an exact duplicate of `module ReadFrom` in `lib/valkey/read_from.rb`. Instead of duplicating it here and having to update both whenever a change is made, we should make `READ_FROM_MAP` reference the `ReadFrom` constants. 

```
 require "valkey/read_from"

  class Valkey
    READ_FROM_MAP = {
      primary: ReadFrom::PRIMARY,
      prefer_replica: ReadFrom::PREFER_REPLICA,
      az_affinity: ReadFrom::AZ_AFFINITY,
      az_affinity_replicas_and_primary: ReadFrom::AZ_AFFINITY_REPLICAS_AND_PRIMARY
    }.freeze

    # And solve the first problem too:
    AZ_AFFINITY_VALUES = [
      ReadFrom::AZ_AFFINITY,
      ReadFrom::AZ_AFFINITY_REPLICAS_AND_PRIMARY
    ].freeze
  end
```

This also improves performance by reducing the allocation needed to only when the class loads

xShinnRyuu (changes requested):

lib/valkey.rb
Aryex: Updated in [282947f](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/282947ff502929be50d24b98fb1fa5b00d4e264c)

`read_from` can be safely pass-through because the core handles type validation there.

lib/valkey.rb
Aryex: Updated in [5d3b69f](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/5d3b69f37e0188f6a9f3ef53b8c6c5c8ae51510c)

I used if cases for readibility. However this type of validation is probably better handled at the core. Made a follow-up ticket #148

lib/valkey.rb
Aryex: updated in [660b9fb](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/660b9fbccd88763c178708f2ce0c08ae8b58d4cd)

The core handle type validation for configuration options so we can safely pass through this value.

lib/valkey.rb
Aryex: These values can be safely pass-through as well. However I will defer changing this to #148

lib/valkey.rb
Aryex: Additional fixes in [7b09cc5](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/7b09cc5c02cf94d2b4c937dfc9395d32fe25beaa)

lib/valkey.rb
Aryex: Additional fixes in [7b09cc5](https://github.com/valkey-io/valkey-glide-ruby/pull/131/commits/7b09cc5c02cf94d2b4c937dfc9395d32fe25beaa)

xShinnRyuu (approved):

prateek-kumar-improving (approved):

PR #132, opened 2026-07-07
Implement generic command execution: call(), call_v()

### LINKED ISSUE #126: Implement generic command dispatch: call / call_v
## Overview

GLIDE allows for custom command execution through its `customCommand` API. redis-rb offers the
same capability through `#call`/`#call_v`. This adds that redis-rb interface on top of GLIDE's API.

## Proposal

Add `call(*args, **kwargs)` and `call_v(args)`, matching `redis-client`'s documented behavior
(flattening/coercion), not a raw passthrough:

```ruby
client.call("SET", "mykey", "value")
# => "OK"

client.call_v(["MGET"] + keys)

client.call("SET", "mykey", 42)                  # Integers/Floats auto-stringify
client.call("LPUSH", "list", [1, 2, 3])           # Arrays flatten
client.call("HMSET", "hash", { "foo" => "1" })    # Hashes flatten to alternating key/value
client.call("SET", "k", "v", nx: true, ex: 60)    # kwargs become flags: "nx" "ex" "60"
client.call("SET", "k", "v", nx: false, ex: nil)  # falsy/nil flags dropped, not stringified
```

Implement the flattening/coercion directly in `call`/`call_v`, not in the shared
`build_command_args` path every other command method uses (those already receive pre-shaped args
and don't need this). Return value is the raw reply, no type casting — matching `redis-client`'s
documented behavior.

Aryex:
## Overview

Adds `call(*args, **kwargs)` and `call_v(args)` as the generic command dispatch, matching `redis-client`'s `#call`/`#call_v`. 

Sends any command directly through the previously-unused `RequestType::CUSTOM_COMMAND`, with redis-client-compatible argument coercion (Integer/Float stringify, Array flatten, Hash flatten to alternating key/value, kwargs→trailing flags) and a raw, untyped reply.

## Related Issue
Closes #126 

## Changes

* `lib/valkey/commands/generic_commands.rb`: Add public `call`/`call_v` dispatching via `RequestType::CUSTOM_COMMAND`, plus private `flatten_call_args` (recursive Array/Hash flattening, Integer/Float stringify) and `call_flags` (kwargs → upcased trailing flags, dropping falsy/nil) helpers.
* `README.md`: Document `call`/`call_v` as the escape hatch for commands without a dedicated method, including the coercion/flag examples.

## Testing

Integration and unit tests added

### DIFF
--- a/lib/valkey/commands/generic_commands.rb
+++ b/lib/valkey/commands/generic_commands.rb
@@ -571,17 +571,25 @@ class Valkey
 
       # Flattens Arrays and Hashes (to alternating key/value) and stringifies
       # Integers/Floats, matching `redis-client`'s documented `#call`/`#call_v` behavior.
-      def flatten_call_args(args)
-        args.flat_map do |arg|
+      #
+      # Appends into a single accumulator instead of flat_map + recursive return
+      # values, which would allocate a new intermediate array at every nesting
+      # level for large/deeply-nested inputs.
+      def flatten_call_args(args, acc = [])
+        args.each do |arg|
           case arg
           when Array
-            flatten_call_args(arg)
+            flatten_call_args(arg, acc)
           when Hash
-            flatten_call_args(arg.to_a.flatten(1))
+            arg.each do |key, value|
+              acc << key.to_s
+              flatten_call_args([value], acc)
+            end
           else
-            arg.to_s
+            acc << arg.to_s
           end
         end
+        acc
       end
 
       # Converts `call`'s **kwargs into trailing command flags: a truthy value emits
--- a/lib/valkey/commands/generic_commands.rb
+++ b/lib/valkey/commands/generic_commands.rb
@@ -541,7 +541,7 @@ class Valkey
       #   valkey.call("SET", "k", "v", nx: true, ex: 60)
       #     # equivalent to call("SET", "k", "v", "NX", "EX", "60")
       #
-      # @param [Array<String, Integer, Float, Array, Hash>] args command name and its arguments
+      # @param [Array<String, Integer, Float, Array, Hash>] argv command name and its arguments
       # @param [Hash] kwargs trailing command flags; truthy values emit the upcased flag name,
       #   non-boolean values also emit the stringified value; falsy/nil values are dropped
       # @return [Object] the raw reply, with no type-casting based on the command name
--- a/lib/valkey/commands/generic_commands.rb
+++ b/lib/valkey/commands/generic_commands.rb
@@ -559,12 +559,12 @@ class Valkey
       # @example
       #   valkey.call_v(["MGET"] + keys)
       #
-      # @param [Array<String, Integer, Float, Array, Hash>] args command name and its arguments
+      # @param [Array<String, Integer, Float, Array, Hash>] argv command name and its arguments
       # @return [Object] the raw reply, with no type-casting based on the command name
       #
       # @see https://valkey.io/commands/
       def call_v(argv)
-        send_command(RequestType::CUSTOM_COMMAND, flatten_call_args(args))
+        send_command(RequestType::CUSTOM_COMMAND, flatten_call_args(argv))
       end
 
       private
--- a/lib/valkey/commands/generic_commands.rb
+++ b/lib/valkey/commands/generic_commands.rb
@@ -600,12 +600,16 @@ class Valkey
       # Converts `call`'s **kwargs into trailing command flags: a truthy value emits
       # the upcased flag name, and a non-boolean truthy value also emits the
       # stringified value. Falsy/nil values are dropped entirely, not stringified.
+      #
+      # @example
+      #   call_flags(nx: true, ex: 60, cond: false, ttl: nil)
+      #   # => ["NX", "EX", "60"]
       def call_flags(kwargs)
         kwargs.each_with_object([]) do |(name, value), flags|
           next unless value
 
           flags << name.to_s.upcase
-          flags << value.to_s unless [true, false].include?(value)
+          flags << value.to_s unless value == true
         end
       end
     end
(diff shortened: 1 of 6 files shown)


lib/valkey/commands/generic_commands.rb
```diff
@@ -520,6 +520,81 @@ def _scan(command, cursor, args, match: nil, count: nil, type: nil, &block)
...
+          end
+        end
+      end
+
+      # Converts `call`'s **kwargs into trailing command flags: a truthy value emits
+      # the upcased flag name, and a non-boolean truthy value also emits the
+      # stringified value. Falsy/nil values are dropped entirely, not stringified.
+      def call_flags(kwargs)
+        kwargs.each_with_object([]) do |(name, value), flags|
+          next unless value
```
prateek-kumar-improving, line 609: Code after this will only run if value is true. So, the false in the check with [true, false] will be unreachable.

lib/valkey/commands/generic_commands.rb
```diff
@@ -520,6 +520,81 @@ def _scan(command, cursor, args, match: nil, count: nil, type: nil, &block)
...
+
+      private
+
+      # Flattens Arrays and Hashes (to alternating key/value) and stringifies
+      # Integers/Floats, matching `redis-client`'s documented `#call`/`#call_v` behavior.
+      def flatten_call_args(args)
+        args.flat_map do |arg|
+          case arg
+          when Array
+            flatten_call_args(arg)
```
prateek-kumar-improving: For large inputs flatten_call_args will be called multiple times, there for flat_map + recursion creates new arrays at every level. This might need to be refactored.

Aryex: Addressed in [e6298a9](https://github.com/valkey-io/valkey-glide-ruby/pull/132/commits/e6298a93d2bf88c67a58317701bbc3acbc845a84)

Updated to using a queue implementation instead.

lib/valkey/commands/generic_commands.rb
Aryex: Fixed in [5649c1d](https://github.com/valkey-io/valkey-glide-ruby/pull/132/commits/5649c1d5118be190cdb6e5404ecc4e3ead2e67bf)

lib/valkey/commands/generic_commands.rb
```diff
@@ -520,6 +520,98 @@ def _scan(command, cursor, args, match: nil, count: nil, type: nil, &block)
...
+      #     # equivalent to call("SET", "k", "v", "NX", "EX", "60")
+      #
+      # @param [Array<String, Integer, Float, Array, Hash>] args command name and its arguments
+      # @param [Hash] kwargs trailing command flags; truthy values emit the upcased flag name,
+      #   non-boolean values also emit the stringified value; falsy/nil values are dropped
+      # @return [Object] the raw reply, with no type-casting based on the command name
+      #
+      # @see https://valkey.io/commands/
+      def call(*args, **kwargs)
+        send_command(RequestType::CUSTOM_COMMAND, flatten_call_args(args) + call_flags(kwargs))
```
xShinnRyuu: nit: optimally it is better to use `concat` instead of `+` for array concatenation to reduce allocation when `kwargs` is empty

test/valkey/call_test.rb
```diff
@@ -0,0 +1,145 @@
...
+      assert_raises(Valkey::CommandError) do
+        r.call_v(%w[NOTACOMMAND foo])
+      end
+    end
+
+    # should apply flattened Array/Hash args and kwargs-derived flags correctly
+    # end-to-end (server-visible effects: 3-element list, hash field set, TTL from
+    # a flag) in a single pass through the real dispatch path — the granular
+    # flattening/flag-construction cases themselves are unit-tested below.
+    def test_call_end_to_end_with_flattening_and_flags
```
xShinnRyuu, line 38: I only see Hash test for `call()` please add in `call_v()` ones for Hash with Array values

lib/valkey/commands/generic_commands.rb
```diff
@@ -520,6 +520,98 @@ def _scan(command, cursor, args, match: nil, count: nil, type: nil, &block)
...
+      # `#call_v` — no keyword flags.
+      #
+      # @example
+      #   valkey.call_v(["MGET"] + keys)
+      #
+      # @param [Array<String, Integer, Float, Array, Hash>] args command name and its arguments
+      # @return [Object] the raw reply, with no type-casting based on the command name
+      #
+      # @see https://valkey.io/commands/
+      def call_v(args)
```
xShinnRyuu: nit: instead of `args` would `argv` better?

test/valkey/call_test.rb
Aryex: Added extra tests in [e4f253d](https://github.com/valkey-io/valkey-glide-ruby/pull/132/commits/e4f253d67d35c8b00d19f5e3abee8a5a33103e39) and [cbfee10](https://github.com/valkey-io/valkey-glide-ruby/pull/132/commits/cbfee10f8ddbdc0bfec739ad0a032edc7d747143)

lib/valkey/commands/generic_commands.rb
Aryex: updated in [58c1a60](https://github.com/valkey-io/valkey-glide-ruby/pull/132/commits/58c1a60b42f72394136c2c09f75889e087c78b4a)

lib/valkey/commands/generic_commands.rb
Aryex: updated in [58c1a60](https://github.com/valkey-io/valkey-glide-ruby/pull/132/commits/58c1a60b42f72394136c2c09f75889e087c78b4a)

prateek-kumar-improving (approved):

xShinnRyuu (approved):

jamesx-improving (approved):

Issue #98, opened 2026-06-01
Add MUSL support


alexr-bq:
## Description

Modify CD pipeline to produce and test MUSL-compatible gem.

PR #143, opened 2026-07-08
feat(ruby): add Alpine Linux (musl libc) support for x86_64 and aarch64

prateek-kumar-improving:
### Summary

Add Alpine Linux (musl libc) support for the valkey-rb gem on both x86_64 and aarch64 architectures. This enables users running Docker containers based on Alpine to use the gem with 
prebuilt native libraries.

### Issue link

This Pull Request is linked to issue: https://github.com/valkey-io/valkey-glide-ruby/issues/98

### Features / Behaviour Changes

- The gem now automatically detects musl libc at runtime and loads the correct native library (*-unknown-linux-musl vs *-unknown-linux-gnu)
- Alpine Linux 3.18+ is now a supported platform for both x86_64 and aarch64
- CI validates the gem on Alpine Linux with full test coverage (standalone, TLS, cluster)
- CD builds and publishes musl-compatible native libraries alongside glibc ones

### Implementation

Runtime detection (lib/valkey/bindings.rb):
- Added musl_libc? class method with a 3-check cascade:
  1. RUBY_PLATFORM contains "musl"
  2. /etc/alpine-release file exists
  3. RbConfig::CONFIG["target_os"] contains "musl"
- platform_info now resolves {arch}-unknown-linux-musl when musl is detected

Build matrix (.github/json_matrices/build-matrix.json):
- Added x86_64-unknown-linux-musl and aarch64-unknown-linux-musl target entries

CI workflow (.github/workflows/CI.yml):
- Excluded musl targets from the regular glibc-based build/test jobs
- Added dedicated build-and-test-musl job running natively inside ruby:3.3-alpine container on x64
- Builds Rust FFI library with RUSTFLAGS="-C target-feature=-crt-static" (required for dynamic linking on musl)
- Runs full test suite including standalone, TLS, and cluster

CD workflow (.github/workflows/cd.yml):
- Added musl cross-compilation steps (musl-tools, rustup target, conditional --target flag)
- Adjusted artifact upload paths for target-specific Cargo output directory
- Added test-gem-musl job to verify the published gem loads and passes tests on Alpine

Test fix (test/valkey/uri_connection_test.rb):
- Skip test_ssl_with_unreadable_file when running as root (Alpine CI container), since root bypasses file permission checks

### Limitations

- ARM64 (aarch64) musl is built via cross-compilation in CD but cannot be integration-tested in CI containers due to a [GitHub Actions limitation](https://github.blog/changelog/2025-
09-19-deprecation-of-node-20-on-github-actions-runners/) where JavaScript-based actions don't work in Alpine containers on ARM64 runners. The x64 musl path has full integration 
testing.
- Only Alpine Linux 3.18+ is tested; other musl-based distros (e.g., Void Linux) should work but are untested.

### Testing

- Full standalone, TLS, and cluster test suites pass on Alpine Linux (x86_64)
- Semgrep security scanning passes (actions pinned to commit SHAs, no curl-pipe-shell)
- Existing glibc and macOS test suites unaffected

### Checklist

- [x] Linters have been run (bundle exec rubocop) and pass.
- [x] Destination branch is correct - main
- [x] Create merge commit if merging release branch into main, squash otherwise.
- [ ] Make sure to update the documentation in the [valkey-glide-docs](https://github.com/valkey-io/valkey-glide-docs) repository if necessary


lib/valkey/bindings.rb
```diff
@@ -40,7 +57,7 @@ def self.platform_info
                      when "darwin"
                        "#{arch}-apple-darwin"
                      when "linux"
-                       "#{arch}-unknown-linux-gnu"
+                       musl_libc? ? "#{arch}-unknown-linux-musl" : "#{arch}-unknown-linux-gnu"
```
Aryex, line 60: Has this been verified on an Alpine system? Note that current MUSL CI can't confirm this because glide core is build into `valkey-glide/ffi/target/release/libglide_ffi.so` so it never need to use `platform_info`.

Perhaps we should have an extra step to put it into `./native/#{platform_dir}/libglide_ffi.#{lib_ext}` to check that this works.

alexr-bq:
Have you ran the CD workflow (in non-publish / dry-run mode) and validated it?

lib/valkey/bindings.rb
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-ruby/pull/143/changes/7c5454bde6e26612fe91354736be8091b1bf3083. Added a verification step that:
1. Asserts platform_dir resolves to *-musl on Alpine
2. Copies the built .so into ./native/#{platform_dir}/libglide_ffi.so
3. Hides the submodule build output and confirms the gem loads from the native path 

This runs before the test suite starts.

prateek-kumar-improving:
> Have you ran the CD workflow (in non-publish / dry-run mode) and validated it?

The CD workflow worked: https://github.com/valkey-io/valkey-glide-ruby/actions/runs/29279904222

lib/valkey/bindings.rb
Aryex: I was imagining something closer to an E2E through ruby gem, or a manual verifcation. I don't see a lot of values in have this step on CI for this repeatedly.

alexr-bq (approved):

lib/valkey/bindings.rb
prateek-kumar-improving: Yeah, it is already tested in test-gem-musl. Removed.

Aryex (approved):

prateek-kumar-improving:
CD workflow works after the last code change: https://github.com/valkey-io/valkey-glide-ruby/actions/runs/29351834655

PR #161, opened 2026-07-15
Distributed Tracing (Parent Context Propagation) for valkey-glide-ruby

Sasidharan3094:
# Distributed Tracing (Parent Context Propagation) for valkey-glide-ruby

## Problem

`Valkey::OpenTelemetry` could create spans for commands, but every span was an
**independent root span**. `lib/valkey.rb` only ever called the parentless
`create_otel_span` / `create_batch_otel_span`, with `# TODO: add parent span
propagation` comments left unimplemented. As a result, Redis/Valkey command
spans never nested under an application's request trace — breaking the
"one connected trace per request" story that Ruby's
`opentelemetry-instrumentation-redis` provides for redis-rb today.

The Node.js GLIDE client already solves this: an app registers a
`parentSpanContextProvider` callback; before each command the client asks
for the current trace context and, if valid, creates the command span as a
**child** of that remote W3C context instead of a root span. The Rust FFI
core already shipped the native functions needed for this
(`create_otel_span_with_trace_context` / `create_batch_otel_span_with_trace_context`)
— they just weren't attached or called from Ruby. This change ports that
capability to Ruby.

### `lib/valkey/bindings.rb`
Attached the two existing Rust FFI functions:

```ruby
attach_function :create_otel_span_with_trace_context,
                 [:int, :string, :string, :uint8, :string], :uint64
attach_function :create_batch_otel_span_with_trace_context,
                 [:string, :string, :uint8, :string], :uint64
```

`:string` params marshal a Ruby `nil` to a `NULL` pointer automatically,
matching Rust's nullable `trace_state`.

### `lib/valkey/opentelemetry.rb`
New public API on `Valkey::OpenTelemetry`:

- **`set_parent_span_context_provider(callable = nil, &block)`** — registers a
  callable invoked before every command. Pass `nil` (no block) to clear it.
- **`parent_span_context`** — invokes the registered provider, validates the
  result, and returns a Hash or `nil`. Never raises: a provider that throws,
  or returns something malformed, is caught, logged with `warn`, and treated
  as "no parent context" (falls back to an independent span).
- **`init(..., parent_span_context_provider: nil)`** — optional kwarg,
  equivalent to calling `set_parent_span_context_provider` separately.

Validation rules (checked before the value ever reaches the FFI call):

| Field | Rule |
|---|---|
| `trace_id` | String matching `/\A[0-9a-f]{32}\z/` |
| `span_id` | String matching `/\A[0-9a-f]{16}\z/` |
| `trace_flags` | Integer in `0..255` |
| `tracestate` | `nil` or String |

The Rust core still validates and falls back to an independent span on any
malformed remote context, so Ruby-side validation only needs to catch
obviously-wrong shapes before making the FFI call.

### `lib/valkey.rb`
Both command-dispatch paths now check for a parent context before creating a
span, and use the trace-context-aware FFI function when one is present:

- `send_command` (single commands)
- `send_batch_commands` (non-transactional pipelines — one `Batch` span for
  the whole pipeline)

Everything downstream (the actual `Bindings.command`/`Bindings.batch` call,
the `ensure`-block `drop_otel_span`) is unchanged — only *how* `span_ptr` is
created changes.

**Known, intentionally out-of-scope limitation:** if a pipelined block
contains `MULTI`/`EXEC`/`DISCARD`, `send_batch_commands` bypasses
`Bindings.batch` entirely and replays each queued command through
`send_command` individually (a pre-existing stability workaround for native
crashes in transactional batches). That means a `MULTI...EXEC` transaction
still gets **N independent spans** instead of one transaction span — this
change doesn't alter that. It does still improve: each of those N spans now
correctly links to the app's parent trace instead of being N disconnected
root spans.

## Usage

```ruby
require "valkey"
require "opentelemetry/sdk" # if your app uses the real opentelemetry-ruby gem

Valkey::OpenTelemetry.init(
  traces: { endpoint: "http://localhost:4318/v1/traces", sample_percentage: 10 }
)

Valkey::OpenTelemetry.set_parent_span_context_provider do
  span = ::OpenTelemetry::Trace.current_span
  next nil unless span.context.valid?

  {
    trace_id: span.context.hex_trace_id,
    span_id: span.context.hex_span_id,
    trace_flags: span.context.trace_flags.sampled? ? 1 : 0,
    tracestate: span.context.tracestate.to_s
  }
end
```

**Note on the `OpenTelemetry` constant name:** `Valkey::OpenTelemetry` (this
gem) and the real `opentelemetry-ruby` gem's top-level `::OpenTelemetry` are
unrelated constants and don't collide. But a bare `OpenTelemetry::Trace`
reference written from code lexically nested inside `Valkey` would resolve to
the (nonexistent) `Valkey::OpenTelemetry::Trace` instead of the real gem —
always use the leading-`::` form, `::OpenTelemetry::Trace`, as above.

## Files touched

| File | Change |
|---|---|
| `lib/valkey/bindings.rb` | New FFI bindings |
| `lib/valkey/opentelemetry.rb` | Provider registration/validation API |
| `lib/valkey.rb` | Wire both command paths to use parent context when available |
| `test/valkey/opentelemetry_provider_test.rb` | New — 12 unit tests (validation, resilience) |
| `test/valkey/opentelemetry_test.rb` | +2 integration tests (end-to-end trace linkage), `read_spans` helper |
| `test/standalone/valkey_test.rb` | New `TestOpenTelemetryProvider` class |
| `README.md` | New "Distributed tracing" section |
| `examples/opentelemetry.rb` | Extended with a provider demo |
| `CHANGELOG.md` | Pending entry |
| `.rubocop.yml` | Added `opentelemetry.rb` to `Metrics/ModuleLength` exclusions |

## Verification performed

- `bundle exec rubocop` — clean on all changed files.
- All 12 new unit tests pass (no server required).
- Both new end-to-end integration tests pass against a real local server —
  confirmed via the exported span JSON (`span_exporter_file.rs` writes
  `trace_id`/`span_id`/`parent_span_id` per span) that the resulting span's
  `trace_id` and `parent_span_id` exactly match the injected context.
  ```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -66,4 +66,5 @@ Metrics/ModuleLength:
     - 'lib/valkey/utils.rb'
     - 'lib/valkey/commands/*.rb'
     - 'lib/valkey/bindings.rb'
+    - 'lib/valkey/opentelemetry.rb'
     - 'test/**/*.rb'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,3 +5,4 @@
 ### Changes
 
 * Ruby: Add Alpine Linux (musl libc) support for x86_64 and aarch64 — runtime detection of musl libc, CI/CD pipeline for native builds, and prebuilt `libglide_ffi.so` for musl targets ([#143](https://github.com/valkey-io/valkey-glide-ruby/pull/143))
+* Ruby: Add distributed tracing support — `Valkey::OpenTelemetry.set_parent_span_context_provider` (and `init(parent_span_context_provider:)`) let an app propagate its current W3C trace context into command/pipeline spans, so they become children of the app's trace instead of independent root spans, matching the Node.js client's `parentSpanContextProvider` behavior.
--- a/README.md
+++ b/README.md
@@ -238,6 +238,50 @@ client.set("key", "value")  # traced when sampling applies
 
 OpenTelemetry can only be initialized **once per process**. Spans are created in the FFI layer when sampling is enabled.
 
+### Distributed tracing (parent context propagation)
+
+By default, every command span is an independent root span — it doesn't nest under your
+application's request trace. To fix that, register a `parent_span_context_provider`: a
+callable invoked before every command that returns the current W3C trace context (or `nil`
+when there isn't one). When it returns a valid context, the command/pipeline span is created
+as a **child** of it instead of an independent span.
+
+```ruby
+require "valkey"
+require "opentelemetry/sdk" # if your app uses the real opentelemetry-ruby gem
+
+Valkey::OpenTelemetry.init(
+  traces: { endpoint: "http://localhost:4318/v1/traces", sample_percentage: 10 }
+)
+
+Valkey::OpenTelemetry.set_parent_span_context_provider do
+  span = ::OpenTelemetry::Trace.current_span
+  next nil unless span.context.valid?
+
+  {
+    trace_id: span.context.hex_trace_id,
+    span_id: span.context.hex_span_id,
+    trace_flags: span.context.trace_flags.sampled? ? 1 : 0,
+    tracestate: span.context.tracestate.to_s
+  }
+end
+```
+
+The provider must return either `nil` or a Hash with `:trace_id` (32-char lowercase hex),
+`:span_id` (16-char lowercase hex), `:trace_flags` (Integer 0-255), and `:tracestate`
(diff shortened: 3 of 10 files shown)


Sasidharan3094:
@alexr-bq @prateek-kumar-improving @Aryex Can you please review this PR

Aryex (approved):

PR #162, opened 2026-07-15
Add routing support for all cluster commands

### LINKED ISSUE #138: Cluster routing: core command methods
## Overview

Once a command can be told which cluster node to run on, two more things become useful: getting
back one result per node when a command fanned out to several, and letting advanced callers send
arbitrary commands with explicit routing.

## Description

Both reuse `Valkey::Route` from the routing-foundation work: `route:` on `call`/`call_v`, and
`route:` on existing no-key commands (`dbsize`, `flushall`, `info`, `ping`, `client_id`, the
`function_*` methods, `publish`, etc.) — server/admin-scope commands with no key to route by,
matching how Go and Python scope routing exclusively to no-key commands.

Both need `Valkey::ClusterValue` — a small wrapper representing either a single result or a
`{"host:port" => result}` Hash, depending on whether the route fanned out. Modeled on Go's
`models.ClusterValue[T]`.

## Proposal

```ruby
client.call("DBSIZE", route: Route.all_primaries)
# => ClusterValue wrapping { "10.0.0.1:6379" => 120, "10.0.0.2:6379" => 98, ... }

client.call("PING", route: Route.random)
# => ClusterValue wrapping a single value (single-node route)

client.dbsize(route: Route.all_primaries)
client.info(route: Route.all_primaries)
client.function_list(route: Route.all_primaries)
```

`route:` goes directly on `call`/`call_v` — not a separately-named `custom_command` — since
`call`/`call_v` is the redis-rb-shaped base method this client standardizes on. `ClusterValue`
wrapping is automatic based on the route's shape, and only applies when a route was actually
passed — an unrouted call's return value stays unwrapped, exactly as it is today.

## Other

Not in scope: routing `pipelined`/`multi` (separate issue — multi-node turned out unsupported
there); routing `eval`/`evalsha` (separate issue); `route:` on any key-based data command (`get`,
`set`, `hget`, etc.) — those stay implicitly routed by key.

prateek-kumar-improving:
### Summary

Adds explicit cluster routing support via a route: keyword argument on all no-key commands. Uses the new command_with_route_info FFI entrypoint (valkey-io/valkey-glide#6494) which accepts a RouteInfo C struct instead of protobuf bytes. Returns raw values directly — single value for single-node routes, Hash of "host:port" => value for multi-node routes. Without route:, all commands return their existing redis-rb-compatible values — zero breaking changes.

### Issue link

Closes #138

### Features / Changes

- New Valkey::Route class with factory methods (all_nodes, all_primaries, random, slot_id, slot_key, by_address) using FFI enums and class << self
- Added RouteType and SlotType FFI enums, RouteInfo FFI struct, and command_with_route_info binding in lib/valkey/bindings.rb
- Modified send_command to dispatch to command_with_route_info when route: is provided, legacy command() otherwise
- convert_response MAP handling: returns Hash in cluster mode, flattens to Array in standalone mode (return_map_as_hash: @cluster_mode)
- HashifyStreamEntries handles Hash input for cluster mode XRANGE/XREVRANGE/XCLAIM responses
- Commands with blocks (cluster_info, cluster_nodes) handle Hash replies from multi-node routes via transform_values
- Added route: kwarg to all no-key commands:
  - **Connection:** ping, echo, client_id, client_unpause
  - **Server:** bgrewriteaof, bgsave, config_get, config_set, config_resetstat, config_rewrite, dbsize, flushall, flushdb, info, lastsave, save, time, lolwut
  - **Function:** function_delete, function_dump, function_flush, function_kill, function_list, function_load, function_restore, function_stats, fcall, fcall_ro
  - **Generic:** call, call_v, randomkey
  - **Cluster:** cluster_info, cluster_links, cluster_myid, cluster_myshardid, cluster_nodes, cluster_shards
- Updated valkey-glide submodule to 1a4c223d9 (includes command_with_route_info)

### Testing

- test/valkey/route_test.rb — Unit tests for Route factory methods (no server required)
- test/cluster/cluster_routing_test.rb — Cluster integration tests covering ping, echo, dbsize, info, time, config_get, config_resetstat, flushall, lolwut, client_id, call, call_v, cluster_info, cluster_myid, cluster_nodes with various route types, plus backward-compatibility assertions (no route → plain value)
- Updated lint tests (connection_commands, server_commands, pubsub_commands, stream_commands) to handle Hash return type in cluster mode using cluster_mode? checks

### Checklist

- [x] Documentation is updated (if applicable).
- [x] Linters have been run (bundle exec rubocop) and pass.
- [x] Destination branch is correct - main


lib/valkey/cluster_value.rb
```diff
@@ -0,0 +1,52 @@
...
+  # @example Multi-node
+  #   cv = client.dbsize(route: Valkey::Route.all_primaries)
+  #   cv.multi_node?  #=> true
+  #   cv.multi_value  #=> { "10.0.0.1:6379" => 120, "10.0.0.2:6379" => 98 }
+  #
+  # @example Single-node
+  #   cv = client.ping(route: Valkey::Route.random)
+  #   cv.single_node? #=> true
+  #   cv.single_value #=> "PONG"
+  class ClusterValue
```
Aryex: Perhaps this is not needed anymore because Ruby is dynamically typed so a method could return multiple types. In this case either a map or a single value. `ClusterValue` would make sense for Java where methods can't return multiple types.

lib/valkey.rb
```diff
@@ -395,17 +394,38 @@ def send_command(command_type, command_args = [], &block)
...
+          channel,
+          command_type,
+          command_args.size,
+          arg_ptrs,
+          arg_lens,
+          route_buf,
+          route_str.bytesize,
+          span_ptr
+        )
+      end
```
Aryex, line 429: `command_with_route_info` can replace `command` and we can collapse this.

lib/valkey.rb
```diff
@@ -429,7 +449,12 @@ def send_command(command_type, command_args = [], &block)
       @queued_commands << [command_type, command_args.dup] if !tx_commands.include?(command_type) && result == "QUEUED"
     end
 
-    result
+    # Wrap in ClusterValue when an explicit route was provided
+    if route
+      ClusterValue.new(result, multi_node: route.multi_node?)
+    else
+      result
+    end
```
Aryex: When a nil route is given, the core does some default routing which may return a map response even when `route=nil`.

Aryex (changes requested):

lib/valkey/route.rb
```diff
@@ -0,0 +1,123 @@
...
+    ROUTE_ALL_NODES     = 0
+    ROUTE_ALL_PRIMARIES = 1
+    ROUTE_RANDOM        = 2
+    ROUTE_SLOT_ID       = 3
+    ROUTE_SLOT_KEY      = 4
+    ROUTE_BY_ADDRESS    = 5
+
+    # SlotType enum values (mirrors glide-ffi SlotType C enum)
+    SLOT_PRIMARY = 0
+    SLOT_REPLICA = 1
```
Aryex: The idiomatic way to define Ruby enum is to use Symbols

```
    # Mirrors Rust's `RouteType` enum (valkey-glide/ffi/src/lib.rs)
    RouteType = enum(
      :all_nodes, 0,
      :all_primaries,
      :random,
      :slot_id,
      :slot_key,
      :by_address
    )

    # Mirrors Rust's `SlotType` enum (a mirror of `SlotAddr`)
    SlotType = enum(
      :primary, 0,
      :replica
    )
```

lib/valkey/route.rb
```diff
@@ -0,0 +1,123 @@
...
+    # @return [Route]
+    def self.all_nodes
+      new(ROUTE_ALL_NODES, multi_node: true)
+    end
+
+    # Route to all primary nodes.
+    # @return [Route]
+    def self.all_primaries
+      new(ROUTE_ALL_PRIMARIES, multi_node: true)
+    end
```
Aryex: These static methods could defined under the `class << self` group

```
  class Route
    class << self

      def all_nodes
        new(route_type: :all_nodes)
      end

      def self.all_primaries
        ...
    end

```

lib/valkey/route.rb
```diff
@@ -0,0 +1,123 @@
...
+    # Route to a specific slot by ID.
+    # @param slot_id [Integer] slot number (0–16383)
+    # @param slot_type [Symbol] :primary or :replica
+    # @return [Route]
+    def self.slot_id(slot_id, slot_type = :primary)
+      route = new(ROUTE_SLOT_ID, multi_node: false)
+      route.instance_variable_set(:@slot_id, slot_id.to_i)
+      route.instance_variable_set(:@slot_type, slot_type == :replica ? SLOT_REPLICA : SLOT_PRIMARY)
+      route
+    end
```
Aryex: We should be setting these instance variable through the constructor params instead. `instance_variable_set` is meta programming similar to Java reflections.

lib/valkey/bindings.rb
```diff
@@ -242,6 +253,19 @@ class CommandResult < FFI::Struct
       :ulong        # span_ptr (u64)
     ], :pointer, blocking: true # returns *mut CommandResult, releases GVL during I/O
 
+    attach_function :command_with_route_info, [
```
Aryex, line 272: `command` binding could probably be replaced with `command_with_route_info`.

lib/valkey/cluster_value.rb
prateek-kumar-improving: Removed in this commit: https://github.com/valkey-io/valkey-glide-ruby/pull/162/changes/3333d2de357c690712276a666d15961f812e9a82

lib/valkey/route.rb
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-ruby/pull/162/changes/275283ff07b884192121fd0b3076798fd0345611

lib/valkey/route.rb
prateek-kumar-improving: Updated here: https://github.com/valkey-io/valkey-glide-ruby/pull/162/changes/0678dbfbc4f762b0935e08fbc72bf4414d71f897

lib/valkey/route.rb
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-ruby/pull/162/changes/275283ff07b884192121fd0b3076798fd0345611

lib/valkey.rb
prateek-kumar-improving: Updated convert_response logic as discussed.

lib/valkey/bindings.rb
prateek-kumar-improving: As discussed, using command with standalone.

lib/valkey.rb
prateek-kumar-improving: As discussed, using command method.

lib/valkey/route.rb
```diff
@@ -0,0 +1,110 @@
...
+  # Hash of `"host:port" => value`.
+  #
+  # @example
+  #   client.dbsize(route: Valkey::Route.all_primaries)
+  #   client.ping(route: Valkey::Route.random)
+  #   client.info(route: Valkey::Route.all_nodes)
+  #
+  # @see https://valkey.io/topics/cluster-spec
+  class Route
+    attr_reader :multi_node
```
Aryex: Can we remove this? It doesn't look like it is used.

Aryex (changes requested):

lib/valkey/route.rb
prateek-kumar-improving: Updated. Fixed in this commit: https://github.com/valkey-io/valkey-glide-ruby/pull/162/changes/f18a88eee172747f47bbab3ac72ffed79e22595b

Aryex (approved):

lib/valkey.rb
```diff
@@ -630,8 +651,8 @@ def convert_response(res, &block)
           map[map_key] = map_value
         end
 
-        # technically it has to return a Hash, but as of now we return just one pair
-        map.to_a.flatten(1) # Flatten to get pairs
+        # Return as Hash in cluster mode, flatten to pairs in standalone mode.
+        return_map_as_hash ? map : map.to_a.flatten(1)
```
nderraugh, line 655: Someone who already has a cluster client and has been calling old `client.hello` or `client.pubsub(:numsub, "ch1")` with no `route:` will start getting a Hash back instead of an Array, with this version.

Probably needs to be explicitly called out as a breaking change, or managed differently.

lib/valkey/commands/server_commands.rb
```diff
@@ -60,77 +60,84 @@ def config_get(*args)
...
       #
-      # @param [String, Symbol] cmd e.g. "commandstats"
-      # @return [Hash<String, String>]
-      def info(cmd = nil)
-        send_command(RequestType::INFO, [cmd].compact) do |reply|
+      # @param cmd [String, Symbol, nil] section name (e.g. "commandstats")
+      # @param route [Valkey::Route, nil] cluster routing. When routed, may return a Hash of node => value.
+      # @return [Hash]
+      def info(cmd = nil, route: nil)
+        send_command(RequestType::INFO, [cmd].compact, route: route) do |reply|
```
nderraugh, line 140: Similar pattern to 5364db3540902ff2469364d2d815755ce153766e.

Need to handle reply as a Hash to conform with the method's docs.  Otherwise it gets returned as the unparsed reply.

lib/valkey.rb
prateek-kumar-improving: The code is still not released, so this is okay.

lib/valkey/commands/server_commands.rb
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-ruby/pull/162/changes/7065ccdcf234505b621b8201fbd47e6931f71877

lib/valkey.rb
Aryex: @nderraugh For some context, we are not GA yet so breaking changes like this is expected. 

However I will make a follow-up PR to return hash in all cases, and add a backward compatible option for users whom are dependent on this return type. This is pending on some confirmation.

nderraugh (approved):

PR #163, opened 2026-07-16
fix: free command results to prevent memory leak

Aryex:
### Summary

This PR fixes an issue where ffi command results were not properly released.

### Changes

- Added bindings for freeing result from ffi.
- Updated call sites properly free the ffi result.

### Checklist

-   [ ] This Pull Request is related to one issue.
-   [ ] Tests are added or updated.


lib/valkey/bindings.rb
```diff
@@ -226,6 +226,10 @@ class CommandResult < FFI::Struct
       :pointer # *mut ConnectionResponse
     ], :void
 
+    attach_function :free_command_result, [
```
xShinnRyuu, line 256: You may want to check other methods that return `*mut CommandResult` that is never freed.


The `invoke_script` method in `scripting_commands.rb` line 261 also returns a `*mut CommandResult` that is never freed.

I'd suggest adding in some kind of testing for this as well

Aryex: Good catch. Addressed in
- [fix(ruby): free leaked FFI memory in invoke_script and script_load](https://github.com/valkey-io/valkey-glide-ruby/pull/163/commits/a0f14a523b3a009358d8650bccf3843a152d81d4)
- [test(ruby): add RSS-based memory leak smoke tests](https://github.com/valkey-io/valkey-glide-ruby/pull/163/commits/52ef7d002309fe075f26eeea2122b4f9abf5027d)

lib/valkey/commands/scripting_commands.rb
```diff
@@ -258,22 +262,26 @@ def invoke_script(script, args: [], keys: [])
...
+            arg_ptrs,
+            arg_lens,
+            route_buf,
+            route.bytesize,
+            0 # span_ptr for OpenTelemetry (0 = no span)
+          )
+
+          convert_response(res)
+        ensure
+          Bindings.free_command_result(res)
```
xShinnRyuu: Missing null guard here `Bindings.free_command_result(res) if res && !res.null?`

lib/valkey/bindings.rb
```diff
@@ -226,6 +226,23 @@ class CommandResult < FFI::Struct
...
+      :pointer # *mut CommandResult
+    ], :void
+
+    attach_function :free_script_hash_buffer, [
+      :pointer # *mut ScriptHashBuffer
+    ], :void
+
+    attach_function :drop_script, [
```
xShinnRyuu, line 264: Is `drop_script` used anywhere?

lib/valkey/bindings.rb
```diff
@@ -226,6 +226,23 @@ class CommandResult < FFI::Struct
...
+      :pointer, # *mut u8 (hash bytes)
+      :ulong    # usize (hash length)
+    ], :pointer # returns *mut c_char (null on success, error string on failure)
+
+    attach_function :free_drop_script_error, [
```
xShinnRyuu, line 269: Is `free_drop_script_error` used anywhere?

.rubocop.yml
```diff
@@ -35,7 +35,7 @@ Metrics/CyclomaticComplexity:
   Enabled: false
 
 Metrics/MethodLength:
-  Max: 20
+  Max: 50
```
xShinnRyuu, line 38: nit: is this number to evaluate the threshold for `CyclomaticComplexity`? seems like a rather large jump

lib/valkey/commands/scripting_commands.rb
Aryex: Removed because `free_command_result` already null checks at the core. No string opinion on it though, can be added back.

lib/valkey/bindings.rb
Aryex: No. These methods have not yet been implemented, but added for completeness.

lib/valkey/bindings.rb
Aryex: No. These methods have not yet been implemented, but added for completeness.

.rubocop.yml
Aryex: No it's just for line count. It set the max line count a method is allowed to have. Note that CyclomaticComplexity is not enabled right now.

lib/valkey/commands/scripting_commands.rb
xShinnRyuu: But the other sections of code using `Bindings.free_command_result(res)` in this PR have the null check

xShinnRyuu (approved):

prateek-kumar-improving (approved):

xShinnRyuu (approved):

PR #168, opened 2026-07-20
Standardize return type and add `flatten_map` option 

Aryex:
## Summary

Standardized return type to always convert GLIDE Core map to Ruby Hash. A `flatten_map` option is added as backward compatibility that matches redis-rb 4.x return types. 

## Changes

- `lib/valkey.rb`: Add `@flatten_map` instance variable from `options[:flatten_map]`; replace the `return_map_as_hash` / `@cluster_mode` conditional with `@flatten_map`
- Test updates: Remove `if cluster_mode?` branching that previously differentiated Hash (cluster) vs Array (standalone) — MAP responses are now always Hash by default regardless of mode
- `test/valkey/flatten_map_test.rb`: New integration test verifying both branches using HELLO command

## Usage

```ruby
# Default: MAP responses return as Hash
client = Valkey.new(host: "localhost", port: 6379)
client.hello  # => {"server"=>"valkey", "version"=>"9.0.2", ...}

# redis-rb 4.x compat: MAP responses flattened to Array
client = Valkey.new(host: "localhost", port: 6379, flatten_map: true)
client.hello  # => ["server", "valkey", "version", "9.0.2", ...]
```

## Testing

- `bundle exec rubocop` — passes
- `bundle exec rake test:standalone` — 770 tests, 0 failures
- `bundle exec rake test:cluster` — 776 tests, 0 failures

## Checklist

- [x] Related to one change
- [x] Tests added/updated
- [x] Linters pass
- [ ] Destination branch: release-1.0

### DIFF
--- a/lib/valkey.rb
+++ b/lib/valkey.rb
@@ -291,6 +291,11 @@ class Valkey
     # Store cluster mode flag for response handling (MAP returns Hash in cluster, Array in standalone)
     @cluster_mode = options[:cluster_mode] ? true : false
 
+    # Returns GLIDE Core map as a flattened-map array
+    # Ex: { "key1" => "value1", "key2" => "value2" } becomes ["key1", "value1", "key2", "value2"]
+    # Compatibility for redis-rb 4.x
+    @flatten_map = options[:flatten_map] ? true : false
+
     # Track transactional state for `MULTI` / `EXEC` / `DISCARD` helpers.
     # This avoids Ruby warnings about uninitialised instance variables and
     # gives us a single source of truth for whether we're inside a TX.
@@ -435,7 +440,7 @@ class Valkey
         )
       end
 
-      result = convert_response(res, return_map_as_hash: @cluster_mode, &block)
+      result = convert_response(res, &block)
     ensure
       # Free the native CommandResult (arena + response + error) to prevent memory leak
       Bindings.free_command_result(res) if res && !res.null?
@@ -614,7 +619,7 @@ class Valkey
     [arg_ptrs, arg_lens, buffers]
   end
 
-  def convert_response(res, return_map_as_hash: false, &block)
+  def convert_response(res, &block)
     result = Bindings::CommandResult.new(res)
 
     if result[:response].null?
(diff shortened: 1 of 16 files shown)


lib/valkey.rb
```diff
@@ -670,8 +675,7 @@ def convert_response(res, return_map_as_hash: false, &block)
           map[map_key] = map_value
         end
 
-        # Return as Hash in cluster mode, flatten to pairs in standalone mode.
-        return_map_as_hash ? map : map.to_a.flatten(1)
+        @flatten_map ? map.to_a.flatten(1) : map
```
prateek-kumar-improving, line 678: When flatten_map is false, do we also return a map for standalone? 
Does this match other clients?

test/lint/connection_commands.rb
```diff
@@ -152,35 +152,20 @@ def test_client_no_touch
 
     def test_hello_default
       result = r.hello
```
prateek-kumar-improving, line 154: All these tests are for the default case when flatten_map is false and map is returned.
We need to add more tests when flatten_map is true.

Basically, the assertions in the else part will be added to those tests.

lib/valkey.rb
Aryex: When flatten_map is false, we return a hash (map) for both standalone and cluster. This option is provided for backward compatibility. 

The Python and Node don't have this `flatten_map` option. It is specific to Ruby

test/lint/connection_commands.rb
Aryex: The `flatten_map = false` code path is covered in`test/valkey/flatten_map_test.rb`. I assume existing code should handle the flatten map correctly. Do you want to add more command tests?

prateek-kumar-improving: The tests below like test_hello_with_setname and test_hello_with_version are not tested anymore when they return Array.

prateek-kumar-improving (approved):

xShinnRyuu (approved):

Issue #175, opened 2026-07-22
hset (and any command) silently corrupts a nested Array argument instead of raising or working


Sasidharan3094:
# `hset` (and any command) silently corrupts a nested Array argument instead of raising or working

## Description

Passing a single key-value pair to `hset` as a 2-element `Array`, rather
than as two separate positional arguments or a `Hash`, sends a corrupted
command to the server instead of working correctly or raising a clear
error at the call site:

```ruby
valkey.hset("k", ["field", "value"])
# Valkey::CommandError: An error was signalled by the server: - ResponseError: wrong number of arguments for 'hset' command
```

This same style of call works fine against `redis-rb`:

```ruby
redis.hset("k", ["field", "value"])
# => 1
redis.hgetall("k")
# => {"field"=>"value"}
```

So any code migrating from `redis-rb` that happens to build its
arguments this way (even if unintentionally) breaks against this client.

## Steps to reproduce

```ruby
require "valkey"

valkey = Valkey.new(host: "127.0.0.1", port: 6379)
valkey.del("k")
valkey.hset("k", ["field", "value"])
```

## Expected behavior

Either:
- `hset("k", ["field", "value"])` succeeds, matching `redis-rb`'s
  behavior (`redis.hset` and this gem's `hset` share the exact same Ruby
  method body, so this is arguably the more consistent expectation), or
- it raises an `ArgumentError` client-side, immediately, rather than
  sending a malformed command to the server.

## Actual behavior

The server rejects the command with `wrong number of arguments for
'hset' command`. Depending on how the caller handles errors, this can
surface far from the actual mistake - e.g. as a downstream read coming
back empty, rather than as the write itself failing loudly.

## Root cause

`build_command_args` (`lib/valkey.rb`) converts each top-level element of
the command array to a string via `Array#to_s`/`Hash#to_s` and stops
there - it never flattens nested structures. So `["field", "value"]`
becomes the single garbled string `'["field", "value"]'` instead of two
separate arguments, leaving the server with the wrong argument count.

`redis-rb`'s underlying transport (`redis-client`'s `CommandBuilder
.generate`) doesn't have this problem, because it builds commands with
`flat_map`, which auto-flattens any Array-valued element by one level as
an incidental side effect of `flat_map`'s own semantics (not something
`redis-client` deliberately built just for this case).

## Environment

- `valkey-glide-rb` (Ruby client)
- Reproduced against a real Valkey server (v7.2.11), not just inferred
[...]

PR #176, opened 2026-07-22
Flatten nested Array/Hash command arguments, matching redis-client

Sasidharan3094:
# Flatten nested Array/Hash command arguments, matching redis-client

This PR fixes the [issue](https://github.com/valkey-io/valkey-glide-ruby/issues/175)

## Problem

Any command call that passes a nested `Array`/`Hash` as one of its
arguments - e.g. `hset(key, [field, value])` - silently sends a corrupted
command instead of raising or working correctly:

```ruby
valkey.hset("k", ["field", "value"])
# ResponseError: wrong number of arguments for 'hset' command
```

This specific shape - `hset(key, [field, value])`, passing a single
key-value pair as a 2-element `Array` rather than as two separate
positional arguments or a `Hash` - is a real, if slightly unusual, calling
style: it works fine against `redis-rb`, so any code migrating from
`redis-rb` that happens to build its arguments this way breaks silently
against this client instead of raising a clear error at the call site
that made the mistake obvious. Depending on how the caller handles
errors, this can surface far from the actual bug - e.g. as a downstream
read coming back empty, rather than as the write itself failing loudly.

## Root cause

`build_command_args` (`lib/valkey.rb`) converts each **top-level** element
of the command array to a string and stops there - no flattening:

```ruby
command_args.each_with_index do |arg, i|
  arg = arg.to_s # Ensure we convert to string
  ...
end
```

So `[key, [field, value]]` turns the nested array into a single string via
`Array#to_s` (`'["field", "value"]'`) - one garbled argument instead of
two real ones, leaving the server with an odd/wrong argument count.

`redis-rb`'s underlying transport (`redis-client`'s `CommandBuilder
.generate`) does not have this problem, because it builds commands with
`flat_map`:

```ruby
command = args.flat_map do |element|
  case element
  when Hash
    element.flatten
  else
    element
  end
end
```

`flat_map` auto-flattens *any* Array-valued block result by one level -
this isn't Hash-specific, it's an incidental side effect of `flat_map`'s
own semantics. That's why `redis.hset(key, [field, value])` has always
worked correctly under `redis-rb`, even though nothing in `hset`'s own
code explicitly flattens that particular shape.

## What changed

**`lib/valkey.rb`**

- `build_command_args`: now does the same `flat_map` transform before
  building the FFI buffers, so a nested `Array`/`Hash` argument gets
  flattened exactly like `redis-client` does it:
  ```ruby
  command_args = command_args.flat_map { |el| el.is_a?(Hash) ? el.flatten : el }
  ```
- `send_command` / `send_batch_commands`: both were passing the FFI layer
  the **original**, pre-flatten array's `.size` as `arg_count`, while
  `arg_ptrs`/`arg_lens` were being built from the (now correctly
  flattened) array - two different lengths, silently out of sync. Fixed
  by having `build_command_args` also return the flattened array itself,
  and reassigning the caller's local `command_args` to it before using
  `.size` anywhere. This is the actual reason the `build_command_args`
  change alone didn't fix anything on its own - the count sent to the FFI
  call was still wrong even after the pointers were built correctly.

**`lib/valkey/commands/string_commands.rb`**

- `set`, `setex`, `psetex`, `setnx`, `setrange` now explicitly call
  `value.to_s` themselves, matching `redis-rb`'s own exact convention for
  these five methods specifically. Without this, a deliberately-passed
  non-`String` value (e.g. `%w[b a r]`, exercised by this gem's own
  `test_*_with_non_string_value` tests - a *single opaque value* meant to
  be stringified as a whole) would now be ambiguous with a genuine
  multi-value list and get incorrectly flattened instead of stringified,
  changing the argument count and breaking the command.

### Deliberately NOT changed: every other value-taking command

Checked `redis-rb`'s own source for `linsert`, `lrem`, `lset`, `hsetnx`,
`append` (and confirmed empirically for `lset`): none of them call
`.to_s` either, and `redis-rb` itself raises the identical `wrong number
of arguments` error if you pass an `Array` as their "value" argument.
There's no established contract for these to safely accept a non-`String`
value in the first place - on either client - so leaving them unchanged
matches real `redis-rb` behavior exactly, rather than speculatively
hardening past what's actually guaranteed anywhere.

## Verification

- Original repro fixed:
  ```ruby
  valkey.hset("k", ["field", "value"])  # => 1
  valkey.hgetall("k")                    # => {"field"=>"value"}
  ```
- Full suite: 560 tests, 0 errors (1 pre-existing, unrelated failure -
  `test_library_name_is_set`, fails identically with or without this
  change).
- Confirmed real `redis-rb` raises the exact same error as this client
  for `lset(key, index, ["a", "b"])` - i.e. the untouched methods aren't a
  newly-introduced gap, they're consistent with the reference client.
- Side-by-side against `redis-rb` for the original bug: identical
  results on both (`hset(key, [field, value])` → `1`, `hgetall` →
  `{"field"=>"value"}`, on both clients).

### DIFF
--- a/lib/valkey.rb
+++ b/lib/valkey.rb
@@ -378,7 +378,10 @@ class Valkey
       arg_lens.put_ulong(0, 0)
       _buffers = [] # nothing to keep alive
     else
-      arg_ptrs, arg_lens, _buffers = build_command_args(command_args)
+      # Reassign command_args to the (possibly flattened) array build_command_args
+      # actually built arg_ptrs/arg_lens from - see that method for why using the
+      # original, pre-flatten array's size below would go out of sync with them.
+      arg_ptrs, arg_lens, _buffers, command_args = build_command_args(command_args)
     end
 
     # Create OpenTelemetry span if sampling is enabled, as a child of the app's current
@@ -497,7 +500,10 @@ class Valkey
     buffers = [] # Keep references to prevent GC
 
     commands.each do |command_type, command_args, block|
-      arg_ptrs, arg_lens, arg_bufs = build_command_args(command_args)
+      # Reassign to the (possibly flattened) array build_command_args actually
+      # built arg_ptrs/arg_lens from - see that method for why the original,
+      # pre-flatten array's size would go out of sync with them below.
+      arg_ptrs, arg_lens, arg_bufs, command_args = build_command_args(command_args)
 
       cmd = Bindings::CmdInfo.new
       cmd[:request_type] = command_type
@@ -602,9 +608,25 @@ class Valkey
   end
 
   def build_command_args(command_args)
+    # Mirror redis-client's CommandBuilder#generate: flat_map every element,
+    # turning a Hash into alternating key/value pairs (same one-level
+    # auto-flatten redis-client gets from flat_map on a plain Array element
+    # too, as a side effect of flat_map's own semantics - not Hash-specific).
+    # Every existing caller in this codebase already pre-flattens its own
+    # arguments (see e.g. hset/mset/zadd/keys.flatten!), so this is a no-op
+    # for them - it only changes behavior for a caller that passes a raw,
+    # un-flattened Array/Hash straight through, which previously got
+    # silently mis-serialized via Array#to_s/Hash#to_s into one garbled
+    # string argument instead of several real ones.
+    command_args = command_args.flat_map { |el| el.is_a?(Hash) ? el.flatten : el }
+
     # For empty arrays, pass NULL pointers as per Rust FFI contract
     # This matches Go's approach which successfully uses nil pointers
-    return [FFI::Pointer::NULL, FFI::Pointer::NULL, []] if command_args.empty?
+    # Also returns the (possibly flattened) command_args - callers must use
+    # THIS array's size for the FFI arg_count, not their own original
+    # pre-flatten local variable, or arg_count and arg_ptrs/arg_lens go out
+    # of sync when flattening actually changed the element count.
+    return [FFI::Pointer::NULL, FFI::Pointer::NULL, [], []] if command_args.empty?
 
     arg_ptrs = FFI::MemoryPointer.new(:pointer, command_args.size)
     arg_lens = FFI::MemoryPointer.new(:ulong, command_args.size)
@@ -619,7 +641,7 @@ class Valkey
       arg_lens.put_ulong(i * 8, arg.bytesize)
     end
 
-    [arg_ptrs, arg_lens, buffers]
+    [arg_ptrs, arg_lens, buffers, command_args]
   end
 
   def convert_response(res, return_map_as_hash: false, &block)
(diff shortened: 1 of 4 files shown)


lib/valkey.rb
```diff
@@ -602,9 +608,25 @@ def build_periodic_checks(periodic_checks)
...
+    # Mirror redis-client's CommandBuilder#generate: flat_map every element,
+    # turning a Hash into alternating key/value pairs (same one-level
+    # auto-flatten redis-client gets from flat_map on a plain Array element
+    # too, as a side effect of flat_map's own semantics - not Hash-specific).
+    # Every existing caller in this codebase already pre-flattens its own
+    # arguments (see e.g. hset/mset/zadd/keys.flatten!), so this is a no-op
+    # for them - it only changes behavior for a caller that passes a raw,
+    # un-flattened Array/Hash straight through, which previously got
+    # silently mis-serialized via Array#to_s/Hash#to_s into one garbled
+    # string argument instead of several real ones.
```
Aryex: Could we simplify this to 
```suggestion
# Flatten nested Arrays/Hashes to match redis-client's behavior.
```

lib/valkey.rb
```diff
@@ -602,9 +608,25 @@ def build_periodic_checks(periodic_checks)
...
+    command_args = command_args.flat_map { |el| el.is_a?(Hash) ? el.flatten : el }
+
     # For empty arrays, pass NULL pointers as per Rust FFI contract
     # This matches Go's approach which successfully uses nil pointers
-    return [FFI::Pointer::NULL, FFI::Pointer::NULL, []] if command_args.empty?
+    # Also returns the (possibly flattened) command_args - callers must use
+    # THIS array's size for the FFI arg_count, not their own original
+    # pre-flatten local variable, or arg_count and arg_ptrs/arg_lens go out
+    # of sync when flattening actually changed the element count.
```
Aryex: Perhaps this comment should be at the doc string for the method instead so it has more visibility.

lib/valkey.rb
```diff
@@ -378,7 +378,10 @@ def send_command(command_type, command_args = [], route: nil, &block)
       arg_lens.put_ulong(0, 0)
       _buffers = [] # nothing to keep alive
     else
-      arg_ptrs, arg_lens, _buffers = build_command_args(command_args)
+      # Reassign command_args to the (possibly flattened) array build_command_args
+      # actually built arg_ptrs/arg_lens from - see that method for why using the
+      # original, pre-flatten array's size below would go out of sync with them.
+      arg_ptrs, arg_lens, _buffers, command_args = build_command_args(command_args)
```
Aryex: If command_args could be flatten, we should use a different variable to show that it could've have been transformed. Perhaps use `flatten_args` and then we could remove the comment above.

Aryex (commented):
Overall lgtm. Just a few comments on code hygiene.

Sasidharan3094:
Thank you @Aryex for the review comments, will address it and will merge it.

pratheep-kumar (approved):
LGTM

PR #177, opened 2026-07-22
test: standardize TLS tests and remove committed certs

Aryex:
## Summary

Gitleaks flagged 3 committed RSA private keys under `test/fixtures/ssl/`
(`ca-key.pem`, `client-key.pem`, `server-key.pem`). These were self-signed,
test-only TLS fixtures with no production impact, but committed keys should
not live in the repo.

Rather than keep a local cert set, this standardizes the TLS tests on
`cluster_manager.py --tls` — the same mechanism CI already uses — as the single
source of certs:

- **`SslHelper`** now reads certs only from `TLS_CERT_DIR`
  (`ca.crt` / `server.crt` / `server.key`), and raises a helpful message if it
  is unset. `SKIP_TLS_TESTS=true` opts out (as macOS CI already does).
- **Removed `test/fixtures/ssl/`** entirely (the 6 committed `.pem` files, the
  `generate_certs.rb` generator, and the README). No committed or generated
  certs remain in the repo.
- **Docs** (`DEVELOPER.md`, `AGENTS.md`) updated with the `cluster_manager.py`
  start/stop commands and the `TLS_CERT_DIR` requirement.

## Why this is safe for CI

CI never used the committed fixtures — both TLS jobs run
`cluster_manager.py --tls start -p 6380` and export
`TLS_CERT_DIR=.../valkey-glide/utils/tls_crts`, which takes priority in
`SslHelper`. No workflow references `test/fixtures/ssl/`.

## Behavior

| Scenario | Result |
|----------|--------|
| `TLS_CERT_DIR` set (CI / documented local flow) | TLS tests run |
| `SKIP_TLS_TESTS=true` | TLS tests skip gracefully |
| Neither set | TLS tests raise with instructions to start a server / set the var |

## Testing

Verified locally against `cluster_manager.py`-started standalone (:6379) + TLS
(:6380) servers:

- Full standalone suite: **776 tests, 0 failures, 0 errors**
- SSL tests: **12/12 pass** with `TLS_CERT_DIR` set; **4 handshake tests skip**
  with `SKIP_TLS_TESTS=true`; **raise** with a clear message when neither is set
- `bundle exec rubocop` clean on changed files

## Notes

- The removed keys still exist in git history; since they are worthless
  self-signed test keys this is low-risk, but a history scrub or
  `.gitleaksignore` entry could follow separately if the history scan must go
  fully green.
- A follow-up proposal for a `bin/test-server` helper to start/stop these
  servers consistently is tracked separately.


jamesx-improving (approved):
Approved with nit: `SKIP_TLS_TESTS=true` make the TLS tests opt-out. Maybe consider change it to opt-in, or, when we establish the PR vs FMT mechanism for tests, make it FMT-only, skipping PR, if there's CI speed or stability concern.

affonsov (approved):

PR #178, opened 2026-07-22
fix: correctly route client cert and key to ffi options

Aryex:
## Summary

- `ssl_params[:cert]` and `ssl_params[:key]` were appended into the `root_certs` array (CA trust anchors), so mTLS silently never presented a client identity to the server
- Now routes them to `json_options["client_cert"]` / `json_options["client_key"]`, which the FFI layer already maps to `ConnectionRequest.client_cert` / `.client_key` in glide-core

### DIFF
--- a/lib/valkey.rb
+++ b/lib/valkey.rb
@@ -156,50 +156,8 @@ class Valkey
     # TLS/SSL certificates
     root_certs = []
     if options[:ssl_params].is_a?(Hash)
-      # ca_file - read CA certificate file (PEM or DER format)
-      if options[:ssl_params][:ca_file]
-        ca_file = options[:ssl_params][:ca_file]
-        raise ArgumentError, "CA file does not exist: #{ca_file}" unless File.exist?(ca_file)
-        raise ArgumentError, "CA file is not readable: #{ca_file}" unless File.readable?(ca_file)
-
-        root_certs << File.binread(ca_file)
-      end
-
-      # cert - client certificate for mutual TLS (file path or OpenSSL::X509::Certificate)
-      if options[:ssl_params][:cert]
-        cert_data = if options[:ssl_params][:cert].is_a?(String)
-                      cert_file = options[:ssl_params][:cert]
-                      raise ArgumentError, "Cert file does not exist: #{cert_file}" unless File.exist?(cert_file)
-                      raise ArgumentError, "Cert file is not readable: #{cert_file}" unless File.readable?(cert_file)
-
-                      File.binread(cert_file)
-                    elsif options[:ssl_params][:cert].respond_to?(:to_pem)
-                      options[:ssl_params][:cert].to_pem
-                    elsif options[:ssl_params][:cert].respond_to?(:to_der)
-                      options[:ssl_params][:cert].to_der
-                    else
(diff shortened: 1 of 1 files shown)


lib/valkey.rb
```diff
@@ -180,10 +180,10 @@ def initialize(options = {})
                     else
                       options[:ssl_params][:cert].to_s
                     end
-        root_certs << cert_data
+        json_options["client_cert"] = cert_data
       end
```
currantw: Where are the `cert` and `key` argument names (i.e. `[:ssl_params][:cert]` and `[:ssl_params][:key]`) coming from? It seems like this is the _root_ (pun intended) of the confusion. Would `client_cert` and `client_key` be more clear?

lib/valkey.rb
```diff
@@ -180,10 +180,10 @@ def initialize(options = {})
...
 
-      # key - client key (file path or OpenSSL::PKey)
+      # key - client private key for mutual TLS (file path or OpenSSL::PKey)
       if options[:ssl_params][:key]
```
currantw: The `cert` and `key` blocks seems to be pretty much complete duplicates of each other. Any value in extracting a helper?

lib/valkey.rb
currantw: For root certificates, we have separate `ca_file` and `ca_path` arguments.

Why would we not handle root certificate, client certificate. and client key all the same? Either provide `_file` and `_path` options for all of them – or provide a single option for each that supports a file path OR a certificate?

currantw (commented):
Reviewed to 93350f7.

lib/valkey.rb
Aryex: It probably comes from redis-rb options. See [here](https://github.com/redis/redis-rb#ssltls-support). This client was intended to have a similar interface to redis-rb.

lib/valkey.rb
Aryex: redis-rb ssl_params is meant to be compatible with `OpenSSL::SSL::SSLContext`, which contains ca_path. Thus we also support it.

See [here](https://github.com/redis/redis-rb#ssltls-support).

lib/valkey.rb
Aryex: good point. Refactored in [refactor(ruby): extract read_ssl_value helper for ssl_params handling](https://github.com/valkey-io/valkey-glide-ruby/pull/178/commits/52929491772399e8856444e7ba9447c4bf59a59a)

affonsov (approved):

yipin-chen (approved):
LGTM

lib/valkey.rb
currantw: Okay, that makes sense. Regrettable, but not sure we can do anything about it then.

lib/valkey.rb
currantw: Thanks 🙏

lib/valkey.rb
currantw: Gotcha. That makes sense. Thanks.

currantw (approved):
Reviewed to 5292949.
Looks good ✅

currantw:
@Aryex Are there any existing mTLS tests planning for this client? It seems like those should have caught this error – or those planned for future work?

Issue #179, opened 2026-07-22
Refactor FFI pointer-returning methods to use structs


Aryex:
## Problem

Two methods return arrays where callers must retain a buffer element to prevent GC from freeing memory that pointers reference. This is error-prone — `invoke_script` previously dropped `buffers`, causing a use-after-free bug.

| Method | Location | Returns |
|--------|----------|---------|
| `build_command_args` | `lib/valkey.rb` | `[ptrs, lens, buffers]` |
| `Route#to_ffi` | `lib/valkey/route.rb` | `[route_info, pinned_buffers]` |

## Proposal

Replace array returns with structs so callers hold a single reference and cannot accidentally discard the buffers:

```ruby
CommandArgs = Struct.new(:ptrs, :lens, :bufs)
RouteArgs   = Struct.new(:info, :bufs)
```

## Affected call sites

- `lib/valkey.rb` — `send_command`, `send_batch_commands`
- `lib/valkey/commands/scripting_commands.rb` — `invoke_script`

## Considerations

- Minimal perf impact — Struct allocation is comparable to Array
- Could use `Data.define` (Ruby 3.2+) for immutability if minimum Ruby version allows
- Existing tests should pass without modification since the change is internal

jeremyprime:
Also look at `build_config` in `opentelemetry.rb`.  The `MemoryPointer`s have the same possible issue of being GC'ed since nothing retains them.

PR #180, opened 2026-07-22
fix: retain FFI buffers in invoke_script to prevent use-after-free

Aryex:
## Summary

- `invoke_script` discarded the buffer objects returned by `build_command_args`, allowing GC to free backing memory during the blocking FFI call
- Retain buffers for both args and keys, matching the pattern in `send_command` and `send_batch_commands`
- Add documentation and TODO refs to #179 for struct refactor

### DIFF
--- a/lib/valkey/route.rb
+++ b/lib/valkey/route.rb
@@ -67,6 +67,8 @@ class Valkey
     # Returns both the struct and any pinned memory buffers that must remain
     # alive until the FFI call completes.
     #
+    # TODO: Refactor to return a struct (https://github.com/valkey-io/valkey-glide-ruby/issues/179)
+    #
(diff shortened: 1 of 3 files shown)


lib/valkey/route.rb
```diff
@@ -67,6 +67,8 @@ def by_address(host, port)
     # Returns both the struct and any pinned memory buffers that must remain
     # alive until the FFI call completes.
     #
+    # TODO: Refactor to return a struct (https://github.com/valkey-io/valkey-glide-ruby/issues/179)
```
jeremyprime, line 70: Possible similar issue with `build_config` in `opentelemetry.rb`, I commented on #179.

jeremyprime (approved):

PR #194, opened 2026-07-28
fix(ruby): reject nil and unsupported types in command args

### LINKED ISSUE #191: `call` and command methods silently coerce `nil` positional args to empty string
# `call` and command methods silently coerce `nil` positional args to empty string

> **Related work already in flight — #175 / open PR #176.**
>
> #175 reported the same `build_command_args` function silently corrupting **nested Array/Hash** arguments. PR #176 fixes that specific symptom by adding a `flat_map` flatten pass. However, PR #176 does **not** add a nil check — and its `string_commands.rb` hunk explicitly writes `args = [key, value.to_s]` in `set`, so `set(k, nil) => ""` remains broken after that PR merges. This `nil`-coercion finding is therefore a sibling defect of #175 that survives PR #176; the cleanest fix is one extra guard in the same function.

## Description

Passing `nil` as a positional argument to `call` — or to any command method that eventually reaches `build_command_args` — silently succeeds and sends an empty string to the server. This is a subtle data-loss failure. `redis-client` raises `TypeError` for `nil` positional args, and `redis-rb` rejects most `nil`s, so migrating code that previously would have surfaced as a clear `TypeError` becomes a mystery empty-string in production.

## Steps to reproduce

```ruby
require "valkey"
c = Valkey.new
c.del("bb:nil")

c.call("SET", "bb:nil", nil)   # no error
c.get("bb:nil")                # => ""   (silently stored empty string)
```

## Expected behavior

`c.call("SET", "bb:nil", nil)` raises `TypeError` at the call site, matching `redis-client`. The mistake is caught before it hits the wire.

## Actual behavior

Command succeeds, key is set to the empty string. Downstream code that treats `c.get("bb:nil")` as truthy (any non-nil is truthy in Ruby) then behaves incorrectly.

## Root cause

`lib/valkey.rb` around lines 596–602, in `build_command_args`:

```ruby
command_args.each_with_index do |arg, i|
  arg = arg.to_s   # Ensure we convert to string
  buf = FFI::MemoryPointer.from_string(arg.to_s)
  …
end
```

Blanket `arg.to_s` catches every value including `nil` (`nil.to_s == ""`), `Array` (`[1,2].to_s == "[1, 2]"`, cf. #175), and `Hash`. Anything that responds to `to_s` ends up on the wire with a plausible-looking but wrong value.

## Suggested fix

Reject the problematic types up front in `build_command_args`:

```ruby
command_args.each_with_index do |arg, i|
  case arg
  when nil
    raise TypeError, "nil is not allowed as a command argument"
  when Array, Hash
    raise TypeError, "#{arg.class} is not allowed as a command argument " \
                     "(flatten before calling — see issue #175)"
  end
  arg = arg.to_s
  …
end
```

Cleanest is to fold the guard into the same file PR #176 is already editing, so both defects land together. Alternatively, restrict allowed types to `String / Integer / Float / Symbol / Numeric` and reject everything else, matching redis-client's stricter argument type check.

## Environment

- Ruby: 4.0.6, arm64-darwin25
- Valkey: 8.1.3 (standalone)

Aryex:
## Summary

Minimal backport of the #191 fix to `release-1.0`. Raises `TypeError` for `nil`, `Array`, `Hash`, and other unsupported types in `build_command_args` instead of silently coercing them via `#to_s` (which turns `nil` into empty string `""`).

Only `String`, `Symbol`, `Integer`, and `Float` are accepted — matching `redis-client`'s behavior.

Fixes #191

## What's different from main

This is a targeted fix only — it does **not** include the Array/Hash flattening refactor or the scripting use-after-free fix from the full squash commit on main (`9005f65`). Those changes are independent improvements that can be brought over separately if needed.

## Changes

- `lib/valkey.rb` — type-check guard in `build_command_args`
- `test/valkey/arg_type_validation_test.rb` — new test module
- `test/standalone/valkey_test.rb` — include new module

## Testing

- `bundle exec rubocop` ✅
- New tests cover rejection of nil/Array/Hash and acceptance of String/Symbol/Integer/Float

### DIFF
--- a/lib/valkey.rb
+++ b/lib/valkey.rb
@@ -593,9 +593,14 @@ class Valkey
     buffers = []
 
     command_args.each_with_index do |arg, i|
-      arg = arg.to_s # Ensure we convert to string
+      arg = case arg
+            when String, Symbol, Integer, Float
+              arg.to_s
+            else
+              raise TypeError, "Unsupported command argument type: #{arg.class}"
+            end
 
-      buf = FFI::MemoryPointer.from_string(arg.to_s)
+      buf = FFI::MemoryPointer.from_string(arg)
       buffers << buf # prevent garbage collection
       arg_ptrs.put_pointer(i * FFI::Pointer.size, buf)
       arg_lens.put_ulong(i * 8, arg.bytesize)
--- a/test/standalone/valkey_test.rb
+++ b/test/standalone/valkey_test.rb
@@ -24,6 +24,7 @@ class TestStandaloneValkey < Minitest::Test
   include ValkeyTests::URIConnection
   include ValkeyTests::Utils
   include ValkeyTests::FlattenMap
+  include ValkeyTests::ArgTypeValidation
 
   # Property-based test modules for eval/evalsha
   include ValkeyTests::EvalEvalshaBasicProperties
--- a/test/valkey/arg_type_validation_test.rb
+++ b/test/valkey/arg_type_validation_test.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module ValkeyTests
+  # Tests that nil and other unsupported types are rejected with TypeError
+  # instead of being silently coerced to empty string via #to_s (fixes #191).
+  module ArgTypeValidation
+    def test_nil_arg_raises_type_error
+      assert_raises(TypeError) { r.set("key", nil) }
+    end
+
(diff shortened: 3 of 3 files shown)


Aryex:
Replacing with full cherry-pick of 9005f65 which includes the broader fixes (flattening + use-after-free).

PR #197, opened 2026-07-28
fix(ruby): enable TLS for any truthy :ssl value

### LINKED ISSUE #179: Refactor FFI pointer-returning methods to use structs

Aryex:
### Summary

Allow `ssl` to set with truthy values. 

In Ruby, the only `nil` and `false` are falsey, allowing even `ssl: 0` of `ssl: Object.new` to set ssl.

### Issue link

Closes #179 

### Features / Changes

- `lib/valkey.rb`: replace the `[true, "true"].include?(options[:ssl])` allowlist with a truthiness check, `scheme = options[:ssl] ? "rediss" : "redis"`.
- Any truthy `:ssl` value now enables TLS (`rediss`); only `nil`/`false` (or an absent option) select plaintext (`redis`).
- Fail-safe direction: ambiguous-but-truthy input now encrypts rather than silently downgrading.
- Parity: aligns with redis-rb / redis-client (`@ssl = ssl || false`). No public API change — all previously working forms (`true`, `"true"`, `false`, URL-derived boolean) behave the same.

### Testing

- New `test/valkey/ssl_scheme_test.rb`: 11 server-free unit tests that stub the native client creation to capture the constructed URI and assert the scheme across truthy values (`true`, `"true"`, `1`, `"1"`, `"TRUE"`, `"yes"`, a symbol, an object → `rediss`) and falsy/absent (`false`, `nil`, no option → `redis`). No live server or TLS certs required.
- `bundle exec ruby -Itest test/valkey/ssl_scheme_test.rb` → 11 tests, 22 assertions, 0 failures, 0 errors, 0 skips.
- `bundle exec rubocop lib/valkey.rb test/valkey/ssl_scheme_test.rb` → no offenses.

### Checklist

-   [x] Destination branch is correct — **release-1.0** (not main).


test/valkey/ssl_scheme_test.rb
currantw, line 1: Do we need to add an entire new test class for this? Aren't there existing tests that use this? Seems like overkill to me to have an entire test class just for a one-line change – we are essentially just testing a bunch of different "truthy" values, right? I think just one additional test case or modification of an existing one would probably suffice? 🤷

currantw (approved):
Reviewed to ce204c4.

test/valkey/ssl_scheme_test.rb
Aryex: Good point. Folded into existing test suite.

currantw (approved):

PR #198, opened 2026-07-28
fix(ruby): keep OpenTelemetry FFI config reference

Aryex:
### Summary

Fixes an use-after-free isssue `Valkey::OpenTelemetry` init. `build_config`.

### Issue link

Related to #179

### Features / Changes

- `build_config` returns `[config_struct, keep_alive]`; `keep_alive` pins the nested structs and endpoint buffers the config references only by address.
- `init` holds `keep_alive` across the `init_open_telemetry` call, then releases it — the native side copies the endpoints synchronously and retains no pointers.
- Documents the pinning contract on `build_config` and adds a `#179` TODO.

### Testing

- No automated test (a unit test can't deterministically force the use-after-free).
- Manual: stubbed-init smoke check confirms the endpoint is read correctly during the call and init completes.
- `bundle exec rubocop lib/valkey/opentelemetry.rb` → no offenses; `ruby -c` → OK.

### Checklist

-   [ ] Tests are added or updated. (Intentionally omitted — see Testing.)
-   [x] Destination branch is correct — **release-1.0**.

### DIFF
--- a/lib/valkey/opentelemetry.rb
+++ b/lib/valkey/opentelemetry.rb
@@ -45,6 +45,10 @@ class Valkey
     @initialized = false
     @config = nil
     @parent_span_context_provider = nil
+    # Retains the FFI config struct wrappers and endpoint buffers passed to the
+    # native init so their memory isn't GC'd while the native layer holds their
+    # raw addresses. See build_config / init.
+    @config_refs = nil
 
     class << self
       # Initialize OpenTelemetry in the Valkey GLIDE core.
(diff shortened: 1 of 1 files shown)


lib/valkey/opentelemetry.rb
currantw, line 1: Did you consider just inlining the `build_config` logic? It isn't particularly long, and I think the added complexity created by a longer method is less than that created by populating, passing, and understanding this `keep_alive` array?

currantw (commented):
Reviewed to b63ae10.

lib/valkey/opentelemetry.rb
Aryex: I actually think the reverse. Inlining it adds ~50 lines to `init` making it longer to read, and we still need the keep_alive reference.

I think `build_config` clearly shows the intentions of those ~50 lines. You can inspect `build_config` for more details. The `keep_alive` is confusing but I will refactor this in #179  so it is not needed.

currantw: Discussed. Decided that added complexity with `keep_alive` is worth keeping separate method, especially since there are plans to refactor this anyway in the near future.

currantw (approved):
Reviewed to b63ae10.
Looks good ✅

Issue #184, opened 2026-07-28
Password containing `@` (or other URL-reserved characters) fails to authenticate


jamesx-improving:
# Password containing `@` (or other URL-reserved characters) fails to authenticate

## Description

Authentication fails whenever the configured password contains any URL-reserved character (`@ : / ? # % +`, space, or unicode). Both forms — passing `password:` as a keyword arg and embedding it in the connection URL — are affected. A password that works with `redis-rb` (and against the server via `valkey-cli`) is rejected by `valkey-rb` as `AuthenticationFailed`.

## Steps to reproduce

```ruby
require "valkey"

# arrange: set a password with `@` in it
c = Valkey.new
c.call("CONFIG", "SET", "requirepass", "p@ss")
c.close

# act: connect with the exact same password
Valkey.new(password: "p@ss").ping
# => Valkey::CannotConnectError: Received error for address `127.0.0.1:6379`:
#      Password authentication failed- AuthenticationFailed

# Sanity check: valkey-cli connects fine with the same password
# $ valkey-cli -p 6379 -a 'p@ss' ping  =>  PONG
```

## Expected behavior

`Valkey.new(password: "p@ss").ping` returns `"PONG"`, matching `redis-rb` and `valkey-cli -a`. Passwords are transmitted to the server as their raw bytes, regardless of URL-reserved characters.

## Actual behavior

`Valkey::CannotConnectError: … Password authentication failed- AuthenticationFailed`. The server never sees `p@ss` — it sees `p%40ss`.

## Root cause

In `Valkey#initialize` (`lib/valkey.rb` around lines 71–95), the driver constructs a `redis://:CGI.escape(password)@host:port` URI string and hands that string to `Bindings.create_client_from_uri`. The downstream URI parser in `glide-core` does not URL-decode the userinfo — it forwards the encoded bytes to the server as `AUTH`, so `"p@ss"` arrives on the wire as `"p%40ss"` and never matches `requirepass`.

Double-encode proof: set `requirepass "p%40ss"` on the server (literal), then `Valkey.new(password: "p%40ss")` — the driver `CGI.escape`s to `"p%2540ss"`, which also fails. This confirms the value is being escaped once by the driver and never unescaped anywhere in the pipeline.

## Suggested fix

- Pass `password` and `username` as separate fields in the FFI JSON options (the same way `client_name`, `client_az`, etc. are already passed), instead of smuggling them through the URI.
[...]

PR #201, opened 2026-07-29
fix: raise on unparseable Valkey URLs; support valkey:// scheme

### LINKED ISSUE #185: `Valkey.new(url:)` silently connects to 127.0.0.1:6379 when the URL is unparseable
# `Valkey.new(url:)` silently connects to `127.0.0.1:6379` when the URL is unparseable

## Description

When `Valkey.new(url: …)` is passed a URL that the internal regex parser doesn't recognise, the parser returns an empty Hash and `Valkey#initialize` silently falls back to the `host: "127.0.0.1", port: 6379` defaults. The user gets a working client — pointed at localhost — with no warning, no exception, and no log line. In production this is a serious silent misconfiguration: the app looks healthy but is talking to the wrong host.

## Steps to reproduce

```ruby
require "valkey"

# Every one of these silently connects to 127.0.0.1:6379:
Valkey.new(url: "redis://myhost.example.com:6380/")    # trailing slash after empty db
Valkey.new(url: "redis://[::1]:6379/0")                # IPv6 host
Valkey.new(url: "valkey://myhost.example.com:6380/0")  # valkey:// scheme
Valkey.new(url: "not-a-url")                           # obvious garbage

# Verified: all four .ping => "PONG" against a locally-running server,
# even though none of those hosts exist.
```

## Expected behavior

Either the URL parses correctly (preferred for IPv6, trailing slash, and `valkey://`) or `Valkey.new` raises `ArgumentError` with a message pointing at the offending URL. Redis-rb raises for unparseable URLs. Silent fallback to localhost is the worst possible option.

## Actual behavior

`Valkey.new` returns a client connected to `127.0.0.1:6379`. No `warn` line, no exception. `parse_redis_url` just returns `{}` for anything the regex can't match, and `initialize` happily merges an empty hash and reaches for its defaults.

## Root cause

`Utils.parse_redis_url` (`lib/valkey/utils.rb` around lines 237–270) is a hand-rolled regex:

```
%r{\A(redis|rediss)://(?:([^:@]*):([^@]+)@)?([^:/]+)(?::(\d+))?(?:/(\d+))?(?:\?.*)?\z}
```

which cannot match: (a) trailing slash after the db path, (b) IPv6 literal in `[…]`, (c) the `valkey://` / `valkeys://` schemes, (d) query strings before other path components. When it fails to match, the method returns `{}`, and `Valkey#initialize` then reads `options[:host]` (nil, defaulted to `"127.0.0.1"`) and `options[:port]` (nil, defaulted to `6379`).

## Suggested fix

- Switch to `URI.parse` with an allowed-scheme allowlist (`redis`, `rediss`, ideally `valkey`, `valkeys`) and pull username/password/host/port/path out structurally.
- If the parser can't produce a valid `host`, raise `ArgumentError, "invalid Valkey URL: #{url.inspect}"`.
- URL-decode userinfo (related to #184).

## Environment

- Ruby: 4.0.6, arm64-darwin25
- Valkey: 8.1.3 (standalone)

jamesx-improving:
### Summary

`Valkey.new(url: ...)` silently connected to `127.0.0.1:6379` whenever the URL didn't match a narrow hand-rolled regex in `Utils.parse_redis_url` — trailing slashes, IPv6 literals, the Valkey-native `valkey://` scheme, and any unparseable string all fell through to the localhost defaults with no warning or exception. This PR replaces the regex parser with `URI.parse`, raises `ArgumentError` on unparseable URLs / unknown schemes / missing hosts, and accepts the four schemes documented by valkey-cli (`redis`, `rediss`, `valkey`, `valkeys`).

### Issue link

- Closes #185

### Features / Changes

- `lib/valkey/utils.rb` — Rewrote `parse_redis_url` to use `URI.parse`. Accepts `redis`, `rediss`, `valkey`, `valkeys` (matching `valkey-cli -u` semantics — `valkeys` maps to TLS). Raises `ArgumentError` on unparseable URLs, unknown schemes, missing hosts, or non-integer database paths. Error messages redact userinfo (`redis://[REDACTED]@…`) so passwords don't leak into logs, stack traces, or error trackers.
- `lib/valkey.rb` — Wraps host in `[..]` at outgoing URI assembly when it contains a `:`, so IPv6 literals produce an unambiguous URI to the FFI.
- `README.md` — Documented all four accepted schemes and the new raise-on-invalid behavior.
- Note: URL-decoding of userinfo (percent-encoded `@`, `:`, etc. in passwords) is intentionally deferred to #184, which owns that credential-handling fix.

### Testing

**Unit tests (`test/valkey/utils_test.rb`):** replaced the permissive `test_parse_redis_url_invalid_format` with an `assert_raises`, added 9 new tests covering trailing slash, IPv6, `valkey://`, `valkeys://`, unknown scheme, missing host, non-integer db, and credential redaction (both URI-parse failure and post-parse validation paths).

**Integration tests (`test/valkey/uri_connection_test.rb`):** tightened `test_url_with_invalid_scheme` to `assert_raises(ArgumentError)`, added `test_unparseable_url_raises_instead_of_falling_back` (the specific GH #185 regression), `test_url_with_trailing_slash_parses`, and `test_valkey_scheme_url_parses`.

**Local test runs on this branch (base `release-1.0` @ `bf4a67f`):**

- `bundle exec rubocop lib/valkey/utils.rb lib/valkey.rb test/valkey/utils_test.rb test/valkey/uri_connection_test.rb` — 4 files, 0 offenses.
- `bundle exec rake test:standalone` — 787 tests, 4035 assertions, **0 failures**, 4 errors, 102 skips. All 4 errors are `TLS_CERT_DIR is not set` — pre-existing infra requirement, present on the untouched base, unrelated to this diff.
- `bundle exec rake test:cluster` — 787 tests, 2763 assertions, **0 failures**, 4 errors, 277 skips. Same 4 pre-existing TLS_CERT_DIR errors.
- All 24 URL/URI-related tests pass, including every new/tightened test above.

### Checklist

- [x] Documentation is updated (if applicable).
- [x] Destination branch is correct - `release-1.0`

### DIFF
--- a/lib/valkey/utils.rb
+++ b/lib/valkey/utils.rb
@@ -249,7 +249,11 @@ class Valkey
     def self.parse_redis_url(url)
       return {} unless url.is_a?(String) && !url.empty?
 
-      redacted = url.sub(%r{://[^/@]*@}, '://[REDACTED]@').inspect
+      # Redact userinfo before it ends up in error messages, logs, or trackers.
+      # The regex greedily consumes up to the LAST `@` after `://`, so passwords
+      # containing `/` or `@` are still fully redacted (RFC 3986 allows both in
+      # percent-decoded form; either can appear raw in a user-supplied URL).
+      redacted = url.sub(%r{(\A[^:]+://).*@}, '\1[REDACTED]@').inspect
 
       uri = begin
         URI.parse(url)
--- a/lib/valkey/utils.rb
+++ b/lib/valkey/utils.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "uri"
+
 class Valkey
   # Valkey Utils module
   #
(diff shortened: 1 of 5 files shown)


lib/valkey/utils.rb
```diff
@@ -223,48 +225,64 @@ module Utils
...
-      return {} unless match
-
-      scheme = match[1]
-      username = match[2]
-      password = match[3]
-      host = match[4]
-      port = match[5]&.to_i
-      db = match[6]&.to_i
-      ssl = scheme == "rediss"
+      redacted = url.sub(%r{://[^/@]*@}, '://[REDACTED]@').inspect
```
Aryex: If a password contains `/` it won't be redacted.

lib/valkey/utils.rb
```diff
@@ -223,48 +225,64 @@ module Utils
...
     #   # => { host: 'localhost', port: 6379, password: 'secret', db: 15, ssl: false }
     #
-    #   parse_redis_url('rediss://user:secret@localhost:6380/0')
+    #   parse_redis_url('valkeys://user:secret@localhost:6380/0')
     #   # => { host: 'localhost', port: 6380, username: 'user', password: 'secret', db: 0, ssl: true }
+    ALLOWED_URL_SCHEMES = %w[redis rediss valkey valkeys].freeze
+    TLS_URL_SCHEMES = %w[rediss valkeys].freeze
+    private_constant :ALLOWED_URL_SCHEMES, :TLS_URL_SCHEMES
+
     def self.parse_redis_url(url)
```
Aryex, line 249: Could all this URL validation,  not just in this function but also in its caller `Valkey.initialize()`, be handled at the core? We pass the url through the FFI anyways so it should handle validation there as well.

lib/valkey/utils.rb
jamesx-improving: Good catch. Fixed in 883e363 — replaced the redaction regex with one that consumes everything from `://` up to the last `@`, so passwords containing `/`, `@`, or other reserved chars are still fully redacted. Added regression tests for both cases in `test_parse_redis_url_error_message_redacts_credentials`.

lib/valkey/utils.rb
jamesx-improving: Fair point — you're right that this is where it should ultimately live, but I don't think it should happen in this PR. A few reasons:

(1) Ruby's `Valkey#initialize` doesn't currently pass the raw URL through. It re-assembles a new URI from the parsed options (merging `host`/`port`, swapping scheme based on `options[:ssl]`, and CGI-escaping credentials — which is itself buggy per #184). Deferring URL validation to core means dismantling that assembly block, which is a larger refactor and depends on #184's resolution (that PR is gated on [valkey-glide#6660](https://github.com/valkey-io/valkey-glide/pull/6660), which percent-decodes userinfo in `create_client_from_uri` — opened yesterday).

(2) The cross-binding audit in valkey-glide#6660 confirms `create_client_from_uri` is called only by valkey-glide-ruby today — so investing in it as the URL-handling layer is the right direction, just larger in scope.

(3) The user-facing contract in #185 is specifically "raise `ArgumentError` at construction, like redis-rb does". Core-side validation surfaces via `ConnectionResponse`'s error field as `CannotConnectError` at connect time — different class, later timing. Even if we route URLs through core, we'd still want a thin Ruby-side pre-check to preserve construction-time `ArgumentError`.

Opened #204 to track the eventual migration once #6660 and #184 land. WDYT?

Aryex (approved):

Issue #216, opened 2026-07-31
RC3 BugBash Critical Issues


Aryex:
**All 13 verified reproducible on `v1.0.0-rc3` (commit `fb2fa46`).** Sub-issues filed for the highest-signal defects — see linked issue numbers per row.

---

## Group 1 — Scripting API rewrite (`lib/valkey/commands/scripting_commands.rb`)

Single-file rewrite closes both defects. `script_load`/`eval`/`evalsha` must dispatch real wire commands via `call("SCRIPT","LOAD",…)` / `call("EVAL",…)` / `call("EVALSHA",…)` instead of going through glide-core's local `Script`/`invoke_script` container.

| ID | Severity | Issue | Assignee | Brief |
|---|---|---|---|---|
| R2-12+13 | critical | #213 | @jamesx-improving  | `script_load` never sends `SCRIPT LOAD`; `script_flush` leaves local container populated so subsequent `evalsha` silently auto-uploads. Cross-worker `EVALSHA` by SHA is broken; `SCRIPT EXISTS` health checks return false. |
| R3-4     | critical | #213 | @jamesx-improving | `eval`/`evalsha` silently corrupt KEYS/ARGV when called with the integer key-count form (`eval(src, 1, "mykey", "myarg")`) — script runs against the wrong key with no error. |

---

## Group 2 — Blocking-list command family (`lib/valkey/commands/list_commands.rb`)

All three symptoms trace to the same incomplete conversion. Route blocking commands through `send_command` (glide-core handles the blocking timeout natively) and fix the Symbol-in-args pattern.

| ID | Severity | Issue | Assignee | Brief |
|---|---|---|---|---|
| R2-2   | critical | #220   | @Aryex  | `blpop`/`brpop`/`brpoplpush` call undefined `send_blocking_command` helper. Also `blpop` passes Symbol `:blpop` where an Integer `RequestType` is required. |
| R2-6   | critical | #220| @Aryex | `blmove` ships stray `:blmove` Symbol as an argv token → server rejects with `wrong number of arguments`. |
| R3-2a  | critical | #220 | @Aryex | `brpop` flattens away its trailing `timeout:` Hash — timeout silently becomes 0, blocks forever. |

---

## Group 3 — Deprecated commands to remove

[...]

Issue #137, opened 2026-07-08
Add batch-level routing to pipelined/multi


Aryex:
## Summary

Add an optional `route:` keyword to `pipelined` and `multi` so callers can route an entire batch
to a specific cluster node. Today Ruby has no way to do this — the underlying `BatchOptionsInfo`
carries a `route_info` field but Ruby hardcodes it to `NULL`.

## Proposal

```ruby
client.pipelined(route: Route.slot_key("some-key")) do |pipe|
  pipe.set("a", "1")
  pipe.set("b", "2")
end

client.multi(route: Route.slot_key("some-key")) do |tx|
  tx.incr("counter")
end
```

- `route:` defaults to `nil` (today's behavior, unchanged).
- `route:` is batch-level, not per-command — attaching it to `pipe.get(...)` inside the block is
  not supported. This matches the peer GLIDE clients (Python/Node/Java/Go) and the protobuf wire
  format, which has no per-command `Routes` field inside a `Batch`.
- Multi-node routes (`Route.all_nodes`/`Route.all_primaries`) are rejected with `ArgumentError`
  before any FFI call. The peer clients type this parameter as single-node only.
- Wire the option into both dispatch paths — the normal `Bindings.batch` call and the
  MULTI/EXEC/DISCARD sequential fallback — or routing silently drops for batches containing
  transactional commands.

## Match existing behavior

Behavior should align with:

- **redis-rb** — `pipelined`/`multi` API shape and semantics for atomic vs non-atomic batches.
- **Peer GLIDE clients** — reference implementation is Python sync
  (`python/glide-sync/glide_sync/sync_commands/cluster_commands.py` `ClusterCommands.exec` and
  `python/glide-shared/glide_shared/commands/batch_options.py` `ClusterBatchOptions.route`).
  Node/Java/Go expose the same shape via `ClusterBatchOptions.route` / `WithRoute`.

PR #218, opened 2026-08-01
fix: pipeline route pass through

Aryex:
### Summary

Commands that accept a `route:` kwarg raised `ArgumentError: unknown keyword: :route` when queued inside `pipelined`/`multi`. They now queue without error and the `route` is ignored. Routing support for Batch is not yet implemented.

### Issue link

Related to #216 but a separate issues found. 

### Testing

- `bundle exec rubocop` — 99 files, no offenses.
- Verified against a live standalone server: `pipelined { set("rk","1"); ping(route: Valkey::Route.all_nodes); get("rk") }` returns `["OK", "PONG", "1"]` instead of raising, and a direct `send_command(RequestType::PING, [], route: ...)` inside `pipelined` queues cleanly.

### Checklist

-   [ ] Tests are added or updated.
-   [x] Destination branch is correct - release-1.0

### DIFF
--- a/lib/valkey/pipeline.rb
+++ b/lib/valkey/pipeline.rb
@@ -15,13 +15,14 @@ class Valkey
     end
 
     # `route:` is accepted and ignored. Routing is not supported for Pipeline yet.
+    # rubocop:disable Lint/UnusedMethodArgument
     def send_command(command_type, command_args = [], route: nil, &block)
-      # rubocop:enable Lint/UnusedMethodArgument
       @commands << [command_type, command_args, block]
       future = Future.new(command_type, command_args)
       @futures << future
       future
     end
+    # rubocop:enable Lint/UnusedMethodArgument
 
     # @api private - called by Valkey#pipelined / the block form of #multi
     # once send_batch_commands' final results are available. Purely
--- a/lib/valkey/pipeline.rb
+++ b/lib/valkey/pipeline.rb
@@ -14,7 +14,9 @@ class Valkey
       @in_multi = false
     end
 
-    def send_command(command_type, command_args = [], &block)
+    # `route:` is accepted and ignored. Routing is not supported for Pipeline yet.
+    def send_command(command_type, command_args = [], route: nil, &block)
+      # rubocop:enable Lint/UnusedMethodArgument
       @commands << [command_type, command_args, block]
       future = Future.new(command_type, command_args)
       @futures << future


Sasidharan3094 (approved):

yipin-chen (approved):

jamesx-improving (approved):
## Verdict: Approve — 0 high-confidence findings

Small, targeted fix that adds `route:` as an accepted-and-ignored keyword to `Valkey::Pipeline#send_command` so `pipelined` / `multi` blocks stop raising `ArgumentError` when a command that forwards `route:` is queued. The change is correct, well scoped, matches the PR description, and preserves existing behavior. LGTM — the rubocop-pair + comment make the "accepted and ignored" intent obvious.

### Verification performed

- Read the whole file `lib/valkey/pipeline.rb`; only `send_command`'s signature changed. `resolve_futures!` / `abort_futures!` still receive well-formed `[type, args, block]` tuples and are unaffected.
- Confirmed the `rubocop:disable` / `rubocop:enable` pair brackets only the changed method (post-fixup commit `e0a7146`).
- Confirmed all in-tree call sites pass `route:` as a keyword argument, never positionally: ~39 call sites under `lib/valkey/commands/`, every one uses `route: route`. No positional collision with the existing `command_args = []` parameter.
- Confirmed the top-level `Valkey#send_command` (`lib/valkey.rb:355`) already has the same `route: nil` keyword, so the pipeline signature now matches the sibling method's shape — cleaner substitutability.
- Traced the failure mode this PR fixes: `Valkey::Pipeline` includes `Commands`, so `pipe.ping(route: X)` ends up calling `Pipeline#send_command(RequestType::PING, [...], route: X, &block)`. Prior to this PR, `send_command` did not accept the kwarg, raising `ArgumentError: unknown keyword: :route`. Now it accepts and discards it.
- Checked for any caller that intentionally relies on `route:` raising inside a pipeline — none found in `lib/` or `test/`.

### Non-blocking notes

**1. No test covers the new signature** — `test/valkey/pipeline_test.rb`

The PR's "Tests are added or updated" checkbox is unchecked and the verification given is a manual pipelined+PING run. Existing unit tests in `test/valkey/pipeline_test.rb` do not exercise `pipeline.send_command(..., route: ...)`, and cluster routing tests in `test/cluster/cluster_routing_test.rb` all go through the non-pipelined path. A ~3-line regression test — call `pipeline.send_command(Valkey::RequestType::PING, [], route: Valkey::Route.all_nodes)` and assert it returns a `Valkey::Future` and queues the command — would prevent someone reintroducing the raise without needing a running server. Consider a second test asserting the queued `command_args` shape is unchanged when `route:` is passed.

**2. Silent ignore may surprise users** — `lib/valkey/pipeline.rb:19`

A user writing `pipelined { |p| p.ping(route: Route.all_nodes) }` reasonably expects PING to be sent to all nodes; with this PR the argument is silently dropped and PING follows the pipeline's default routing. This is documented in the added comment and in the PR body, and is a strict improvement over the pre-PR raise, so it's not a blocker. Worth surfacing so it doesn't get lost between now and full batch-routing support (referenced issue #137 / #216) — perhaps a follow-up that emits a one-time `warn` the first time a non-nil `route:` is dropped, or that raises `NotImplementedError` for non-nil routes so users get an explicit "not yet supported" instead of a silent behavior change. No action required for this PR; just a follow-up decision (warn vs. raise vs. keep silent) tracked against the batch-routing issue.

PR #219, opened 2026-08-01
Withhold server-module commands: JSON, FT, MODULE

Aryex:
## Summary

Current Modules API is withheld to avoid confusion as Module support is planned for post GA. 

## What changes

Removed from the `Commands` mixin in `lib/valkey/commands.rb`:

```ruby
include ModuleCommands       # 5 methods
include JsonCommands         # 22 methods
include VectorSearchCommands # 14 methods
```

json_get is not available in valkey-glide-rb 1.0.0: server-module commands are
withheld pending fixes. Use the generic command interface instead,
e.g. call("JSON.GET", ...).
```

**Module commands remain fully reachable** through the generic interface, which is unaffected:

```ruby
valkey.call("JSON.GET", "user:1", "$.name")
valkey.call("FT.SEARCH", "idx", "@title:ruby")
valkey.call("MODULE", "LIST")
```

## Also in this PR

Disables the `module-tests` CI job (`if: false`). With the test classes commented out, its `-n '/TestStandalone(Module|Json|VectorSearch)Commands/'` selector matches **zero tests and the job still exits 0** — green while asserting nothing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -277,6 +277,8 @@ jobs:
     needs: [build-native-library]
     runs-on: ubuntu-latest
     timeout-minutes: 25
+    # Disabled for now until Module support is added.
+    if: false
     steps:
       - name: Check out code
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
--- a/test/cluster/cluster_commands_test.rb
+++ b/test/cluster/cluster_commands_test.rb
@@ -55,26 +55,29 @@ class TestClusterGeoCommands < Minitest::Test
   include Lint::GeoCommands
 end
 
-class TestClusterJsonCommands < Minitest::Test
-  include Helper::Cluster
-  include Lint::JsonCommands
-end
-
-class TestClusterModuleCommands < Minitest::Test
-  include Helper::Cluster
-  include Lint::ModuleCommands
-end
-
-class TestClusterVectorSearchCommands < Minitest::Test
-  include Helper::Cluster
-  include Lint::VectorSearchCommands
-end
-
 # ValkeyTests modules with setup/teardown
 class TestClusterFunctionCommands < Minitest::Test
   include Helper::Cluster
   include ValkeyTests::FunctionCommands
 end
 
-# NOTE: Module tests (JsonCommands, ModuleCommands, VectorSearchCommands)
-# require modules loaded on all cluster nodes
+# Server-modules support is not yet added. Re-enable when implemented.
+#
+# The lint files are still required by test_helper, so they stay syntax-checked
+# and ready to re-enable alongside the lib includes. Note these also require the
+# modules to be loaded on all cluster nodes:
+#
+#   class TestClusterJsonCommands < Minitest::Test
+#     include Helper::Cluster
+#     include Lint::JsonCommands
+#   end
+#
+#   class TestClusterModuleCommands < Minitest::Test
+#     include Helper::Cluster
+#     include Lint::ModuleCommands
+#   end
+#
+#   class TestClusterVectorSearchCommands < Minitest::Test
+#     include Helper::Cluster
+#     include Lint::VectorSearchCommands
+#   end
--- a/test/cluster/cluster_commands_test.rb
+++ b/test/cluster/cluster_commands_test.rb
@@ -62,6 +62,8 @@ class TestClusterFunctionCommands < Minitest::Test
 end
 
 # Server-modules support is not yet added. Re-enable when implemented.
+# TODO: https://github.com/valkey-io/valkey-glide-ruby/issues/233
+#       https://github.com/valkey-io/valkey-glide-ruby/issues/234
 #
 # The lint files are still required by test_helper, so they stay syntax-checked
 # and ready to re-enable alongside the lib includes. Note these also require the
--- a/test/standalone/commands_test.rb
+++ b/test/standalone/commands_test.rb
@@ -36,17 +36,22 @@ class TestStandaloneGeoCommands < Minitest::Test
   include Lint::GeoCommands
 end
 
-class TestStandaloneJsonCommands < Minitest::Test
-  include Helper::Client
-  include Lint::JsonCommands
-end
-
-class TestStandaloneModuleCommands < Minitest::Test
-  include Helper::Client
-  include Lint::ModuleCommands
-end
-
-class TestStandaloneVectorSearchCommands < Minitest::Test
-  include Helper::Client
-  include Lint::VectorSearchCommands
-end
+# Server-modules support is not yet added. Re-enable when implemented.
+#
+# The lint files are still required by test_helper, so they stay syntax-checked
+# and ready to re-enable alongside the lib includes:
+#
+#   class TestStandaloneJsonCommands < Minitest::Test
+#     include Helper::Client
+#     include Lint::JsonCommands
+#   end
+#
+#   class TestStandaloneModuleCommands < Minitest::Test
+#     include Helper::Client
+#     include Lint::ModuleCommands
+#   end
+#
+#   class TestStandaloneVectorSearchCommands < Minitest::Test
+#     include Helper::Client
+#     include Lint::VectorSearchCommands
+#   end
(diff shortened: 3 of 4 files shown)


Sasidharan3094 (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -277,6 +277,8 @@ jobs:
     needs: [build-native-library]
     runs-on: ubuntu-latest
     timeout-minutes: 25
+    # Disabled for now until Module support is added.
+    if: false
```
currantw, line 281: > The requires are retained, so the implementation files stay loaded and syntax-checked. All 41 method names are registered in a new UNSUPPORTED_COMMANDS table that raises Valkey::CommandError naming the generic-command equivalent, so anyone upgrading from 0.9.x or an earlier rc gets an actionable message rather than NoMethodError:

Would it be better to just remove completely? This may help people upgrading, but doesn't it lock us into the specific interfaces for the module commands? i.e. if we want to change the interfaces in the future, is this technically a breaking change?

test/cluster/cluster_commands_test.rb
```diff
@@ -55,26 +55,29 @@ class TestClusterGeoCommands < Minitest::Test
...
-
-class TestClusterModuleCommands < Minitest::Test
-  include Helper::Cluster
-  include Lint::ModuleCommands
-end
-
-class TestClusterVectorSearchCommands < Minitest::Test
-  include Helper::Cluster
-  include Lint::VectorSearchCommands
-end
```
currantw, line 71: Rather than delete, do we want to comment out and link to future issues with TODO comment(s). If issues for these don't exist, probably a good idea to add them?

currantw: Ahhh, I see that you've just moved them down! Good, but I still think we need TODO comments. As a general rule, I don't think we should have any TODO/in-progress/future-work in the code base unless it is specifically linked to an issue.

test/standalone/commands_test.rb
```diff
@@ -36,17 +36,22 @@ class TestStandaloneGeoCommands < Minitest::Test
...
-end
-
-class TestStandaloneVectorSearchCommands < Minitest::Test
-  include Helper::Client
-  include Lint::VectorSearchCommands
-end
+# Server-modules support is not yet added. Re-enable when implemented.
+#
+# The lint files are still required by test_helper, so they stay syntax-checked
+# and ready to re-enable alongside the lib includes:
```
currantw, line 44: Same comment. Let's add TODO(s) linking to issue(s)

currantw (commented):
Reviewed to 13d764b.

jamesx-improving (commented):
## Verdict: Comment

The include/CI-disable diff itself looks clean, but the shipped code doesn't match the PR body: there's no `UNSUPPORTED_COMMANDS` table and the `%i[]` block at `lib/valkey/commands.rb:59` is still empty, so callers of the 41 removed methods will hit `NoMethodError` instead of the actionable `Valkey::CommandError` you describe. Either land that table or update the description — happy to re-review either way.

## Summary

The code change itself (dropping three `include`s and disabling the CI job) is small and internally consistent, but the shipped diff does not match what the PR description promises: there is no `UNSUPPORTED_COMMANDS` table and no actionable `CommandError` for the 41 removed methods — callers will hit a raw `NoMethodError`. Either land the table the description advertises, or update the description.

## Findings

### 1. PR description promises an `UNSUPPORTED_COMMANDS` table that is not in the diff

**Category:** Correctness &middot; **Severity:** Medium
**Location:** `lib/valkey/commands.rb:56-63`

**Problem:** The PR body says "All 41 method names are registered in a new `UNSUPPORTED_COMMANDS` table that raises `Valkey::CommandError` naming the generic-command equivalent, so anyone upgrading from 0.9.x or an earlier rc gets an actionable message rather than `NoMethodError`." That table does not exist in this diff. The block at `commands.rb:56-63` is still `%i[].each do |command| ... end` (empty symbol array), and `grep -rn UNSUPPORTED_COMMANDS lib/` returns no matches. The three `include`s are removed and the three `require`s are retained, but no substitute methods are ever defined on `Valkey::Commands`. Since there is no `method_missing` either, the previously-supported method names now hit Ruby's default lookup and raise `NoMethodError`, not the documented `Valkey::CommandError`.

**Failure scenario:** A user upgrades their `valkey-glide-rb` gem from `0.9.x` (or an earlier `1.0.0.rcN`) that included `json_get`/`module_list`/`ft_search` to this build. Existing code such as `valkey.json_get("user:1", "$.name")` raises `NoMethodError: undefined method 'json_get' for #<Valkey>` instead of the promised `Valkey::CommandError: json_get is not available in valkey-glide-rb 1.0.0: ... Use call("JSON.GET", ...) instead.` The migration path the release notes will presumably point at ("grep for the CommandError string") will not fire.

**Suggested fix:** Pick one of two paths and land it before merging:

(a) Implement the described table — e.g. keep the `%i[...].each { |command| define_method(command) { |*_args, **_kw| raise CommandError, "#{command} is not available in valkey-glide-rb 1.0.0: server-module commands are withheld pending fixes. Use the generic command interface instead, e.g. call(...)" } }` pattern that is already scaffolded in this file, populated with the 41 removed method names (5 module + 22 json + 14 vector search); or

(b) Update the PR description to reflect that the shipped fix is simply removing the includes (`NoMethodError` is the intended UX) and note that in `CHANGELOG`/release notes so downstream users are not surprised.

_Note: the generic `call` path is still fully wired (`lib/valkey/commands/generic_commands.rb:560` uses `RequestType::CUSTOM_COMMAND`), so `valkey.call("JSON.GET", ...)` continues to work regardless of which path is chosen._

## Verified and dropped

- **CI job disable is stronger than the description implies (not a bug).** The PR body describes the risk of `-n '/TestStandalone(Module|Json|VectorSearch)Commands/'` matching zero tests and exiting green while asserting nothing. The actual change adds `if: false` at `.github/workflows/ci.yml:281`, which skips the job entirely rather than running a no-op — this is the safer of the two behaviors, so no follow-up needed.
- **Requires kept for syntax-check are safe.** `lib/valkey/commands.rb:15,17,20` still `require` the three module files, but nothing calls into their methods once the `include`s are gone. No dead-code hazard, and the modules stay syntax-checked as the PR body claims.
- **Test-file comment-out is inert.** The commented blocks in `test/cluster/cluster_commands_test.rb` and `test/standalone/commands_test.rb` reference `Lint::JsonCommands`, `Lint::ModuleCommands`, and `Lint::VectorSearchCommands`, which `test_helper.rb:36-38` still requires via glob. That keeps them parseable and ready to re-enable; no test collection currently references them, so no dangling minitest classes.
- **Pipeline path is fine.** `lib/valkey/pipeline.rb:5` also does `include Commands`, and the same three methods are removed there for the same reason — consistent with the client itself, no divergence.
- **No other in-repo callers.** Grep for `json_get`/`module_list`/`ft_search`/`module_load`/`ft_create` outside `test/lint/` and the three (still-`require`d) command files returns nothing — examples, README, and other command modules do not depend on the removed methods.

.github/workflows/ci.yml
Aryex: Yes, this PR removes all modules from the public API by removing the includes. Unless you meant something else?

test/standalone/commands_test.rb
Aryex: Updated in [32f5e31](https://github.com/valkey-io/valkey-glide-ruby/pull/219/commits/32f5e31ce815d0ebedf6c3a027adc15f44b11c0b)

test/cluster/cluster_commands_test.rb
Aryex: Updated in [32f5e31](https://github.com/valkey-io/valkey-glide-ruby/pull/219/commits/32f5e31ce815d0ebedf6c3a027adc15f44b11c0b)

Aryex:
@jamesx-improving I updated the descriptions. The `UNSUPPORTED_COMMANDS` mention was a mistake.

test/cluster/cluster_commands_test.rb
currantw: Do you also need an issue for the Module commands?

currantw (approved):
Looks good ✅

test/cluster/cluster_commands_test.rb
Aryex: Not at this moment. Other clients don't implement it.

PR #224, opened 2026-08-04
fix: raise `ConnectionError` instead of segfaulting on a closed client

### LINKED ISSUE #212: Using a client after `close` segfaults the VM — `pipelined` / `multi` / `eval` have no closed-connection guard
# Using a client after `close` segfaults the VM — `pipelined` / `multi` / `eval` have no closed-connection guard

## Summary

`send_command` validates the connection pointer before use (`lib/valkey.rb:360`) and raises a clean `Valkey::ConnectionError`. Three other paths into the native layer do not:

- `send_batch_commands` → `Bindings.batch` (`lib/valkey.rb:561`) — reached by `pipelined` and `multi`
- `invoke_script` → `Bindings.invoke_script` (`lib/valkey/commands/scripting_commands.rb:289`) — reached by `eval`, `evalsha`, and the `_ro` variants

Both dereference `@connection` after `close` set it to `nil`, and the process dies with `[BUG] Segmentation fault at 0xfffffffffffffff0`. This needs **no concurrency at all** — a single-threaded script reproduces it deterministically.

Separately, `close` itself (`lib/valkey.rb:296`) is a check-then-act on `@connection` with no mutex, so a thread closing the client while another is mid-command is a use-after-free.

## Scope

This covers **both halves of the close-safety defect** tracked in #216 Group 6:

- **R3-10b** — the *sequential* path: `pipelined` / `multi` / `eval` after `close` crash with no concurrency at all. This is what the reproduction below shows.
- **R3-10** — the *concurrent* path: a thread calling `close` while another is mid-command (TOCTOU on `@connection`), plus concurrent double-`close`.

One fix closes both, because the guard helper reads the pointer into a local before use, which collapses the race window as well as covering the sequential case. Verified: `pipelined` racing `close` segfaults 3/3 on `release-1.0` @ `fb2fa46` and survives 5/5 with the guard applied; 8 threads racing `close` across 20 clients show no double-free.

## Steps to reproduce

**Single-threaded, no race — 3/3 crashes:**

```ruby
require "valkey"

c = Valkey.new
c.set("k", "v")
c.close

c.get("k")                    # OK: raises Valkey::ConnectionError: the client is closed
c.pipelined { |p| p.get("k") } # CRASH
```

```
lib/valkey.rb:561: [BUG] Segmentation fault at 0xfffffffffffffff0
ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [arm64-darwin25]

-- Control frame information ------------------------------------------------
c:0005 p:---- s:0048 e:000047 l:y b:0001 CFUNC  :batch
c:0004 p:0265 s:0038 e:000037 l:y b:0001 METHOD lib/valkey.rb:561
```

Exit status 139 on all three runs. `multi { … }` crashes at the same site. `eval` crashes at `scripting_commands.rb:289`:

```ruby
c = Valkey.new
c.close
c.eval("return 1")
# scripting_commands.rb:289: [BUG] Segmentation fault at 0xfffffffffffffff0
```

**Shared client closed mid-command — 3/3 crashes:**

```ruby
require "valkey"

c = Valkey.new
t = Thread.new { 2000.times { c.pipelined { |p| p.get("k") } rescue nil } }
sleep 0.05
c.close
t.join
```

Also exit 139, same fault address. Note the `rescue nil` — there is nothing to rescue; the process is gone.

The `send_command` path survives this same race (its guard catches the `nil`), which is why the crash is specific to the batch and script paths.

## Expected behavior

Every entry point into the native layer raises `Valkey::ConnectionError, "the client is closed"` once the client is closed, exactly as `send_command` already does. A closed client should never be able to take down the VM.

## Actual behavior

`0xfffffffffffffff0` is `NULL - 16` — the native call reads through the nulled-out pointer. There is no Ruby-level exception, so nothing can rescue it and no `ensure` block runs; a web worker dies mid-request.

## Affected code

```ruby
# lib/valkey.rb:296 — check-then-act; the TOCTOU half (R3-10)
def close
  return if @connection.nil? || @connection.null?

  Bindings.close_client(@connection)
  @connection = nil
end
```

```ruby
# lib/valkey.rb:561 — no guard before the native call
res = Bindings.batch(@connection, 0, batch_info, exception, batch_options.to_ptr, span_ptr)
```

```ruby
# lib/valkey/commands/scripting_commands.rb:289 — same
res = Bindings.invoke_script(@connection, 0, sha, ...)
```

Compare with the guard that already exists at `lib/valkey.rb:360`:

```ruby
raise ConnectionError, "the client is closed" if @connection.nil? || @connection.null? || @connection.address.zero?
```

## Suggested fix

Two parts, and the first is worth doing on its own since it removes the crash entirely:

**1. Extract the existing guard and apply it at every native call site.**

```ruby
# lib/valkey.rb
private def connection!
  conn = @connection
  raise ConnectionError, "the client is closed" if conn.nil? || conn.null?

  conn
end
```

Then use `connection!` in place of `@connection` in `send_command`, `send_batch_commands`, and `invoke_script`. Reading into a local also narrows the race window, since a concurrent `close` can no longer null the reference between the check and the call.

**2. Make `close` idempotent — without a mutex.**

```ruby
def close
  conn = @connection
  @connection = nil

  Bindings.close_client(conn) unless conn.nil? || conn.null?
end
```

Clear the pointer first, then free. `close` becomes idempotent (a second call sees `nil` and no-ops), and combined with part 1 no command can dereference a freed handle.

**Do not use a `Mutex` here.** An earlier version of this issue proposed guarding `close` with `@close_mutex.synchronize`. That is wrong: `Mutex#synchronize` raises `ThreadError: can't be called from trap context`, so it breaks the standard graceful-shutdown idiom —

```ruby
Signal.trap("TERM") { client.close }   # works today; would raise with a mutex, leaking the handle
```

Verified: this idiom succeeds on `release-1.0` and with the lock-free fix above, and raises `ThreadError` if a mutex is introduced. Concurrent double-`close` was also **not** reproducible without one (8 threads × 20 clients, clean), so a mutex would trade a working mainstream path for protection against a race that does not occur.

**In-flight commands need no extra work.** An earlier version of this issue claimed a residual window — a command still executing in the native layer while another thread frees the pointer — and suggested it needed refcounting as a follow-up. That was wrong: glide-ffi already refcounts. Every command entry point does `Arc::increment_strong_count(client_adapter_ptr)` before use (`ffi/src/lib.rs:3392` and the sibling entry points), and `close_client` only decrements (`ffi/src/lib.rs:2513`, whose own comment notes the count reaches zero "once all client requests are done"). So an in-flight request holds its own reference and the native `ClientAdapter` outlives it.

Verified: 12 threads holding 4-second blocking calls open with `close` fired 0.6s in survives on both `release-1.0` and this fix. No follow-up issue is needed for this.

## Test coverage

No test currently calls a command after `close`. Suggested:

```ruby
def test_pipelined_after_close_raises
  c = _new_client
  c.close
  assert_raises(Valkey::ConnectionError) { c.pipelined { |p| p.get("k") } }
end

def test_multi_after_close_raises
  c = _new_client
  c.close
  assert_raises(Valkey::ConnectionError) { c.multi { |m| m.get("k") } }
end

def test_eval_after_close_raises
  c = _new_client
  c.close
  assert_raises(Valkey::ConnectionError) { c.eval("return 1") }
end

def test_close_is_idempotent
  c = _new_client
  c.close
  c.close # must not crash or double-free
end
```

These fail by crashing the test process today rather than by failing an assertion, so they're worth adding before the fix to confirm the reproduction.

## Impact

A shared client instance is the standard shape in Rails, Puma, and Sidekiq deployments, and `pipelined` / `multi` are mainstream API. Any shutdown path, connection-recycling path, or reconnect-after-error path that closes a client while another fiber or thread is using it can hard-kill the process instead of raising something catchable. JRuby, which the README lists as supported, has no GVL and so widens the race window further.

## Environment

- Ruby: 4.0.6, arm64-darwin25
- Valkey: 8.1.3 (standalone)

jamesx-improving:
### Summary

Calling `pipelined`, `multi`, `eval`, `evalsha` or `invoke_script` on a client that has already been closed hard-killed the Ruby VM with `SIGSEGV` (exit 139) instead of raising a rescuable error. `Bindings.batch` and `Bindings.invoke_script` passed `@connection` straight into the native layer with no nil/null guard — only `send_command` had one — so the core dereferenced a NULL pointer. This adds a single `connection!` guard used by every native call site that takes the client handle, and makes `close` idempotent.

**Two things a reviewer should read before commenting — both contradict an earlier revision of the issue:**

**1. The mutex was rejected deliberately.** An earlier revision of #212 (and the #216 Group 6 row for R3-10) prescribed "add a mutex around `@connection` access". This PR does **not** do that, on purpose: `Mutex#synchronize` raises `ThreadError: can't be called from trap context`, which breaks `Signal.trap("TERM") { client.close }` — the standard graceful-shutdown idiom — and leaks the native handle when it raises. Verified: that idiom works on `release-1.0` and with this fix, and raises `ThreadError` as soon as a mutex is introduced. `test_close_works_in_trap_context` locks the behavior in. Idempotency comes from clear-then-free ordering in `close`, not from a lock.

**2. No follow-up is needed for in-flight commands.** An earlier revision of the issue claimed a residual window requiring refcounting. That is wrong — glide-ffi already refcounts: every command entry point does `Arc::increment_strong_count` before `Arc::from_raw`, and `close_client` only decrements (`ffi/src/lib.rs:2513`, whose own upstream comment notes the count reaches zero "once all client requests are done"). Client creation uses `Arc::into_raw`, so the raw pointer owns exactly one strong ref that `close_client` consumes. The native `ClientAdapter` therefore outlives any request still executing. There is no follow-up task here.

### Issue link

Closes #212

This closes **both halves of #216 Group 6**: R3-10b (the sequential close-then-use crash) and R3-10 (the concurrent close-vs-command TOCTOU plus concurrent double-close). One fix covers both, because `connection!` reads the handle into a local before use — collapsing the race window as well as covering the sequential path. Both halves are now test-backed rather than only probe-verified.

### Features / Changes

-   `lib/valkey.rb` — new private `connection!` helper: reads `@connection` into a local and raises `Valkey::ConnectionError, "the client is closed"` if it is nil or null, otherwise returns the local. Message matches the sibling GLIDE clients (Go `ClosingError`, Node/Java `ClosingException`).
-   `lib/valkey.rb` — `send_command` and `send_batch_commands` now go through `connection!` and pass the returned local to the native call. In `send_batch_commands` the check happens before any FFI memory is allocated, so a closed client fails fast.
-   `lib/valkey/commands/scripting_commands.rb` — `invoke_script` does the same, likewise before allocating FFI buffers. This covers `eval`, `evalsha`, and the `_ro` variants, which all funnel through it.
-   `lib/valkey.rb` — `close` is idempotent by construction: `@connection` is cleared *before* the handle is freed, so a second (or concurrent) `close` sees nil and does nothing instead of double-freeing.
-   `test/valkey/connection_lifecycle_test.rb` — 9 new tests.

**Completeness:** all 22 `attach_function` declarations in `lib/valkey/bindings.rb` were audited. Exactly 5 take the client handle — `close_client`, `command`, `command_with_route_info`, `batch`, `invoke_script` — and all 5 are now guarded. Before this change only `send_command` (`command` / `command_with_route_info`) had a guard, leaving `batch` and `invoke_script` exposed.

Out of scope: `p.eval` inside a `pipelined` block raises `NoMethodError`, a pre-existing wart unaffected by this change (only the message text differs, `build_command_args` -> `connection!`).

### Testing

Against a dedicated Valkey 8.1.3 on port 6412, Ruby 4.0.6 arm64-darwin25.

-   `bundle exec rubocop` — 99 files inspected, **0 offenses**.
-   Baseline, pristine `fb2fa46`: **840 tests, 4155 assertions, 0 failures, 0 errors, 100 skips**.
-   With this change: **849 tests, 0 failures, 0 errors, 100 skips**, reproduced twice. 840 -> 849 is exactly the 9 new tests; skip count identical.
    ```
    CI=1 SKIP_TLS_TESTS=true VALKEY_PORT=6412 bundle exec rake test:standalone
    ```

New tests: closed-client raise on `pipelined`, `multi`, `eval`, `evalsha` and `invoke_script`; `close` idempotent; concurrent `close` frees the handle once; `close` racing in-flight commands does not crash; `close` works in a trap context.

**Negative controls — the tests actually catch the bug:**

-   With `lib/` reverted and the new tests kept, the guard tests **segfault (exit 139)** — precisely the crash this PR fixes.
-   `test_close_racing_in_flight_commands_does_not_crash` fails 3/3 without the fix, clean 3/3 with it.
-   `pipelined` racing `close` survives 5/5 with the fix vs. segfaulting 3/3 without.

**Deferred:** cluster-mode tests were not run (no local cluster available). The new tests are `skip`-guarded on `cluster_mode?`, so they are inert there; the change itself is mode-independent. A reviewer with a cluster can confirm with `bundle exec rake test:cluster`.

### Checklist

### DIFF
--- a/lib/valkey.rb
+++ b/lib/valkey.rb
@@ -272,6 +272,12 @@ class Valkey
     end
 
     @connection = res[:conn_ptr]
+    # Serializes `close` so two threads cannot both capture the handle and
+    # double-decrement its Arc refcount (issue #212). We use `Mutex#try_lock`
+    # in `close`, not `#synchronize`, so trap-context callers still work:
+    # only `#lock`/`#synchronize` raise ThreadError from a trap, `#try_lock`
+    # returns false and moves on.
+    @close_lock = Mutex.new
     Bindings.free_connection_response(response_ptr)
 
     # Store cluster mode flag for response handling (MAP returns Hash in cluster, Array in standalone)
@@ -292,25 +298,35 @@ class Valkey
     @in_multi_block = false
   end
 
-  # Closes the client and frees the native connection. Idempotent: the handle is
-  # cleared before it is freed, so a second `close` sees nil and does nothing
-  # rather than freeing the same pointer twice.
+  # Closes the client and frees the native connection. Idempotent and
+  # thread-safe: `@close_lock.try_lock` lets exactly one caller free the
+  # handle, while every other concurrent `close` (and every subsequent one)
+  # is a no-op. Without this, two threads could both read `@connection`
+  # before either nulled it, both call `close_client`, and double-decrement
+  # the Arc refcount - which is UB per Rust (issue #212).
   #
-  # Deliberately not synchronized. A Mutex here would break the common
-  # `Signal.trap("TERM") { client.close }` shutdown idiom, because Ruby forbids
-  # locking from a trap context (ThreadError) - and it would leave the handle
-  # leaked when it raised.
+  # `try_lock` (not `synchronize`) is deliberate: `Mutex#synchronize` /
+  # `#lock` raise ThreadError from a trap context, but `#try_lock` does not.
+  # That keeps the standard `Signal.trap("TERM") { client.close }` shutdown
+  # idiom working - the trap either wins the lock and closes, or another
+  # thread has already closed and it silently returns.
   #
-  # In-flight commands are safe without a lock: every glide-ffi command entry
-  # point does `Arc::increment_strong_count` on the handle before using it, and
+  # In-flight commands are safe: every glide-ffi command entry point does
+  # `Arc::increment_strong_count` on the handle before using it, and
   # `close_client` only decrements, so the native ClientAdapter outlives any
-  # request still executing and is dropped once the last one finishes. Verified
-  # with 12 concurrent blocking calls held open across a `close`. See issue #212.
+  # request still executing and is dropped once the last one finishes.
+  # Verified with 12 concurrent blocking calls held open across a `close`.
   def close
-    conn = @connection
-    @connection = nil
+    return unless @close_lock&.try_lock
 
-    Bindings.close_client(conn) unless conn.nil? || conn.null?
+    begin
+      conn = @connection
+      @connection = nil
+
+      Bindings.close_client(conn) unless conn.nil? || conn.null?
+    ensure
+      @close_lock.unlock
+    end
   end
 
   alias disconnect! close
@@ -470,21 +486,7 @@ class Valkey
   private
 
   # Returns the live native client handle, raising if the client has been
-  # closed. Every call into the native layer that takes the client handle must
-  # go through this method: `close` frees the handle and sets `@connection` to
-  # nil, and passing that nil on to the FFI layer makes the core dereference a
-  # NULL pointer, which segfaults the whole VM instead of raising something
-  # Ruby code can rescue. See issue #212.
-  #
-  # The message matches the sibling GLIDE clients (Go ClosingError,
-  # Node/Java ClosingException).
-  #
-  # `nil?` is tested first because `@connection` is nil if `initialize` raised
-  # before connecting, and `nil.null?` would raise NoMethodError.
-  #
-  # The handle is returned rather than read from the ivar at the call site so
-  # callers pass a local, which keeps a concurrent `close` from nulling the
-  # reference between this check and the native call.
+  # closed.
   def connection!
     conn = @connection
     raise ConnectionError, "the client is closed" if conn.nil? || conn.null?
--- a/lib/valkey.rb
+++ b/lib/valkey.rb
@@ -292,11 +292,25 @@ class Valkey
     @in_multi_block = false
   end
 
+  # Closes the client and frees the native connection. Idempotent: the handle is
+  # cleared before it is freed, so a second `close` sees nil and does nothing
+  # rather than freeing the same pointer twice.
+  #
+  # Deliberately not synchronized. A Mutex here would break the common
+  # `Signal.trap("TERM") { client.close }` shutdown idiom, because Ruby forbids
+  # locking from a trap context (ThreadError) - and it would leave the handle
+  # leaked when it raised.
+  #
+  # In-flight commands are safe without a lock: every glide-ffi command entry
+  # point does `Arc::increment_strong_count` on the handle before using it, and
+  # `close_client` only decrements, so the native ClientAdapter outlives any
+  # request still executing and is dropped once the last one finishes. Verified
+  # with 12 concurrent blocking calls held open across a `close`. See issue #212.
   def close
-    return if @connection.nil? || @connection.null?
-
-    Bindings.close_client(@connection)
+    conn = @connection
     @connection = nil
+
+    Bindings.close_client(conn) unless conn.nil? || conn.null?
   end
 
   alias disconnect! close
@@ -353,11 +367,7 @@ class Valkey
   # explicit receiver to issue commands with no dedicated wrapper method yet
   # (e.g. `DEBUG SLEEP`, raw `HSET` in vector search fixtures).
   def send_command(command_type, command_args = [], route: nil, &block)
-    # Validate connection. A nil/null pointer means the client was closed (or
-    # never established a usable connection); surface it as a typed error with
-    # the same "the client is closed" wording the sibling GLIDE clients use
-    # (Go ClosingError, Node/Java ClosingException).
-    raise ConnectionError, "the client is closed" if @connection.nil? || @connection.null? || @connection.address.zero?
+    conn = connection!
 
     channel = 0
 
@@ -399,7 +409,7 @@ class Valkey
         # Use command_with_route_info when an explicit route is provided.
         route_info, _pinned_bufs = route.to_ffi
         res = Bindings.command_with_route_info(
-          @connection,
+          conn,
           channel,
           command_type,
           flattened_args.size,
@@ -415,7 +425,7 @@ class Valkey
         route_str = ""
         route_buf = FFI::MemoryPointer.from_string(route_str)
         res = Bindings.command(
-          @connection,
+          conn,
           channel,
           command_type,
           flattened_args.size,
@@ -459,6 +469,29 @@ class Valkey
 
   private
 
+  # Returns the live native client handle, raising if the client has been
+  # closed. Every call into the native layer that takes the client handle must
+  # go through this method: `close` frees the handle and sets `@connection` to
+  # nil, and passing that nil on to the FFI layer makes the core dereference a
+  # NULL pointer, which segfaults the whole VM instead of raising something
+  # Ruby code can rescue. See issue #212.
+  #
+  # The message matches the sibling GLIDE clients (Go ClosingError,
+  # Node/Java ClosingException).
+  #
+  # `nil?` is tested first because `@connection` is nil if `initialize` raised
+  # before connecting, and `nil.null?` would raise NoMethodError.
+  #
+  # The handle is returned rather than read from the ivar at the call site so
+  # callers pass a local, which keeps a concurrent `close` from nulling the
+  # reference between this check and the native call.
+  def connection!
+    conn = @connection
+    raise ConnectionError, "the client is closed" if conn.nil? || conn.null?
+
+    conn
+  end
+
   # Read an SSL value
   # Accepts a file path (String), an OpenSSL object (#to_pem / #to_der), or a fallback #to_s.
   def read_ssl_value(value, label)
@@ -501,6 +534,9 @@ class Valkey
         return results
       end
     end
+
+    # Checked before allocating any FFI memory below, so a closed client fails fast.
+    conn = connection!
 
     cmds = []
     blocks = []
@@ -559,7 +595,7 @@ class Valkey
 
     begin
       res = Bindings.batch(
-        @connection,
+        conn,
         0,
         batch_info,
         exception,
--- a/lib/valkey.rb
+++ b/lib/valkey.rb
@@ -272,11 +272,7 @@ class Valkey
     end
 
     @connection = res[:conn_ptr]
-    # Serializes `close` so two threads cannot both capture the handle and
-    # double-decrement its Arc refcount (issue #212). We use `Mutex#try_lock`
-    # in `close`, not `#synchronize`, so trap-context callers still work:
-    # only `#lock`/`#synchronize` raise ThreadError from a trap, `#try_lock`
-    # returns false and moves on.
+    # Lock for serializing close() (issue #212).
     @close_lock = Mutex.new
     Bindings.free_connection_response(response_ptr)
 
@@ -298,24 +294,7 @@ class Valkey
     @in_multi_block = false
   end
 
-  # Closes the client and frees the native connection. Idempotent and
-  # thread-safe: `@close_lock.try_lock` lets exactly one caller free the
-  # handle, while every other concurrent `close` (and every subsequent one)
-  # is a no-op. Without this, two threads could both read `@connection`
-  # before either nulled it, both call `close_client`, and double-decrement
-  # the Arc refcount - which is UB per Rust (issue #212).
-  #
-  # `try_lock` (not `synchronize`) is deliberate: `Mutex#synchronize` /
-  # `#lock` raise ThreadError from a trap context, but `#try_lock` does not.
-  # That keeps the standard `Signal.trap("TERM") { client.close }` shutdown
-  # idiom working - the trap either wins the lock and closes, or another
-  # thread has already closed and it silently returns.
-  #
-  # In-flight commands are safe: every glide-ffi command entry point does
-  # `Arc::increment_strong_count` on the handle before using it, and
-  # `close_client` only decrements, so the native ClientAdapter outlives any
-  # request still executing and is dropped once the last one finishes.
-  # Verified with 12 concurrent blocking calls held open across a `close`.
+  # Closes the client and frees the native connection.
   def close
     return unless @close_lock&.try_lock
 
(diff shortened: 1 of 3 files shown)


lib/valkey.rb
```diff
@@ -292,11 +292,25 @@ def initialize(options = {})
...
+  # Deliberately not synchronized. A Mutex here would break the common
+  # `Signal.trap("TERM") { client.close }` shutdown idiom, because Ruby forbids
+  # locking from a trap context (ThreadError) - and it would leave the handle
+  # leaked when it raised.
+  #
+  # In-flight commands are safe without a lock: every glide-ffi command entry
+  # point does `Arc::increment_strong_count` on the handle before using it, and
+  # `close_client` only decrements, so the native ClientAdapter outlives any
+  # request still executing and is dropped once the last one finishes. Verified
+  # with 12 concurrent blocking calls held open across a `close`. See issue #212.
```
nderraugh: I think the counts can still be off.  What are the counts used for?

lib/valkey.rb
```diff
@@ -459,6 +469,29 @@ def send_command(command_type, command_args = [], route: nil, &block)
...
+  #
+  # The message matches the sibling GLIDE clients (Go ClosingError,
+  # Node/Java ClosingException).
+  #
+  # `nil?` is tested first because `@connection` is nil if `initialize` raised
+  # before connecting, and `nil.null?` would raise NoMethodError.
+  #
+  # The handle is returned rather than read from the ivar at the call site so
+  # callers pass a local, which keeps a concurrent `close` from nulling the
+  # reference between this check and the native call.
```
Aryex:
```suggestion
  # Returns the live native client handle, raising if the client has been
  # closed.
```

lib/valkey.rb
```diff
@@ -559,7 +595,7 @@ def send_batch_commands(commands, exception: true, is_atomic: false)
 
     begin
       res = Bindings.batch(
-        @connection,
+        conn,
```
Aryex, line 579: Can we inline these `conn` with `connection!` instead?
```suggestion
        connection!,
```

test/valkey/connection_lifecycle_test.rb
```diff
@@ -27,6 +27,178 @@ def test_closed_client_raises_error
...
+        unexpected = nil
+        worker = Thread.new do
+          200.times { operation.call(client) }
+        rescue Valkey::BaseError
+          nil # expected once the client is closed mid-flight
+        rescue StandardError => e
+          unexpected = e
+        end
+
+        sleep 0.02
```
Aryex, line 227: Could you double check if this is flaky. Perhaps tries this over 100 interations.

lib/valkey.rb
nderraugh: I believe close is not idempotent here when we account for concurrency.

Say a first thread has captured the pointer but has not cleared @connection.

Then a second thread sees the same pointer, clears @connection and closes it.

Then the first thread clears @connection and closes its captured pointer too.

Calling Arc::decrement_strong_count again with that same pointer violates Rust’s safety requirements because the pointer no longer references a live Arc allocation.

nderraugh: I found this code helpful in demonstrating the issue.
```
# frozen_string_literal: true

require "test_helper"
require "timeout"

# Pure unit regression tests for close concurrency. These tests use a dummy
# pointer and stub the native close function, so they require no Valkey server
# and cannot dereference or release native memory.
class TestConnectionLifecycleRace < Minitest::Test
  def test_concurrent_close_releases_native_handle_once
    client = Valkey.allocate
    pointer = FFI::Pointer.new(1)
    client.instance_variable_set(:@connection, pointer)

    source_path, source_line = Valkey.instance_method(:close).source_location
    pause_line = source_line + 2 # @connection = nil
    assert_equal "@connection = nil", File.readlines(source_path)[pause_line - 1].strip

    reached_pause = Queue.new
    release_first_close = Queue.new
    close_calls = []
    first_close = nil

    trace = TracePoint.new(:line) do |event|
      next unless Thread.current[:first_close]
      next unless event.path == source_path && event.lineno == pause_line

      # A :line event fires before the line executes. At this point the first
      # close has copied @connection into conn but has not cleared @connection.
      reached_pause << true
      release_first_close.pop
    end

    Valkey::Bindings.stub(:close_client, ->(conn) { close_calls << conn }) do
      trace.enable
      first_close = Thread.new do
        Thread.current[:first_close] = true
        client.close
      end

      Timeout.timeout(2) { reached_pause.pop }

      # The second close can capture and release the same pointer while the
      # first close is paused between reading and clearing @connection.
      Thread.new { client.close }.join
    ensure
      release_first_close << true
      first_close&.join
      trace.disable
    end

    assert_equal 1, close_calls.size,
                 "close_client should be invoked exactly once for one native ownership count"
    assert_same pointer, close_calls.first
  end
end
```

Aryex: The other clients `close` methods all implement some form of locking/concurrency control. I think we should do the same here instead.

The original `mutex` solution was rejected due to concerns that a client could be used inside a `Signal.trap() { client.close() }` and may happen indirectly through a framework. However I think this overstated because:
1. Mutex are common concurrency control. It makes little sense for a framework to disallow its usage because it may use `Signal.trap`.
2. I did a quick AI query and it looks like no Ruby framework explicitly mention Mutex is not allowed.
3. Should mutex becomes an issue, we can use an Atomic similar to Java.

jamesx-improving: Thanks for the catch — and thanks especially for the TracePoint reproducer, which forced the interleave much more reliably than the timing-only `test_concurrent_close_frees_handle_once` did.

Addressed in e627610: `close` now serializes via `@close_lock.try_lock`, so exactly one caller frees the handle while the rest short-circuit. I picked `try_lock` (not `synchronize`) deliberately because only `Mutex#lock`/`#synchronize` raise `ThreadError` from a trap context, so `Signal.trap("TERM") { client.close }` still works — the trap either wins the lock and closes, or another thread already closed and it silently returns. Docblock now records that reasoning.

I ported your reproducer into `test/valkey/connection_lifecycle_test.rb` as `test_concurrent_close_releases_native_handle_once_traced`, with attribution to you in the comment. It locates the pause line dynamically so a future refactor of `close` can't silently invalidate the pause point.

lib/valkey.rb
jamesx-improving: Accepted — trimmed to the two-line summary in e627610.

lib/valkey.rb
jamesx-improving: Left as-is on purpose — the `conn = connection!` assignment is hoisted before any FFI allocation ([L540-L541](https://github.com/valkey-io/valkey-glide-ruby/blob/e627610/lib/valkey.rb#L540-L541)), so a closed client fails fast without churning through `CmdInfo`/`BatchInfo`/buffer allocation and OTel span setup. Inlining as `Bindings.batch(connection!, ...)` would move the check after all of that. `send_command` uses the same shape at [L386](https://github.com/valkey-io/valkey-glide-ruby/blob/e627610/lib/valkey.rb#L386) for the same reason.

test/valkey/connection_lifecycle_test.rb
jamesx-improving: Ran it as `bundle exec ruby -Ilib -Itest test/standalone/valkey_test.rb --name test_close_racing_in_flight_commands_does_not_crash` in a shell loop, 100 iterations against Valkey 8.1.3 on macOS at head e627610 — **100 / 100 passing, 0 flakes**. Thanks for the nudge; 3/3 in the PR body was thin.

lib/valkey.rb
```diff
@@ -272,6 +272,12 @@ def initialize(options = {})
     end
 
     @connection = res[:conn_ptr]
+    # Serializes `close` so two threads cannot both capture the handle and
+    # double-decrement its Arc refcount (issue #212). We use `Mutex#try_lock`
+    # in `close`, not `#synchronize`, so trap-context callers still work:
+    # only `#lock`/`#synchronize` raise ThreadError from a trap, `#try_lock`
+    # returns false and moves on.
```
Aryex: Can we make this more concise
```suggestion
    # Lock for serializing close(). 
```

lib/valkey.rb
```diff
@@ -292,11 +298,35 @@ def initialize(options = {})
...
+  # `#lock` raise ThreadError from a trap context, but `#try_lock` does not.
+  # That keeps the standard `Signal.trap("TERM") { client.close }` shutdown
+  # idiom working - the trap either wins the lock and closes, or another
+  # thread has already closed and it silently returns.
+  #
+  # In-flight commands are safe: every glide-ffi command entry point does
+  # `Arc::increment_strong_count` on the handle before using it, and
+  # `close_client` only decrements, so the native ClientAdapter outlives any
+  # request still executing and is dropped once the last one finishes.
+  # Verified with 12 concurrent blocking calls held open across a `close`.
```
Aryex:
```suggestion
  # Closes the client and frees the native connection.
```

Aryex (commented):
Just some small nit to reduce the doc strings.

lib/valkey.rb
jamesx-improving: Trimmed in `8bc4d35`. Kept a one-line pointer (`See `close` for why try_lock is used.`) so the trap-context rationale is still discoverable from here, but the detail now lives only in the `close` docblock.

jamesx-improving: Correction on the final form (`785b913`): since the `close` docblock is also being trimmed to a single line (per your other suggestion), the field comment now reads `# Lock for serializing close() (issue #212).` — pointing at the issue for the full rationale rather than at `close`.

lib/valkey.rb
jamesx-improving: Done in `785b913` — collapsed to the one-line summary as suggested. The concurrency/trap-context rationale stays in git history and issue #212 for anyone who needs the detail.

yipin-chen (approved):
LGTM

Aryex (approved):

PR #205, opened 2026-07-30
Backport #192: Valkey::Future placeholders for pipelined/multi to release-1.0

Aryex:
## Summary
- Backports #192 (`f1199aef`) to `release-1.0`.
- Adds `Valkey::Future` placeholders returned by commands queued inside `pipelined` / `multi`, so callers can reference each queued reply after the batch resolves.
- Also carries the collateral changes from that PR: nested-array arg flattening, stricter `build_command_args` argument-type checks, `Integer()` coercion for `setex`/`set`/`getex`/`setrange`/`incrby`/`decrby` numeric args, and the scripting use-after-free fix.

Cherry-picked cleanly with `git cherry-pick -x` — no conflicts.

PR #203, opened 2026-07-29
Backport #164: multi atomic batch support to release-1.0

Aryex:
## Summary

Backport of #164 (commit c932ad2) to `release-1.0`.

Adopts core's BATCH support for atomic MULTI transactions, restores `setex`/`psetex`/`setnx` (fixed in the bumped `valkey-glide` submodule), and adds transaction lint coverage.

## Conflicts resolved

Cherry-pick surfaced conflicts against release-1.0's newer code:

- `lib/valkey.rb` — kept release-1.0's `read_ssl_value` helper alongside the new `send_batch_commands(..., is_atomic:)` signature from #164.
- `lib/valkey/commands/string_commands.rb` (`setnx`) — dropped the `value.to_s` coercion in favor of #164's raw `value`. Release-1.0's `build_command_args` already handles String/Symbol/Integer/Float coercion (and rejects other types), so the explicit `.to_s` is redundant and #164's form is preferred.

PR #202, opened 2026-07-29
Backport #174: flexible positional eval/evalsha to release-1.0

Aryex:
## Summary

Backport of #174 (commit d053b27) to `release-1.0`.

Updates `eval` and `evalsha` to accept the flexible positional argument form used by redis-rb, matching the API on `main`.

PR #225, opened 2026-08-04
fix: remove `get` interception of "shared" that discards WATCH

### LINKED ISSUE #214: `get` inside `MULTI` silently discards `WATCH` when the key is named `"shared"`, losing concurrent writes
# `get` inside `MULTI` silently discards `WATCH` when the key is named `"shared"`, losing concurrent writes

## Summary

`get` calls three private helpers in `lib/valkey/commands/string_commands.rb` that — under a narrow set of conditions — issue a `DISCARD`, run a standalone `GET` outside the transaction, re-issue `MULTI`, and replay the queued commands. Because `DISCARD` clears watched keys server-side, the transaction then proceeds **unguarded**: a concurrent write that should have aborted `EXEC` is silently overwritten instead.

The gate is the key name being literally `"shared"`. That is not a reserved or internal name — a shared counter, shared config blob, or shared lock is an entirely ordinary thing to call `shared`.

The behaviour traces to `test_transaction_isolation` (`test/lint/transaction_commands.rb:373`), which asserts that a same-connection `GET` during `MULTI` returns the live pre-transaction value. No correct client can satisfy that: after `MULTI` the server queues every command and replies `QUEUED`. The same file contains `test_watch_with_an_unmodified_key` (line 448) asserting the correct semantics on an identical shape, so the suite currently asserts both behaviours at once — and the `"shared"` gate is what keeps them from colliding.

## Steps to reproduce

Standalone (not cluster). Two connections, ordinary optimistic-locking shape:

```ruby
require "valkey"

def scenario(key)
  a = Valkey.new
  b = Valkey.new

  a.set(key, "initial")
  a.watch(key)
  a.multi
  a.set(key, "my_update")
  got = a.get(key)                   # <-- interception point

  b.set(key, "CONCURRENT_WRITER")    # concurrent write AFTER the WATCH

  puts "key=%p  get-in-multi=>%p  exec=>%p  final=>%p" %
       [key, got, a.exec, b.get(key)]
end

scenario("shared")
scenario("normal")
```

```
key="shared"  get-in-multi=>"initial"  exec=>["OK"]  final=>"my_update"
key="normal"  get-in-multi=>"QUEUED"   exec=>nil     final=>"CONCURRENT_WRITER"
```

The `"normal"` line is correct: `WATCH` saw the concurrent write, `EXEC` returned `nil`, the transaction aborted, and the concurrent writer's value survived. The `"shared"` line lost that write with no exception, no warning, and no log entry.

`redis-rb` 5.4.1 on the same scenario aborts correctly and preserves `"CONCURRENT_WRITER"`.

## Expected behavior

`get` behaves identically regardless of key name. Inside `MULTI`, `GET` returns `"QUEUED"`; `WATCH` remains in force; a concurrent write to a watched key causes `EXEC` to return `nil` and the transaction to abort.

## Actual behavior

Two separate wrong results when the gate is hit:

1. **Silent lost update** — `WATCH` is destroyed by the hidden `DISCARD`, so `EXEC` succeeds and overwrites the concurrent write.
2. **Wrong `EXEC` arity** — `exec` returns `["OK"]` (one reply) where two commands were queued. This one is deterministic and needs no concurrency at all to observe.

## Conditions

All three must hold, which is why this hasn't been hit in normal use:

- the key is exactly `"shared"` (`isolation_check_pattern?`, `string_commands.rb:379`)
- the non-block `multi` / `exec` form is used — the block form `multi { … }` is unaffected
- the `GET` is the second queued command, following a `SET` of that same key

The check runs *at `GET` time*, so adding a third command **after** the `GET` does not help — `set` / `get` / `set` still trips it, and `exec` then returns 2 replies for 3 queued commands. What is safe is a `GET` that is not itself the second command: with `set` / `set` / `get`, the `GET` correctly returns `"QUEUED"`.

Verified: `multi { |m| m.set("shared","x"); m.get("shared") }` returns `["OK", "x"]` correctly, and with a third queued command `GET` returns `"QUEUED"` as it should.

## Affected code

- `lib/valkey/commands/string_commands.rb:209` — the call site in `get`
- `lib/valkey/commands/string_commands.rb:349` — `should_intercept_get?`
- `lib/valkey/commands/string_commands.rb:355` — `handle_transaction_isolation_get`, which issues the `DISCARD`
- `lib/valkey/commands/string_commands.rb:376` — `isolation_check_pattern?`, holding the `key == "shared"` comparison

Introduced by #74 (`1f846f5`), "Adding support for transaction commands".

## Suggested fix

Remove the three helpers and the call site so `get` uses the normal command path, and drop the now-unused guard state (`@in_multi_block`, `@queued_commands` bookkeeping where it exists only to serve this path).

`test_transaction_isolation` then needs rewriting to assert something a correct client can actually deliver — which is also a better test of isolation:

```ruby
def test_transaction_isolation
  skip("MULTI/EXEC not supported in cluster mode") if cluster_mode?

  observer = _new_client(db: 15)   # _new_client does not run init, so it would land on db 0
  r.set("shared", "initial")

  r.multi
  r.set("shared", "transaction_value")

  # same connection: commands are queued, not executed
  assert_equal "QUEUED", r.get("shared")

  # a second connection sees the pre-transaction value until EXEC
  assert_equal "initial", observer.get("shared")

  assert_equal ["OK", "transaction_value"], r.exec
  assert_equal "transaction_value", observer.get("shared")
ensure
  observer&.close
end
```

Worth adding alongside it: a test that a watched-key conflict aborts `EXEC` for a key named `"shared"` specifically, so this cannot regress silently.

*(Corrected: an earlier version of this snippet used a bare `_new_client`. Because `_new_client` does not call `init`, the observer connects to db 0 while `r` is on db 15, so `observer.get("shared")` returns `nil` and the assertion fails. Passing `db: 15` fixes it — see `test/lint/generic_commands.rb:35` for the same pattern.)*

## Impact

Silent data loss in an atomicity primitive. Any watch-based read-modify-write on a key named `shared` can lose a concurrent update with nothing surfacing to the caller. The narrow gate limits how many users hit it, but nothing a user can observe distinguishes the broken path from the correct one, and the gated code path is exercised only by the test that motivated it — so it has no real verification.

## Environment

- Ruby: 4.0.6, arm64-darwin25
- Valkey: 8.1.3 (standalone)
- Compared against: `redis-rb` 5.4.1

jamesx-improving:
### Summary

`get` contained a hardcoded interception path that fired only when the key was literally `"shared"`. On that path it issued a hidden `DISCARD`, read the key outside the transaction, then re-opened a fresh `MULTI` — destroying server-side `WATCH` state and silently losing a concurrent write. This PR deletes the interception path so `get` always takes the normal command path, and rewrites the test that the hack existed to satisfy.

Two observable failures went away:

- **Silent lost update.** `DISCARD` unwatches all keys server-side, so the subsequent `EXEC` committed unconditionally and overwrote a concurrent write that should have aborted it. Nothing distinguished the broken path — no exception, no warning, no log entry — and `"shared"` is an entirely ordinary key name (shared counter, shared config, shared lock).
- **Wrong `EXEC` arity.** The hack popped the `GET` from `@queued_commands` and reset the ledger, so `exec` returned one reply fewer than the caller queued. Deterministic; needs no concurrency to observe.

### Issue link

Closes #214

### Features / Changes

`lib/valkey/commands/string_commands.rb`
- `get` is now just `send_command(RequestType::GET, [key])`.
- Deleted three private helpers — `should_intercept_get?`, `handle_transaction_isolation_get`, `isolation_check_pattern?` — and the now-useless `private` keyword that guarded only those three (leaving it would trip RuboCop `Lint/UselessAccessModifier`).
- Public surface of `StringCommands` is unchanged: 24 public methods before and after. Only dead private methods were removed.

`lib/valkey.rb`
- Removed the dead `@in_multi_block` ivar. It was written `false` exactly once and read exactly once, from the deleted predicate; it was never set `true` anywhere in the repo, so the `!@in_multi_block` conjunct was a tautology.
- `@in_multi` and `@queued_commands` are untouched — they are real transaction bookkeeping still consumed by `transaction_commands.rb` (`exec` / `reconvert_queued_replies`).

`test/lint/transaction_commands.rb`
- Rewrote `test_transaction_isolation`; added two regression tests.

**Why rewriting a pre-existing test is correct here, not merely convenient:**

1. **The old assertion was impossible for any correct client.** It asserted that a same-connection `GET` issued during `MULTI` returns the live pre-transaction value. After `MULTI` the server replies `QUEUED` to every command. That is protocol, not a judgment call — no client can satisfy the old assertion without faking it, which is exactly what the deleted code did.
2. **The impossible test and the production hack landed in the same commit.** `1f846f5` ("Adding support for transaction commands", #74) added both the three helpers and the test requiring them. The deleted helper carried the inline comment *"Only intercepts when key is `shared` to match the specific test case"*. The production code existed solely to satisfy its own test — no user requested it and nothing independently verified it.
3. **The suite already contradicted itself.** `test_queued_commands` (`test/lint/transaction_commands.rb:117`) asserts `assert_equal "QUEUED", r.get("foo")` on an identical shape. The `"shared"` key gate was the only thing keeping the two assertions from colliding. That pre-existing test is also the precedent that makes the rewrite's `"QUEUED"` assertion uncontroversial.
4. **The rewrite tests isolation more strictly, not less.** It keeps the isolation claim and observes it the only way possible: a second connection sees `"initial"` until `EXEC`, then `"transaction_value"`. It additionally pins the `EXEC` arity, which the old test could not.

The two added regression guards: `test_watch_with_a_modified_key_named_shared` (a concurrent write to a `WATCH`ed key named `"shared"` must abort `EXEC`) and `test_exec_returns_a_reply_per_queued_command_with_a_key_named_shared` (arity, observable without concurrency).

### Testing

- `bundle exec rubocop` — 99 files inspected, 0 offenses.
- Baseline on pristine `fb2fa46`: **840 tests, 4149 assertions, 0 failures, 0 errors, 100 skips.**
- With this change: **842 tests, 4199 assertions, 0 failures, 0 errors, 100 skips.**
  Command: `CI=1 SKIP_TLS_TESTS=true VALKEY_PORT=6414 bundle exec rake test:standalone` (dedicated Valkey 8.1.3 on port 6414).
- **Bisect proof that the new tests actually pin the fix.** With `lib/` reverted to the buggy state and only the tests kept, all three tests fail — `test_transaction_isolation` and `test_watch_with_a_modified_key_named_shared` both show `-"QUEUED"` / `+"initial"`, and the arity test shows `Expected "QUEUED", Actual nil`. All three pass with the `lib/` fix applied.
- **Gate reach verified** across four command shapes before deleting: `set`/`get`; `set`/`get`/`set` (returned 2 replies for 3 queued commands); `set`/`set`/`get` (correct — `GET` third does not trip the gate); and the block form (correct and unaffected, since `multi { }` builds a `Valkey::Pipeline` and never sets `@in_multi`).
- **Cluster suite deferred** — no local cluster available. Every new test carries `skip(...) if cluster_mode?`, matching the surrounding tests, since `test/lint/` is shared by both suites. A reviewer with a cluster can confirm with `bundle exec rake test:cluster`.

### Checklist

-   [ ] Documentation is updated (if applicable). — n/a, no public API or documented behaviour change; the removed methods were private.
-   [x] Destination branch is correct - main (destination here is `release-1.0`, per the 1.0.0 GA release-blocker track — matching #207, #205, #203, #202, #201)

### DIFF
--- a/lib/valkey/commands/string_commands.rb
+++ b/lib/valkey/commands/string_commands.rb
@@ -205,9 +205,7 @@ class Valkey
       # @param [String] key
       # @return [String]
       def get(key)
-        result = send_command(RequestType::GET, [key])
-        result = handle_transaction_isolation_get(key, result) if should_intercept_get?(result)
-        result
+        send_command(RequestType::GET, [key])
       end
 
       # Atomically set +key+ to +value+ and return the previous value.
@@ -341,42 +339,6 @@ class Valkey
         args << "WITHMATCHLEN" if with_match_len
 
         send_command(RequestType::LCS, args)
-      end
-
-      private
-
-      # Check if GET should be intercepted for transaction isolation check
-      def should_intercept_get?(result)
-        @in_multi && !@in_multi_block && result == "QUEUED" && !@queued_commands.nil? && @queued_commands.size == 2
-      end
-
-      # Handle GET interception for transaction isolation (test_transaction_isolation pattern)
-      # Only intercepts when key is "shared" to match the specific test case
-      def handle_transaction_isolation_get(key, result)
-        first_cmd = @queued_commands.first
-        last_cmd = @queued_commands.last
-        return result unless isolation_check_pattern?(first_cmd, last_cmd, key)
-
-        # Remove the GET that was just added
-        @queued_commands.pop
-        saved_commands = @queued_commands.dup
-        send_command(RequestType::DISCARD)
-        @in_multi = false
-        result = send_command(RequestType::GET, [key])
-        send_command(RequestType::MULTI)
-        @in_multi = true
-        @queued_commands = []
-        saved_commands.each do |cmd_type, cmd_args|
-          send_command(cmd_type, cmd_args)
-        end
-        result
-      end
-
-      # Check if this matches the isolation check pattern
-      def isolation_check_pattern?(first_cmd, last_cmd, key)
-        first_cmd && first_cmd[0] == RequestType::SET && first_cmd[1] && first_cmd[1][0] == key &&
-          last_cmd && last_cmd[0] == RequestType::GET && last_cmd[1] && last_cmd[1][0] == key &&
-          key == "shared"
       end
     end
   end
--- a/test/lint/transaction_commands.rb
+++ b/test/lint/transaction_commands.rb
@@ -402,32 +402,6 @@ module Lint
       observer&.close
     end
 
-    def test_get_between_multi_and_exec_does_not_discard_watch
-      # In cluster mode, MULTI/EXEC transactions require all keys in same slot
-      # and behave differently with connection routing
-      skip("MULTI/EXEC not supported in cluster mode") if cluster_mode?
-
-      # Regression guard: a GET issued between MULTI and EXEC must stay queued
-      # and must not discard WATCH, otherwise a concurrent write to the
-      # watched key is silently lost.
-      writer = _new_client(db: 15)
-
-      r.set("account:1", "initial")
-      r.watch("account:1")
-      r.multi
-      r.set("account:1", "my_update")
-      assert_equal "QUEUED", r.get("account:1")
-
-      # Concurrent write to the watched key, after the WATCH
-      writer.set("account:1", "concurrent_writer")
-
-      # WATCH is still in force, so the transaction must abort
-      assert_nil r.exec
-      assert_equal "concurrent_writer", r.get("account:1")
-    ensure
-      writer&.close
-    end
-
     def test_complex_transaction_scenario
       # In cluster mode, MULTI/EXEC transactions require all keys in same slot
       # and behave differently with connection routing
--- a/test/lint/transaction_commands.rb
+++ b/test/lint/transaction_commands.rb
@@ -375,19 +375,76 @@ module Lint
       # and behave differently with connection routing
       skip("MULTI/EXEC not supported in cluster mode") if cluster_mode?
 
+      # A second connection is required to observe isolation: on the connection
+      # that issued MULTI the server replies QUEUED to every command, so that
+      # connection cannot read the pre-transaction value. `_new_client` does not
+      # run `init`, so it must be pointed at the test database explicitly.
+      observer = _new_client(db: 15)
+
       r.set("shared", "initial")
 
       # Start transaction but don't execute yet
       r.multi
       r.set("shared", "transaction_value")
 
-      # Value should still be initial since transaction not executed
-      assert_equal "initial", r.get("shared")
+      # Same connection: commands are queued, not executed
+      assert_equal "QUEUED", r.get("shared")
 
-      # Execute transaction
+      # Another connection still sees the pre-transaction value until EXEC
+      assert_equal "initial", observer.get("shared")
+
+      # Execute transaction: two queued commands produce two replies
       result = r.exec
-      assert_equal ["OK"], result
+      assert_equal %w[OK transaction_value], result
       assert_equal "transaction_value", r.get("shared")
+      assert_equal "transaction_value", observer.get("shared")
+    ensure
+      observer&.close
+    end
+
+    def test_watch_with_a_modified_key_named_shared
+      # In cluster mode, MULTI/EXEC transactions require all keys in same slot
+      # and behave differently with connection routing
+      skip("MULTI/EXEC not supported in cluster mode") if cluster_mode?
+
+      # Regression guard: `get` must not special-case any key name. A GET issued
+      # between MULTI and EXEC must stay queued and must never discard WATCH,
+      # otherwise a concurrent write to the watched key is silently lost.
+      writer = _new_client(db: 15)
+
+      r.set("shared", "initial")
+      r.watch("shared")
+      r.multi
+      r.set("shared", "my_update")
+
+      assert_equal "QUEUED", r.get("shared")
+
+      # Concurrent write to the watched key, after the WATCH
+      writer.set("shared", "concurrent_writer")
+
+      # WATCH is still in force, so the transaction must abort
+      assert_nil r.exec
+      assert_equal "concurrent_writer", r.get("shared")
+    ensure
+      writer&.close
+    end
+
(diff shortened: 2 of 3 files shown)


test/lint/transaction_commands.rb
```diff
@@ -375,19 +375,76 @@ def test_transaction_isolation
...
       result = r.exec
-      assert_equal ["OK"], result
+      assert_equal %w[OK transaction_value], result
       assert_equal "transaction_value", r.get("shared")
+      assert_equal "transaction_value", observer.get("shared")
+    ensure
+      observer&.close
+    end
+
+    def test_watch_with_a_modified_key_named_shared
```
Aryex: Do we need these tests? I don't think we need to test some special hard coded behavior

lib/valkey/commands/string_commands.rb
```diff
@@ -205,9 +205,7 @@ def mapped_msetnx(hash)
       # @param [String] key
       # @return [String]
       def get(key)
-        result = send_command(RequestType::GET, [key])
-        result = handle_transaction_isolation_get(key, result) if should_intercept_get?(result)
-        result
+        send_command(RequestType::GET, [key])
       end
```
currantw, line 211: How/why was this even added in the first place?! 🤯

test/lint/transaction_commands.rb
currantw: "+1"

currantw (commented):
Reviewed to c801942.

test/lint/transaction_commands.rb
jamesx-improving: Consolidated into a single MULTI/WATCH test (`test_get_between_multi_and_exec_does_not_discard_watch`) using a generic key name (`account:1`). `test_transaction_isolation` (also in this PR) retains coverage of the "shared" literal + arity of EXEC. Addressed in d3f2170. Happy to revisit if you'd prefer a different shape.

Aryex (approved):

yipin-chen (approved):

PR #226, opened 2026-08-04
fix: dispatch `slowlog` subcommands to their RequestType constants

### LINKED ISSUE #215: `slowlog` is entirely unusable — every subcommand raises `TypeError: no implicit conversion of Array into Integer`
# `slowlog` is entirely unusable — every subcommand raises `TypeError: no implicit conversion of Array into Integer`

## Summary

`slowlog` passes an `Array` to `send_command` where the first parameter is a numeric `RequestType`. Every call raises, for every subcommand and both spellings, so the method is 100% dead. There is no test coverage for `slowlog` anywhere in the suite, which is why this shipped.

```ruby
def slowlog(subcommand, length = nil)
  args = [:slowlog, subcommand]
  args << Integer(length) if length
  send_command(args)          # <-- Array in the command_type slot
end
```

`send_command`'s signature is `send_command(command_type, command_args = [], route: nil, &block)`, so `args` lands in `command_type` and the first integer conversion inside blows up.

The `[:slowlog, subcommand]` line has been broken since the initial import, **`b222363`** ("add all files") — `send_command`'s first parameter was already the numeric `command_type` slot at that point (the FFI `attach_function :command` declared `:int` for it), so this method never worked.

It was then **missed by #38 (`6d9751f`), "server commands - time"**, which converted this file to numeric dispatch: that commit rewrote `slaveof` immediately above (`send_command([:slaveof, …])` → `send_command(RequestType::SLAVE_OF, …)`) and converted `sync`/`save`/`lastsave`/`bgsave`/`bgrewriteaof`/`time` immediately below, leaving only the `slowlog` hunk sitting between them untouched. So #38 is the commit that should have fixed this and didn't, not the commit that introduced it — `git log -L` on the method shows `b222363` and nothing else.

*(Corrected twice: an earlier version attributed this to #23 (`4dcd95d`) — a pure rename, 97% similarity, no logic change. A later version said #38 introduced it; #38 in fact only failed to fix it. The defect dates to the initial import.)*

## Steps to reproduce

```ruby
require "valkey"
c = Valkey.new

c.slowlog(:len)      # TypeError: no implicit conversion of Array into Integer
c.slowlog("len")     # TypeError
c.slowlog(:get)      # TypeError
c.slowlog("GET")     # TypeError
c.slowlog(:reset)    # TypeError
c.slowlog(:get, 2)   # TypeError

# the underlying command is fine — only the wrapper is broken
c.call("SLOWLOG", "LEN")   # => 0
```

## Expected behavior

```ruby
c.slowlog(:len)     # => 0
c.slowlog(:get)     # => []
c.slowlog(:get, 2)  # => [] (at most 2 entries)
c.slowlog(:reset)   # => "OK"
```

## Actual behavior

`TypeError: no implicit conversion of Array into Integer` on every call. The message is also misleading — it names `Array` and `Integer`, neither of which appears in the user's call, giving no hint that the wrapper is at fault.

## Affected code

`lib/valkey/commands/server_commands.rb:222-227`.

Relevant constraint: there is no bare `SLOWLOG` request type. `lib/valkey/request_type.rb:278-280` defines only the three subcommand-specific constants:

```ruby
SLOWLOG_GET = 1157
SLOWLOG_LEN = 1158
SLOWLOG_RESET = 1159
```

Confirmed: `Valkey::RequestType.const_defined?(:SLOWLOG)` is `false`. So this can't be fixed by swapping in a single constant — it needs real subcommand-to-constant dispatch.

## Suggested fix

Map the subcommand through a frozen lookup table, preserving the public signature:

```ruby
SLOWLOG_SUBCOMMANDS = {
  "get" => RequestType::SLOWLOG_GET,
  "len" => RequestType::SLOWLOG_LEN,
  "reset" => RequestType::SLOWLOG_RESET
}.freeze
private_constant :SLOWLOG_SUBCOMMANDS

# Interact with the slowlog (get, len, reset)
#
# @param [String, Symbol] subcommand one of `get`, `len`, `reset`
# @param [Integer] length maximum number of entries to return (`get` only)
# @return [Array<Array>, Integer, String] depends on subcommand
def slowlog(subcommand, length = nil)
  request_type = SLOWLOG_SUBCOMMANDS[subcommand.to_s.downcase]
  unless request_type
    raise ArgumentError,
          "unknown SLOWLOG subcommand #{subcommand.inspect} " \
          "(expected one of: #{SLOWLOG_SUBCOMMANDS.keys.join(', ')})"
  end

  args = []
  args << Integer(length) if length
  send_command(request_type, args)
end
```

`.to_s.downcase` accepts both `:len` and `"LEN"`, matching how the rest of the client treats subcommand names.

One deliberate behaviour note: `slowlog(:help)` has no corresponding `RequestType` constant, so it surfaces as the explicit `ArgumentError` above rather than a `TypeError`. That seems preferable to special-casing it, but happy to add `SLOWLOG HELP` via a generic passthrough if the maintainers would rather keep it working.

## Test coverage

There is currently no `slowlog` test at all — `grep -rn slowlog test/` returns nothing. Suggested lint coverage:

```ruby
def test_slowlog_len
  assert_kind_of Integer, r.slowlog(:len)
end

def test_slowlog_get
  assert_kind_of Array, r.slowlog(:get)
end

def test_slowlog_get_with_length
  assert_kind_of Array, r.slowlog(:get, 2)
end

def test_slowlog_reset
  assert_equal "OK", r.slowlog(:reset)
end

def test_slowlog_accepts_string_and_symbol_subcommands
  assert_equal r.slowlog(:len), r.slowlog("LEN")
end

def test_slowlog_unknown_subcommand_raises_argument_error
  assert_raises(ArgumentError) { r.slowlog(:nope) }
end
```

## Impact

A documented, shipped operational command that cannot be invoked in any way. `SLOWLOG` is a primary tool for diagnosing latency, so its absence is felt precisely during an incident. It does fail loudly and `call("SLOWLOG", "LEN")` works as a workaround, but a public API method that is dead for every possible input seems worth fixing before 1.0.0 GA — particularly since the zero test coverage means nothing would catch a similar gap elsewhere.

## Environment

- Ruby: 4.0.6, arm64-darwin25
- Valkey: 8.1.3 (standalone)

jamesx-improving:
### Summary

`Valkey#slowlog` has never worked. It built a redis-rb-style argument array (`[:slowlog, subcommand]`) and passed it as the **first** argument to `send_command`, whose first parameter is the numeric `command_type` slot — the FFI `attach_function :command` declares it `:int`. Every call raised `TypeError: no implicit conversion of Array into Integer`. This PR dispatches each subcommand to its own `RequestType` constant (`SLOWLOG_GET` / `SLOWLOG_LEN` / `SLOWLOG_RESET`) and rejects unknown subcommands with a clear `ArgumentError`.

### Issue link

Closes #215

### Features / Changes

**`lib/valkey/commands/server_commands.rb`** — rewrote `slowlog` to normalize the subcommand with `to_s.downcase` and `case`-dispatch to the right `RequestType`; `length` is forwarded as a command argument for `get`. Unknown subcommands raise `ArgumentError`. YARD docs on the method were expanded (params, `@raise`, `@example`, `@see`).

**`test/lint/server_commands.rb`** — 8 new tests covering all three subcommands, string/symbol/mixed-case spellings, `length` pass-through, and the `ArgumentError` paths.

#### Provenance

Broken **since the initial import `b222363`** ("add all files") — `send_command`'s first parameter was already the numeric `command_type` slot at that commit, so this method never worked at any point in history. It was then **missed by #38 (`6d9751f`, "server commands - time")**, which converted this file to numeric dispatch: that commit rewrote `slaveof` immediately above and converted `sync`/`save`/`lastsave`/`bgsave`/`bgrewriteaof`/`time` below, leaving only the `slowlog` hunk in between untouched. #38 should have fixed it and didn't. `git log -L` on the method body shows `b222363` and nothing else.

> **Retraction:** the issue originally attributed this to #23 (`4dcd95d`). That is incorrect — `4dcd95d` is a pure rename (`server.rb` → `server_commands.rb`, `module Server` → `module ServerCommands`) at 97% similarity with no logic change.

#### ⚠️ This reverses a written reviewer preference — please confirm

Issue #216 Group 5, row R3-3a records verbatim: *"Reviewer preference: **remove from GA** (no other GLIDE binding exposes it)."* That text is live in #216 today. This PR **fixes the dispatch instead of removing the command**, for these reasons:

- **Fixing is non-breaking; removal deletes a documented public API.** `slowlog` is a shipped, documented method.
- **redis-rb exposes `slowlog`.** Removing it widens the very parity gap the README advertises at line 3 ("an interface similar to redis-rb") and line 12 ("**Drop-in Replacement**: Familiar redis-rb-style API"). `CLAUDE.md` also mandates matching redis-rb conventions.
- **The premise argues the wrong axis.** "No other GLIDE binding exposes it" is an argument for parity with other *bindings*; this gem's stated contract is parity with *redis-rb*.

If maintainers still prefer removal for 1.0.0 GA, say so and I'll close this in favour of a removal PR.

#### ⚠️ Deliberate narrowing: `slowlog(:help)` now raises `ArgumentError`

There is no `SLOWLOG_HELP` `RequestType`, so `:help` surfaces a clear `ArgumentError` instead of today's misleading `TypeError`. `call("SLOWLOG", "HELP")` remains available and is verified working. This diverges from redis-rb's verbatim passthrough, so it is an explicit reviewer decision point rather than an oversight.

#### Design notes (for reviewers comparing against the issue's suggested code)

- **Inline `case`, not a frozen `Hash` + `private_constant`.** The issue suggested the latter. I used an inline `case` to match `xinfo` (`lib/valkey/commands/stream_commands.rb:541`) and `xgroup` (`:261`) — the repo's only comparable subcommand dispatchers — including identical error-message wording, making it 3-of-3. `private_constant` appears nowhere in `lib/valkey/commands/`.
- **No `route:` kwarg, deliberately.** SLOWLOG is per-node and 14 neighbouring admin commands do take `route:`, but `Valkey::Pipeline#send_command` accepts no `route:` — so all 14 (`bgrewriteaof`, `bgsave`, `config_get`/`config_set`/`config_resetstat`/`config_rewrite`, `dbsize`, `flushall`, `flushdb`, `info`, `lastsave`, `save`, `time`, `lolwut`) **already** raise `ArgumentError: wrong number of arguments` inside `pipelined` on the unmodified baseline — verified. Route-free `slowlog` works in `pipelined` today (`pipelined { |p| p.slowlog(:len); p.slowlog(:get) }` → `[0, []]`). Adding `route:` would inherit that latent breakage. **A separate issue is warranted for the pipeline/`route:` defect** (not filed as part of this PR).
- **Cluster-mode skips are intentional.** SLOWLOG is per-node and may return a per-node Hash in cluster mode (unverified — no local cluster available). The 6 wire-touching tests carry `skip(...) if cluster_mode?`. The 2 `ArgumentError` tests deliberately run in **both** modes, since they raise before any `send_command` — giving real cluster coverage of the validation path.
- **Two known warts left as-is for consistency with `xinfo`/`xgroup`:** `length` is silently dropped for `:len`, and `slowlog(nil)` produces an empty-subcommand message.

### Testing

**Lint** — `bundle exec rubocop lib/valkey/commands/server_commands.rb test/lint/server_commands.rb` → **2 files inspected, 0 offenses**.

**Standalone suite** — `CI=1 SKIP_TLS_TESTS=true VALKEY_PORT=6415 bundle exec rake test:standalone` against a dedicated Valkey 8.1.3 on port 6415:

| | tests | assertions | failures | errors | skips |
|---|---|---|---|---|---|
| Baseline (pristine `fb2fa46`) | 840 | 4167 | 0 | 0 | 100 |
| With this fix | 848 | 4189 | 0 | 0 | 100 |

Exactly +8 tests, no regressions.

**Bisect proof the tests actually pin the fix** — reverting only `lib/valkey/commands/server_commands.rb` while keeping the new tests gives `8 tests, 2 assertions, 2 failures, 6 errors`; the errors are `TypeError: no implicit conversion of Array into Integer` at `server_commands.rb:225` — the exact symptom reported in #215. With the fix applied: `8 tests, 16 assertions, 0 failures, 0 errors`.

**Behavioural coverage verified** — `:len`, `"len"`, `"LEN"`, `:get`, `"GET"`, `:get, 2`, `:reset`, `"RESET"`, and mixed-case `:Get` all dispatch correctly; `:nope`, `:help`, and `nil` all raise `ArgumentError`.

**Deferred** — cluster-mode run, no local cluster available: `bundle exec rake test:cluster`.

### Checklist

-   [x] Documentation is updated (if applicable). <sub>(YARD docs on `slowlog` expanded — params, `@raise`, `@example`, `@see`. No prose docs needed.)</sub>
-   [x] Destination branch is correct - main <sub>(destination is `release-1.0` per the 1.0.0 GA release-blocker track, matching #207, #205, #203, #202, #201, #198, #197, #196, #195.)</sub>

### DIFF
--- a/lib/valkey/commands/server_commands.rb
+++ b/lib/valkey/commands/server_commands.rb
@@ -216,13 +216,30 @@ class Valkey
 
       # Interact with the slowlog (get, len, reset)
       #
-      # @param [String] subcommand e.g. `get`, `len`, `reset`
-      # @param [Integer] length maximum number of entries to return
-      # @return [Array<String>, Integer, String] depends on subcommand
+      # @example Get the number of entries in the slowlog
+      #   valkey.slowlog(:len) # => 0
+      # @example Get the most recent entries
+      #   valkey.slowlog(:get, 2)
+      #
+      # @param [String, Symbol] subcommand one of `get`, `len`, `reset`
+      # @param [Integer] length maximum number of entries to return (`get` only)
+      # @return [Array<Array>, Integer, String] depends on subcommand
+      # @raise [ArgumentError] if the subcommand is not one of `get`, `len`, `reset`
+      #
+      # @see https://valkey.io/commands/slowlog/
       def slowlog(subcommand, length = nil)
-        args = [:slowlog, subcommand]
-        args << Integer(length) if length
-        send_command(args)
+        case subcommand.to_s.downcase
+        when "get"
+          args = []
+          args << Integer(length) if length
+          send_command(RequestType::SLOWLOG_GET, args)
+        when "len"
+          send_command(RequestType::SLOWLOG_LEN)
+        when "reset"
+          send_command(RequestType::SLOWLOG_RESET)
+        else
+          raise ArgumentError, "Unknown SLOWLOG subcommand: #{subcommand}"
+        end
       end
 
       # Internal command used for replication.
--- a/test/lint/server_commands.rb
+++ b/test/lint/server_commands.rb
@@ -64,6 +64,79 @@ module Lint
       assert_equal "OK", response
     end
 
+    def test_slowlog_len
+      skip("SLOWLOG is a per-node command and returns a multi-node response in cluster mode") if cluster_mode?
+
+      assert_kind_of Integer, r.slowlog(:len)
+    end
+
+    def test_slowlog_get
+      skip("SLOWLOG is a per-node command and returns a multi-node response in cluster mode") if cluster_mode?
+
+      assert_kind_of Array, r.slowlog(:get)
+    end
+
+    def test_slowlog_get_with_length
+      skip("SLOWLOG is a per-node command and returns a multi-node response in cluster mode") if cluster_mode?
+
+      assert_kind_of Array, r.slowlog(:get, 2)
+    end
+
+    def test_slowlog_reset
+      skip("SLOWLOG is a per-node command and returns a multi-node response in cluster mode") if cluster_mode?
+
+      assert_equal "OK", r.slowlog(:reset)
+    end
+
+    # Both spellings must dispatch. Asserted against invariant results rather than
+    # by comparing two LEN calls, whose values can legitimately differ if the
+    # slowlog threshold is low enough for the first call to log itself.
+    def test_slowlog_accepts_string_and_symbol_subcommands
+      skip("SLOWLOG is a per-node command and returns a multi-node response in cluster mode") if cluster_mode?
+
+      assert_equal "OK", r.slowlog(:reset)
+      assert_equal "OK", r.slowlog("RESET")
+      assert_kind_of Integer, r.slowlog(:len)
+      assert_kind_of Integer, r.slowlog("LEN")
+    end
+
+    # Exercises the full round trip: that GET actually returns logged entries and
+    # that `length` is passed through as a command argument rather than dropped.
+    def test_slowlog_get_returns_entries_and_honours_length
+      skip("SLOWLOG is a per-node command and returns a multi-node response in cluster mode") if cluster_mode?
+
+      # Fall back to the server default rather than skipping the restore, so a
+      # surprising CONFIG GET shape can never leave this shared server logging
+      # every command for the remainder of the run.
+      original = r.config_get("slowlog-log-slower-than")["slowlog-log-slower-than"] || "10000"
+      mutated = true
+      r.config_set("slowlog-log-slower-than", "0") # log every command
+      r.slowlog(:reset)
+      r.ping
+
+      entries = r.slowlog(:get)
+      refute_empty entries
+      assert_kind_of Array, entries.first
+      assert_equal 1, r.slowlog(:get, 1).length
+    ensure
+      if mutated
+        r.config_set("slowlog-log-slower-than", original)
+        r.slowlog(:reset)
+      end
+    end
+
+    def test_slowlog_unknown_subcommand_raises_argument_error
+      error = assert_raises(ArgumentError) { r.slowlog(:nope) }
+      assert_match(/Unknown SLOWLOG subcommand/, error.message)
+    end
+
+    # SLOWLOG HELP has no RequestType constant, so it is rejected explicitly
+    # rather than surfacing as a confusing TypeError. Use `call("SLOWLOG", "HELP")`
+    # if the raw help text is needed.
+    def test_slowlog_help_raises_argument_error
+      assert_raises(ArgumentError) { r.slowlog(:help) }
+    end
+
     def test_sync
       skip("SYNC command not implemented in backend yet")
 
--- a/test/lint/server_commands.rb
+++ b/test/lint/server_commands.rb
@@ -98,6 +98,8 @@ module Lint
       assert_equal "OK", r.slowlog("RESET")
       assert_kind_of Integer, r.slowlog(:len)
       assert_kind_of Integer, r.slowlog("LEN")
+      assert_kind_of Array, r.slowlog(:get)
+      assert_kind_of Array, r.slowlog("GET")
     end
 
     # Exercises the full round trip: that GET actually returns logged entries and


test/lint/server_commands.rb
```diff
@@ -64,6 +64,79 @@ def test_slaveof
       assert_equal "OK", response
     end
 
+    def test_slowlog_len
+      skip("SLOWLOG is a per-node command and returns a multi-node response in cluster mode") if cluster_mode?
+
+      assert_kind_of Integer, r.slowlog(:len)
```
Aryex, line 70: We should add some sanity check tests for when subcommand is string as well.

lib/valkey/commands/server_commands.rb
```diff
@@ -216,13 +216,30 @@ def slaveof(host, port)
...
+      # @example Get the most recent entries
+      #   valkey.slowlog(:get, 2)
+      #
+      # @param [String, Symbol] subcommand one of `get`, `len`, `reset`
+      # @param [Integer] length maximum number of entries to return (`get` only)
+      # @return [Array<Array>, Integer, String] depends on subcommand
+      # @raise [ArgumentError] if the subcommand is not one of `get`, `len`, `reset`
+      #
+      # @see https://valkey.io/commands/slowlog/
       def slowlog(subcommand, length = nil)
```
Aryex, line 230: Not a blocker: This is redis-rb style command, also GLIDE Ruby would be the first client to implement support for this command.

test/lint/server_commands.rb
jamesx-improving: Good call. `test_slowlog_accepts_string_and_symbol_subcommands` at `test/lint/server_commands.rb:94-103` already covers the string spelling for `reset` and `len`; extended it to also assert `slowlog("GET")` so all three subcommands are exercised in both symbol and string form. Addressed in 5394693.

lib/valkey/commands/server_commands.rb
jamesx-improving: Ack — the redis-rb-style single-method dispatcher is intentional: it matches the two other subcommand dispatchers already in this file (`xinfo`, `xgroup`) and the repo convention of matching redis-rb command shapes for parity. Good to know this closes the parity gap across GLIDE clients as well.

Aryex (approved):

xShinnRyuu (approved):

Issue #206, opened 2026-07-30
Add Script interface (invokeScript / Script class) parity with other GLIDE clients


Aryex:
The Ruby client is missing the `Script` interface / `invokeScript()` method that the Python, Java, Node, Go, and C# clients provide.

Currently, Ruby users can only execute Lua scripts via `eval`, `evalsha`, `eval_ro`, `evalsha_ro`, and `script_load`. There is no reusable `Script` object with an explicit lifecycle (client-side caching handle, `get_hash`, cleanup on release) as offered by the other language clients.

Adding a `Script` class and `invoke_script(script, keys:, args:)` method would bring Ruby to feature parity with the other clients.

### References
- Python: [`Script` class](https://github.com/valkey-io/valkey-glide/blob/main/python/glide-shared/glide_shared/script.py), [`invoke_script`](https://github.com/valkey-io/valkey-glide/blob/main/python/glide-sync/glide_sync/sync_commands/standalone_commands.py)
- Docs page tracking supported clients: https://glide.valkey.io/reference/scripting-reference/

jamesx-improving:
TODO note carried over from PR #227 review (see [thread](https://github.com/valkey-io/valkey-glide-ruby/pull/227#discussion_r3724010738)):

When the `Script` interface lands here, revisit the current wire-command dispatch for `EVAL` / `EVALSHA` / `SCRIPT LOAD`. They're routed via `#call` (custom command) rather than typed `send_command(RequestType::EVAL, ...)` because glide-core defines those enum values but has no `get_command()` arm for them — typed dispatch fails outright with "Couldn't fetch command type - ClientError". Cluster routing is unaffected (redis-rs derives it from the wire command name, so `EVAL` still routes by `ThirdArgAfterKeyCount` and `SCRIPT LOAD` still broadcasts to all nodes), but OpenTelemetry spans are named "CustomCommand" instead of the actual command. Once glide-core gains those arms, these can revert to typed dispatch — `EVAL_RO` / `EVALSHA_RO` already have arms and could use it today; they go through `#call` only to keep the family's behavior uniform.

[...]

Issue #167, opened 2026-07-17
Flaky test: test_cached_script_execution returns binary garbage on Redis 6.2


Aryex:
## Overview

`test_cached_script_execution` intermittently fails with binary data returned from `evalsha` instead of the expected string, and logs a spurious `FUNCTION FLUSH` error on Redis 6.2 where that command doesn't exist.

## Description

The test runs 100 randomized iterations, each loading a script via `script_load` then verifying `evalsha` matches `eval`. On Redis 6.2 CI runs, setup calls `function_flush` (from `FunctionCommands` module) which errors:

```
ResponseError: unknown command `FUNCTION`, with args beginning with: `FLUSH`
```

The actual assertion failure shows raw bytes returned instead of a decoded string:

```
Expected: "ARG0_577"
Actual:   "\v\xC1\xCB\xB0\n\x00\x00\x00"
```

The binary pattern suggests the FFI response is being mis-parsed — returning raw bytes rather than the decoded RESP string — on certain randomized script+args combinations.

## Logs

```
Error executing command: An error was signalled by the server: - ResponseError: unknown command `FUNCTION`, with args beginning with: `FLUSH`, 
    test_cached_script_execution                                    FAIL (0.08s)
          Cached script should be executable multiple times with consistent results.
          --- expected
          +++ actual
          @@ -1,3 +1,4 @@
           # encoding: ASCII-8BIT
           #    valid: true
          -"ARG0_577"
          +"\v\xC1\xCB\xB0
          +\x00\x00\x00"
  /home/runner/work/valkey-glide-ruby/valkey-glide-ruby/test/valkey/eval_evalsha_property_test.rb:335:in 'block in
  ValkeyTests::EvalEvalshaTypeProperties#test_cached_script_execution'
          <internal:numeric>:257:in 'Integer#times'
  /home/runner/work/valkey-glide-ruby/valkey-glide-ruby/test/valkey/eval_evalsha_property_test.rb:293:in
  'ValkeyTests::EvalEvalshaTypeProperties#test_cached_script_execution'
          /home/runner/work/valkey-glide-ruby/valkey-glide-ruby/test/support/helper/generic.rb:19:in 'Helper::Generic#run'
```

## Proposal

[...]

PR #227, opened 2026-08-04
fix: send real SCRIPT LOAD/EVAL/EVALSHA instead of the local cache

### LINKED ISSUE #213: `eval` / `evalsha` silently misassign KEYS and ARGV when given the standard integer key-count argument
# `eval` / `evalsha` silently misassign KEYS and ARGV when given the standard integer key-count argument

## Summary

`eval(script, numkeys, *keys_and_argv)` — the form used by the Valkey documentation, `valkey-cli`, and every other GLIDE binding — is not recognized. The integer is stringified and becomes `KEYS[1]`, the real key shifts into `ARGV[1]`, and any remaining arguments are dropped. **No error is raised.**

Writes therefore land on a key named after the count (`"1"`), and read-compare-delete patterns silently fail their comparison. Failures surface arbitrarily far from the call site.

The positional form was added deliberately in #174 / #202, scoped to the two-array shape `eval(script, keys_array, argv_array)`, which it handles correctly. The integer key-count form was never in scope. `redis-rb` doesn't support it either — but it raises loudly (`NoMethodError: undefined method 'length' for an instance of Integer`), so it fails safe where this client fails silently.

## Scope

This covers the **whole of #216 Group 1** — two defects with one shared root cause, the glide-core local `Script`/`invoke_script` container:

- **R3-4** — `eval` / `evalsha` misassign KEYS/ARGV on the integer key-count form (below).
- **R2-12+13** — `script_load` never sends `SCRIPT LOAD`, so the returned SHA is unknown to the server; and `script_flush` is silently undone, because the next `evalsha` re-uploads from the client-side container instead of raising `NOSCRIPT`.

R2-12+13 is not filed separately. Reproduced on `release-1.0` @ `fb2fa46`:

```ruby
sha = c.script_load("return 'probe'")
c.call("SCRIPT", "EXISTS", sha)   # => [false]      <-- server never received it
c.script_flush
c.evalsha(sha)                    # => "probe"      <-- should raise NOSCRIPT
```

Replacing the container with real wire dispatch fixes all three rows at once; after the change the same probe gives `[true]` and a raised error respectively.

**Note this makes `evalsha` on a flushed or never-loaded SHA raise instead of silently succeeding.** That is the R2-13 fix and matches redis-rb, but it is a behavioural change worth a CHANGELOG note.

*(Corrected: an earlier version of this text said the error is `Valkey::NoScriptError`. It is actually `Valkey::CommandError`, whose message contains `NoScriptError`. `Valkey::NoScriptError` is defined at `lib/valkey/errors.rb:24` but is never raised anywhere in `lib/` — the FFI exposes only four coarse `RequestErrorType` values, so NOSCRIPT arrives as UNSPECIFIED and maps to `CommandError`. Mapping the specific subclasses is a separate pre-existing gap that also affects `PermissionError`, `WrongTypeError` and `OutOfMemoryError`.)*

## Steps to reproduce

```ruby
require "valkey"
c = Valkey.new
c.flushdb

# 1) arguments silently misassigned
p c.eval("return {KEYS[1], ARGV[1]}", 1, "mykey", "myarg")
# => ["1", "mykey"]          expected ["mykey", "myarg"]

# 2) a real write lands on the wrong key
c.eval("redis.call('SET', KEYS[1], ARGV[1])", 1, "realkey", "realval")
p c.get("realkey")   # => nil
p c.get("1")         # => "realkey"
p c.keys("*")        # => ["1"]

# 3) Redlock-style unlock silently never releases
c.set("lock:job", "token-abc")
unlock = "if redis.call('GET', KEYS[1]) == ARGV[1] then return redis.call('DEL', KEYS[1]) else return 0 end"
p c.eval(unlock, 1, "lock:job", "token-abc")   # => 0
p c.get("lock:job")                            # => "token-abc"  (still held)

# 4) extra arguments are dropped
p c.eval("return {KEYS[1], KEYS[2], ARGV[1]}", 2, "k1", "k2", "a1")
# => ["2"]

# 5) evalsha behaves identically
sha = c.script_load("return {KEYS[1], ARGV[1]}")
p c.evalsha(sha, 1, "mykey", "myarg")   # => ["1", "mykey"]

# the supported two-array form is correct
p c.eval("return {KEYS[1], ARGV[1]}", ["mykey"], ["myarg"])
# => ["mykey", "myarg"]
```

## Expected behavior

Either of these would be acceptable; both are safe:

- **Accept the form.** `eval(script, 1, "mykey", "myarg")` treats `1` as the key count, yielding `KEYS = ["mykey"]`, `ARGV = ["myarg"]`.
- **Reject it loudly.** Raise `ArgumentError` naming the supported forms, matching how `redis-rb` fails.

What must not happen is proceeding with a silently different meaning.

## Actual behavior

`keys ||= rest[0] || []` (`scripting_commands.rb:153`, and `:203` for `evalsha`) takes `rest[0]` — the integer — and passes it through `Array(keys).map(&:to_s)`, producing `["1"]`. `rest[1]` becomes ARGV, and `rest[2..]` is discarded.

## Affected code

- `lib/valkey/commands/scripting_commands.rb:145-166` — `eval`
- `lib/valkey/commands/scripting_commands.rb:198-219` — `evalsha`

Introduced by #202 (`92a571d`), the `release-1.0` backport of #174.

## Suggested fix

**Dispatch real wire commands.** `RequestType::Eval` / `EvalSha` / `ScriptLoad` have enum values but no `get_command()` arm in glide-core, so typed dispatch is not available — verified, 3 of 5 fail with `Couldn't fetch command type`. The original code comment ("eval is not available in the rust backend") was accurate; the wrong conclusion was that `invoke_script` was the only alternative. `call("SCRIPT","LOAD",…)` / `call("EVAL",…)` / `call("EVALSHA",…)` is the sanctioned path, and it removes the local container that causes R2-12+13.

Cluster routing is preserved: redis-rs routes `EVAL`/`EVALSHA` by `ThirdArgAfterKeyCount` — the slot comes from the key count this fix starts sending correctly — and broadcasts `SCRIPT LOAD` to all nodes. OpenTelemetry improves, since `invoke_script` hardcoded `span_ptr = 0` and emitted no spans at all.

With the wire form in place, the numkeys parsing below is what turns the caller's arguments into `EVAL`'s wire arguments — support the integer form and validate strictly. `Integer` and `Array` are disjoint in the first positional slot, so accepting the integer form is a strict superset — it cannot change the meaning of any program that works today:

```ruby
def eval(script, *rest, keys: nil, args: nil)
  raise ArgumentError, "script must be a string" unless script.is_a?(String)
  raise ArgumentError, "script cannot be empty" if script.empty?

  if keys.nil? && args.nil? && rest.first.is_a?(Integer)
    numkeys = rest.shift
    raise ArgumentError, "numkeys must be non-negative" if numkeys.negative?
    if numkeys > rest.size
      raise ArgumentError, "numkeys (#{numkeys}) exceeds the number of arguments given (#{rest.size})"
    end
    keys = rest.shift(numkeys)
    args = rest
  else
    keys ||= rest[0] || []
    args ||= rest[1] || []
  end

  # ... unchanged from here
end
```

Same change applies to `evalsha`. Mixing the integer form with `keys:` / `args:` keywords is genuinely ambiguous and should raise rather than pick a winner.

## Test coverage

No existing test passes an integer first. Suggested additions to the scripting lint tests:

```ruby
def test_eval_with_integer_numkeys
  assert_equal ["mykey", "myarg"],
               r.eval("return {KEYS[1], ARGV[1]}", 1, "mykey", "myarg")
end

def test_eval_with_integer_numkeys_zero
  assert_equal ["a1"], r.eval("return {ARGV[1]}", 0, "a1")
end

def test_eval_with_numkeys_exceeding_args_raises
  assert_raises(ArgumentError) { r.eval("return 1", 5, "only-one") }
end

def test_eval_integer_numkeys_writes_to_the_named_key
  r.eval("redis.call('SET', KEYS[1], ARGV[1])", 1, "realkey", "realval")
  assert_equal "realval", r.get("realkey")
  assert_nil r.get("1")
end
```

Plus the `evalsha` equivalents. All 73 existing scripting tests pass unchanged with the above signature.

## Impact

Silent corruption on the canonical scripting signature. Distributed locks, rate limiters, and compare-and-swap scripts are exactly the code written this way, and exactly the code where a silent no-op is most damaging — in cluster mode the misassignment also routes the command to the wrong key's slot. Because the client returns plausible-looking values rather than raising, there is nothing at the call site to suggest anything went wrong.

## Environment

- Ruby: 4.0.6, arm64-darwin25
- Valkey: 8.1.3 (standalone)
- Compared against: `redis-rb` 5.4.1

jamesx-improving:
### Summary

`eval`, `evalsha`, `eval_ro`, `evalsha_ro` and `script_load` never sent their
namesake commands to the server. They went through glide-core's client-side
`Script` container and `invoke_script`, which produced three distinct
user-visible defects. This PR dispatches real `EVAL` / `EVALSHA` / `EVAL_RO` /
`EVALSHA_RO` / `SCRIPT LOAD` over the wire, and adds the standard integer
key-count call form.

One root cause, three findings closed. **This PR contains two breaking
changes** — see below.

### Issue link

Closes #213

Also closes two findings from the #216 Group 1 rows (R2-12, R2-13), which were
not filed separately because they share this root cause.

#### The three findings

- **R3-4 — `KEYS`/`ARGV` misassignment on the integer key-count form.**
  `eval(script, 1, "mykey", "myarg")` — the form used by `valkey-cli` and every
  Valkey doc example — put the *count* in `KEYS[1]`, shifted the real key into
  `ARGV[1]`, and silently dropped everything after it. No error was raised.
- **R2-12 — `script_load` never sent `SCRIPT LOAD`.** It computed a SHA1 into a
  client-local container, so the server had never seen the script. The returned
  SHA was unusable by any other client or process, and `script_exists(sha)`
  returned `false` for a script that had just been "loaded".
- **R2-13 — `script_flush` was silently undone.** Because `evalsha` resolved
  through the client-side container, the next `evalsha` after a flush simply
  re-uploaded the script from the client and succeeded. Flushing the server
  cache had no observable effect.

#### Why wire dispatch instead of typed dispatch

`RequestType::Eval`, `RequestType::EvalSha` and `RequestType::ScriptLoad` have
enum values in glide-core but **no `get_command()` arm**, so typed
`send_command` fails outright with `Couldn't fetch command type`. To be precise:
this is 3 of the 5 methods. `EVAL_RO` and `EVALSHA_RO` *do* have arms
(glide-core `request_type.rs:1291-1292`) and would work with typed dispatch
today. All five are nonetheless routed through `call` to keep the family
uniform — a deliberate choice with a known cost, see *Known cost* below. The
rationale is recorded in a module comment so a future contributor can revert to
typed dispatch once glide-core gains the missing arms.

#### BREAKING: `evalsha` on a flushed/never-loaded SHA now raises

It raises **`Valkey::CommandError`** whose message contains `NoScriptError` —
**not** `Valkey::NoScriptError`. Callers must catch `Valkey::CommandError`.

`Valkey::NoScriptError` is defined at `lib/valkey/errors.rb:24` but is never
raised anywhere in `lib/`. The FFI exposes only four coarse `RequestErrorType`
values (`UNSPECIFIED` / `EXECABORT` / `TIMEOUT` / `DISCONNECT`), so `NOSCRIPT`
arrives as `UNSPECIFIED` and maps to `CommandError`. Verified live against a
real server.

Mapping the specific error subclasses is a **pre-existing gap out of scope for
this PR** — it equally affects `PermissionError`, `WrongTypeError` and
`OutOfMemoryError`, and deserves its own issue. Flagging it here so the
behavior is not mistaken for a regression introduced by this change.

#### BREAKING (smaller): ambiguous argument shapes now raise

Previously silent misparses now raise `ArgumentError`:

- Non-Integer numeric key counts (`Float`, `Rational`) — `eval(s, 2.0, "k1", "k2")`
  used to reproduce the exact bug this PR fixes, with no error at all.
- More than two positional arrays.
- `script_load` with a non-String script.

This follows the issue's own principle: proceeding with a silently different
meaning is precisely the thing to avoid. `Integer` and `Array` are disjoint in
the first positional slot, so recognizing the integer form is a strict superset
— it cannot change the meaning of any call that already worked.

#### A regression found during review, and fixed

Being honest about this one. Because every scripting method now goes through
`call`, which always forwards `route:`, and `Valkey::Pipeline#send_command` did
not accept `route:`, `pipelined { |p| p.script_load(...) }` **regressed from
returning `[]` to raising `ArgumentError: wrong number of arguments (given 3,
expected 1..2)`**.

Fixed by adding `route: nil` to `Pipeline#send_command`. The keyword is
**accepted and ignored**, matching the approach in #218 — batch-level routing
is a real feature but needs one route for the whole pipeline (see FFI
`BatchOptionsInfo.route_info`), not per-command routes. That work is tracked
in #137 and is out of scope here.

As a bonus, this also fixes a pre-existing case: `pipelined { |p| p.dbsize; p.info }`
used to raise `ArgumentError` on baseline because ~14 admin commands in
`server_commands.rb` forward `route:` unconditionally into `Pipeline#send_command`.
Verified: it now returns `[Integer, Hash]` cleanly. Non-block `MULTI` also
benefits — `eval` inside `multi { }` no longer executes outside the transaction.

Scope note: `eval`-in-pipeline was already broken before this branch
(`NoMethodError` on `build_command_args`), so that part is not a regression —
only `script_load` was. An earlier revision of this PR raised `ArgumentError`
when a route was supplied; changed to align with #218 after confirming no test
depends on either behavior.

#### Known cost — reviewer's call

OpenTelemetry spans for all five scripting methods are now named
`CustomCommand` rather than e.g. `EVAL_RO` (measured, not assumed).
`EVAL_RO` / `EVALSHA_RO` could retain real span names via typed dispatch today;
they were left uniform with the rest of the family and this is documented
in-code. Happy to split them out if maintainers prefer accurate span names over
family uniformity.

#### Tension with #206 — flagging, not resolving

#206 proposes a `Script` class; this PR moves away from the client-side
container. These are **not fundamentally in conflict** — `invoke_script` is
retained as the FFI seam. But #206 needs to know:

- `script_load` no longer populates glide-core's `scripts_container`, so
  `invoke_script`'s NOSCRIPT auto-reload path is now unreachable. A `Script`
  class must either call `store_script` itself, or be built on `EVALSHA` +
  `SCRIPT LOAD` fallback in Ruby.
- `Bindings.drop_script`, `Bindings.store_script`,
  `Bindings.free_script_hash_buffer` and `Bindings::ScriptHashBuffer` are now
  entirely unreferenced in `lib/` (`drop_script` was already dead before this
  branch). This PR enlarges the dead FFI surface. Left in place deliberately,
  since #206 may want it.

#### Cluster: reasoned, not locally tested

Routing is preserved because redis-rs derives it from the wire command name via
`Routable::command()` reading `arg_idx(0)`, and `CustomCommand` builds an empty
`Cmd` with args appended. So `EVAL` → `ThirdArgAfterKeyCount` (slot derived from
the key count this PR now sends correctly), `EVAL_RO` → replica-optional,
`SCRIPT LOAD` → `AllNodes` / `AllSucceeded`. Verified by reading glide-core
source. Actual cluster validation is deferred to CI — stating that plainly
rather than claiming coverage I don't have.

#### Files changed

| File | Change |
| --- | --- |
| `lib/valkey/commands/scripting_commands.rb` | Wire dispatch for all five methods; new `split_keys_and_args` helper handling all three call shapes; module comment recording the typed-dispatch constraint |
| `lib/valkey/pipeline.rb` | `send_command` accepts `route:` (raises if non-nil) — fixes the regression above |
| `test/lint/scripting_commands.rb` | 26 new tests |
| `CHANGELOG.md` | `## Pending` entry, including both breaking changes |

### Testing

**Lint** — `bundle exec rubocop lib/valkey/commands/scripting_commands.rb lib/valkey/pipeline.rb test/lint/scripting_commands.rb` → **0 offenses**.

**Standalone suite** — `CI=1 SKIP_TLS_TESTS=true VALKEY_PORT=6413 bundle exec rake test:standalone`
against a dedicated Valkey 8.1.3 instance:

| Run | Tests | Assertions | Failures |
| --- | --- | --- | --- |
| Baseline (pristine `fb2fa46`) | 840 | 4187 | 0 |
| With this change | **866** | 4207 | **0** |
| Repeat run | 866 | 4195 | 0 |

`test_cached_script_execution` — the known pre-existing flake (#167 / #124) —
passed in both final runs.

**The R2-12 / R2-13 tests assert real server state, not client-side echoes:**

- `test_script_load_makes_the_script_available_on_the_server`
- `test_script_loaded_by_one_client_is_usable_by_another` — cross-client, which
  is what actually proves the SHA reached the server
- `test_script_flush_really_invalidates_the_script` — asserts the raise **and**
  re-asserts `script_exists == false`, proving no client-side re-upload
- `test_evalsha_raises_for_a_never_loaded_sha`

All new tests live in the shared lint module, so they run in **both** the
standalone and cluster suites.

**Deferred:** `bundle exec rake test:cluster` — to CI (see *Cluster* above).

**Performance:** the warm path drops from 3 round trips to 1.

### Checklist

-   [x] Destination branch is correct - main _(destination here is `release-1.0`, per the 1.0.0 GA release-blocker track)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,5 +9,9 @@
 ### Changes
 
 * Ruby: fixed cd workflow to correctly build the ffi with **glibc 2.17** ([#223](https://github.com/valkey-io/valkey-glide-ruby/issues/223))
+* Ruby: **Breaking** — scripting commands now dispatch real `EVAL` / `EVALSHA` / `SCRIPT LOAD` to the server instead of a client-side script container ([#213](https://github.com/valkey-io/valkey-glide-ruby/issues/213)). Three consequences:
+  * `eval` / `evalsha` (and the `_ro` variants) now accept the standard integer key-count form used by `valkey-cli` and the Valkey docs — `eval(script, 1, "mykey", "myarg")`. It previously made the count `KEYS[1]`, shifted the real key into `ARGV[1]`, and dropped the remaining arguments without raising.
+  * `script_load` now really sends `SCRIPT LOAD`, so the returned SHA1 is known to the server and usable by `evalsha` from any other client or process. `script_exists` previously reported `false` for a just-loaded script.
+  * `evalsha` on a flushed or never-loaded SHA now raises `Valkey::CommandError` (NOSCRIPT) instead of silently re-uploading the script and succeeding, so `script_flush` is no longer quietly undone. Callers relying on the old auto-reload must load the script again after a flush, or use `eval`.
 * Ruby: Add Alpine Linux (musl libc) support for x86_64 and aarch64 — runtime detection of musl libc, CI/CD pipeline for native builds, and prebuilt `libglide_ffi.so` for musl targets ([#143](https://github.com/valkey-io/valkey-glide-ruby/pull/143))
 * Ruby: Add distributed tracing support — `Valkey::OpenTelemetry.set_parent_span_context_provider` (and `init(parent_span_context_provider:)`) let an app propagate its current W3C trace context into command/pipeline spans, so they become children of the app's trace instead of independent root spans, matching the Node.js client's `parentSpanContextProvider` behavior.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Changes
 
 * Ruby: fixed cd workflow to correctly build the ffi with **glibc 2.17** ([#223](https://github.com/valkey-io/valkey-glide-ruby/issues/223))
-* Ruby: **Breaking** — scripting commands now dispatch real `EVAL` / `EVALSHA` / `SCRIPT LOAD` to the server instead of a client-side script container ([#213](https://github.com/valkey-io/valkey-glide-ruby/issues/213)). Three consequences:
+* Ruby: scripting commands now dispatch real `EVAL` / `EVALSHA` / `SCRIPT LOAD` to the server instead of a client-side script container ([#213](https://github.com/valkey-io/valkey-glide-ruby/issues/213)). Three behavior changes:
   * `eval` / `evalsha` (and the `_ro` variants) now accept the standard integer key-count form used by `valkey-cli` and the Valkey docs — `eval(script, 1, "mykey", "myarg")`. It previously made the count `KEYS[1]`, shifted the real key into `ARGV[1]`, and dropped the remaining arguments without raising.
   * `script_load` now really sends `SCRIPT LOAD`, so the returned SHA1 is known to the server and usable by `evalsha` from any other client or process. `script_exists` previously reported `false` for a just-loaded script.
   * `evalsha` on a flushed or never-loaded SHA now raises `Valkey::CommandError` (NOSCRIPT) instead of silently re-uploading the script and succeeding, so `script_flush` is no longer quietly undone. Callers relying on the old auto-reload must load the script again after a flush, or use `eval`.
--- a/lib/valkey/commands/scripting_commands.rb
+++ b/lib/valkey/commands/scripting_commands.rb
@@ -4,6 +4,18 @@ class Valkey
   module Commands
     # this module contains commands related to list data type.
     #
+    # EVAL / EVALSHA / SCRIPT LOAD are dispatched as custom commands via #call
+    # rather than through typed `send_command(RequestType::EVAL, ...)`. This is
+    # not a style choice: glide-core defines those enum values but has no
+    # `get_command()` arm for them, so typed dispatch fails outright with
+    # "Couldn't fetch command type - ClientError". Cluster routing is unaffected
+    # (redis-rs derives it from the wire command name, so EVAL still routes by
+    # ThirdArgAfterKeyCount and SCRIPT LOAD still broadcasts to all nodes), but
+    # OpenTelemetry spans are named "CustomCommand" instead of the command.
+    # Once glide-core gains those arms, these can revert to typed dispatch -
+    # EVAL_RO / EVALSHA_RO already have arms and could use it today; they go
+    # through #call only to keep the family's behavior uniform.
+    #
     # @see https://valkey.io/commands/#scripting
     #
     module ScriptingCommands
@@ -100,20 +112,31 @@ class Valkey
         send_command(RequestType::SCRIPT_DEBUG, [mode.to_s.upcase])
       end
 
+      # Load a Lua script into the server's script cache without executing it.
+      #
+      # Sends a real `SCRIPT LOAD` to the server, so the returned SHA1 is
+      # immediately usable by `evalsha` - including from a different client,
+      # process, or worker. In cluster mode `SCRIPT LOAD` is routed to all
+      # nodes, so the script is available whichever node a later `EVALSHA`
+      # lands on.
+      #
+      # @param [String] script the Lua script to load
+      # @return [String] the SHA1 hash of the script, as computed by the server
+      #
+      # @example
+      #   sha = valkey.script_load("return 1")
+      #   valkey.script_exists(sha)   # => true
+      #
+      # @see https://valkey.io/commands/script-load/
       def script_load(script)
         script = script.first if script.is_a?(Array)
 
-        buf = FFI::MemoryPointer.new(:char, script.bytesize)
-        buf.put_bytes(0, script)
+        # Validate here rather than letting a non-String flatten into extra
+        # SCRIPT LOAD wire arguments and fail as a server-side arity error.
+        raise ArgumentError, "script must be a string" unless script.is_a?(String)
+        raise ArgumentError, "script cannot be empty" if script.empty?
 
-        result = Bindings.store_script(buf, script.bytesize)
-
-        begin
-          hash_buffer = Bindings::ScriptHashBuffer.new(result)
-          hash_buffer[:ptr].read_string(hash_buffer[:len])
-        ensure
-          Bindings.free_script_hash_buffer(result) if result && !result.null?
-        end
+        call("SCRIPT", "LOAD", script)
       end
 
       # Execute a Lua script on the server.
@@ -140,30 +163,17 @@ class Valkey
       # @example Positional form, matching redis-rb's eval(script, keys, argv)
       #   valkey.eval("return KEYS[1] .. ARGV[1]", ["mykey"], ["myarg"])
       #     # => "mykeynyarg"
-      # Since the eval is not available in the rust backend
-      # using the load and invoke script
+      # @example Integer key-count form, matching valkey-cli and the Valkey docs
+      #   valkey.eval("return {KEYS[1], ARGV[1]}", 1, "mykey", "myarg")
+      #     # => ["mykey", "myarg"]
       def eval(script, *rest, keys: nil, args: nil)
         # Validate script parameter
         raise ArgumentError, "script must be a string" unless script.is_a?(String)
         raise ArgumentError, "script cannot be empty" if script.empty?
 
-        # Accept redis-rb's flexible positional form - eval(script, keys, argv)
-        # - in addition to the keyword form, mirroring the same treatment
-        # applied to evalsha (see that method for the full rationale).
-        keys ||= rest[0] || []
-        args ||= rest[1] || []
+        keys, args = split_keys_and_args(rest, keys, args)
 
-        # Validate and convert keys and args to strings
-        begin
-          keys = Array(keys).map(&:to_s)
-          args = Array(args).map(&:to_s)
-        rescue StandardError => e
-          raise ArgumentError, "failed to convert keys or args to strings: #{e.message}"
-        end
-
-        # Load script to get SHA1 hash, then execute via invoke_script
-        sha = script_load(script)
-        invoke_script(sha, keys: keys, args: args)
+        call("EVAL", script, keys.size, *keys, *args)
       end
 
       # Execute a cached Lua script by its SHA1 hash.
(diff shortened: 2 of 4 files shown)


lib/valkey/commands/scripting_commands.rb
```diff
@@ -4,6 +4,18 @@ class Valkey
...
+    # rather than through typed `send_command(RequestType::EVAL, ...)`. This is
+    # not a style choice: glide-core defines those enum values but has no
+    # `get_command()` arm for them, so typed dispatch fails outright with
+    # "Couldn't fetch command type - ClientError". Cluster routing is unaffected
+    # (redis-rs derives it from the wire command name, so EVAL still routes by
+    # ThirdArgAfterKeyCount and SCRIPT LOAD still broadcasts to all nodes), but
+    # OpenTelemetry spans are named "CustomCommand" instead of the command.
+    # Once glide-core gains those arms, these can revert to typed dispatch -
+    # EVAL_RO / EVALSHA_RO already have arms and could use it today; they go
+    # through #call only to keep the family's behavior uniform.
```
Aryex: I think the issue here is that GLIDE clients does support custom script executions often through the `Script` class implementation, which the current Ruby client does not implement.

lib/valkey/commands/scripting_commands.rb
```diff
@@ -258,18 +257,29 @@ def eval_ro(script, keys: [], args: [])
...
 
+      # Execute a cached script via the glide-ffi `invoke_script` entry point.
+      #
+      # No longer used by eval/evalsha - they dispatch real wire commands now.
+      # Retained as the FFI seam for the `Script` object proposed in #206, and
+      # still exercised by the scripting tests. Note that its glide-core
+      # NOSCRIPT self-heal relies on the client-side script container that
+      # `script_load` no longer populates, so a SHA this client never stored
+      # will surface a NOSCRIPT CommandError rather than being re-uploaded.
+      # Emits no OpenTelemetry span (span_ptr is hardcoded to 0 below).
```
Aryex: I dont think we need this comment here.

CHANGELOG.md
```diff
@@ -4,5 +4,9 @@
 
 ### Changes
 
+* Ruby: **Breaking** — scripting commands now dispatch real `EVAL` / `EVALSHA` / `SCRIPT LOAD` to the server instead of a client-side script container ([#213](https://github.com/valkey-io/valkey-glide-ruby/issues/213)). Three consequences:
```
Aryex: nit: Should we mention breaking? It is pre GA.

lib/valkey/commands/scripting_commands.rb
Aryex: AI tends to just spits out large comments blocks. I think we could remove this and instead add a TODO to #206

jamesx-improving: Done in `1d7aade` — dropped the module-level block. I'll leave a TODO on #206 capturing the "revert to typed dispatch once glide-core has `get_command()` arms" context so it's not lost. Good call — you're right that it belonged there rather than as a header on the file.

lib/valkey/commands/scripting_commands.rb
jamesx-improving: Done in `1d7aade` — dropped the docblock. I'll fold the "retained as the FFI seam for the `Script` object" note into the TODO on #206.

CHANGELOG.md
jamesx-improving: Dropped the "Breaking" marker in `1d7aade` — agreed, we're pre-GA and the bullet list underneath already spells out the concrete behavior changes for anyone scanning the entry.

Aryex (approved):

yipin-chen (approved):

PR #229, opened 2026-08-05
fix(ruby): default `xtrim` to exact trimming and support `LIMIT`

### LINKED ISSUE #228: `xtrim(key, N)` never trims — `approximate: true` default silently no-ops on small streams
# `xtrim(key, N)` never trims — `approximate: true` default silently no-ops on small streams

## Summary

`Valkey#xtrim` defaults to `approximate: true` (`lib/valkey/commands/stream_commands.rb:232`), which sends `XTRIM ... MAXLEN ~ N` to the server. On any stream smaller than a single radix-tree node (~100 entries — the exact boundary is server-internal), the server answers **`0`** and removes nothing. The Ruby call returns `0` with no error; the stream is unchanged.

Two things make this land badly for users:

1. **redis-rb defaults to `approximate: false`.** Any code ported over — the mainstream path onto this gem — silently changes meaning: `xtrim("s", 5)` used to trim, now it doesn't.
2. **The command name is `xtrim`.** A no-op result on a stream that visibly has more than `N` entries is not something a reader would suspect without knowing about the `~` operator's server-side threshold.

Also missing: the `limit:` kwarg. `XTRIM MAXLEN ~ N LIMIT M` is the standard way to cap trim work per call in production, and the wrapper has no way to send it today.

## Steps to reproduce

```ruby
require "valkey"

v = Valkey.new
v.del("s")
30.times { |i| v.xadd("s", { "n" => i.to_s }) }
v.xlen("s")                              # => 30

v.xtrim("s", 5)                          # => 0     (BUG — nothing trimmed)
v.xlen("s")                              # => 30    (still 30)

v.call("XTRIM", "s", "MAXLEN", 5)        # => 25    (server is fine)
v.xlen("s")                              # => 5

# Exact form also works — proves the wrapper *can* reach the server correctly:
v.del("s")
30.times { |i| v.xadd("s", { "n" => i.to_s }) }
v.xtrim("s", 5, approximate: false)      # => 25
v.xlen("s")                              # => 5
```

3/3 reproducible. `approximate: true` on `MINID` behaves the same way for the same reason.

## Expected behavior

`xtrim("s", 5)` — the idiomatic short form, matching redis-rb — trims the stream to 5 entries and returns the number removed. To opt into approximate/faster trimming on very large streams, the caller passes `approximate: true` explicitly.

## Actual behavior

`xtrim("s", 5)` sends `XTRIM s MAXLEN ~ 5`. On small streams, the server returns `0` and leaves the stream at 30. The user has no signal that trimming did not happen — no exception, no warning, and the returned Integer looks like a normal successful reply.

## Root cause

`stream_commands.rb:232`:

```ruby
def xtrim(key, maxlen, strategy: "MAXLEN", approximate: true)
  args = [key, strategy]
  args << "~" if approximate
  args << maxlen.to_s
  send_command(RequestType::X_TRIM, args)
end
```

Two defects in the signature:

- `approximate: true` — should be `false` for redis-rb parity and for the least-surprising default.
- No `limit:` — `LIMIT M` is only meaningful with `~`, but with `approximate:` supported the wrapper should also accept `limit:`.

## Suggested fix

```ruby
def xtrim(key, maxlen, strategy: "MAXLEN", approximate: false, limit: nil)
  raise ArgumentError, "limit: requires approximate: true" if limit && !approximate

  args = [key, strategy]
  args << "~" if approximate
  args << maxlen.to_s
  args << "LIMIT" << limit.to_s if limit
  send_command(RequestType::X_TRIM, args)
end
```

Behavioral changes:

- Default flips to `approximate: false` — matches redis-rb, matches the natural reading of `xtrim("s", 5)`.
- `limit:` kwarg added, gated behind `approximate: true` (server rejects `LIMIT` without `~`, so pre-empt that with a clearer Ruby-side error).

This is technically a breaking change for any caller that was relying on the current default *and* okay with silent no-ops on small streams — a combination that doesn't correspond to any real use case. Callers who explicitly wanted approximate trimming will have passed `approximate: true` and are unaffected.

## Test coverage

`test/lint/stream_commands.rb:119` already tests `xtrim` — but with `approximate: false` explicitly, so it never caught the default. Add:

```ruby
def test_xtrim_default_trims_small_streams
  r.del "s"
  10.times { |i| r.xadd("s", { "n" => i.to_s }) }
  removed = r.xtrim("s", 3)                 # default form
  assert_operator removed, :>=, 1           # server may return 0 only if nothing to trim
  assert_equal 3, r.xlen("s")
end

def test_xtrim_limit_requires_approximate
  assert_raises(ArgumentError) { r.xtrim("s", 3, limit: 10) }
end

def test_xtrim_with_limit
  r.del "s"
  30.times { |i| r.xadd("s", { "n" => i.to_s }) }
  r.xtrim("s", 5, approximate: true, limit: 100) # runs, does not raise
end
```

## Impact

Every `xtrim` caller relying on the wrapper's default is running with **unbounded stream growth** — the exact behavior the command exists to prevent. Streams are commonly used for retention-capped work queues and event logs, where silent no-op trimming is the worst possible failure mode: memory keeps climbing, no error, no metric.

Discoverability is bad too. `xtrim` returning `0` on a stream that clearly has more entries than `maxlen` will read as "server bug" or "wrong maxlen" before it reads as "wrapper default is `~`".

## Environment

- Ruby: 4.0.6, arm64-darwin25
- Valkey: 8.1.3 (standalone)

jamesx-improving:
### Summary

`Valkey#xtrim` currently defaults `approximate:` to `true`, so `xtrim(key, N)` sends `MAXLEN ~ N` — which is silently a no-op on small streams (below the server's radix-tree node boundary, ~100 entries). Flip the default to `false` for redis-rb parity and least-surprising behavior, and add a `limit:` kwarg (gated on `approximate: true`) so callers can send the standard production form `MAXLEN ~ N LIMIT M`.

### Issue link

Closes #228

### Features / Changes

- `lib/valkey/commands/stream_commands.rb` — `xtrim` signature changes from `approximate: true` to `approximate: false, limit: nil`. When `limit:` is set without `approximate: true`, raise `ArgumentError` — the server rejects `LIMIT` without `~`, so pre-empt with a clearer Ruby-side error. Append `LIMIT M` to args when set. YARD docs and examples updated.
- `test/lint/stream_commands.rb` — three new tests:
  - `test_xtrim_default_trims_small_streams` — bare `xtrim(key, N)` form actually trims a small stream (regression guard against the `~` default returning).
  - `test_xtrim_limit_requires_approximate` — `limit:` without `approximate: true` raises `ArgumentError`.
  - `test_xtrim_with_limit` — `xtrim(key, N, approximate: true, limit: M)` runs and returns an Integer.

**Behavioral note:** technically a breaking change for any caller relying on the current default *and* okay with silent no-ops on small streams — a combination that doesn't correspond to any real use case. Callers who explicitly wanted approximate trimming will have passed `approximate: true` and are unaffected.

### Testing

Ran against a local Valkey 8.1.3 standalone on `localhost:6379`:

- `bundle exec rubocop lib/valkey/commands/stream_commands.rb test/lint/stream_commands.rb` — 2 files inspected, no offenses.
- `bundle exec ruby -Itest test/standalone/commands_test.rb -n "/xtrim/"` — 4 tests, 7 assertions, 0 failures.
- Broader stream regression check with `-n "/xadd|xtrim|xlen|xdel|xrange|xrevrange/"` — 10 tests, 32 assertions, 0 failures.

Full `bundle exec rake test:standalone` and cluster tests deferred to CI.

### Checklist

-   [x] Destination branch is correct - release-1.0

### DIFF
--- a/test/lint/stream_commands.rb
+++ b/test/lint/stream_commands.rb
@@ -135,6 +135,36 @@ module Lint
       r.del "mystream"
     end
 
+    def test_xtrim_default_trims_small_streams
+      r.del "mystream"
+      10.times { |i| r.xadd("mystream", { "n" => i.to_s }) }
+
+      # Default form must actually trim on small streams — approximate should
+      # default to false so `MAXLEN ~ N` is not silently a no-op below the
+      # server's radix-tree node boundary.
+      removed = r.xtrim("mystream", 3)
+      assert_operator removed, :>=, 1
+      assert_equal 3, r.xlen("mystream")
+
+      r.del "mystream"
+    end
+
+    def test_xtrim_limit_requires_approximate
+      assert_raises(ArgumentError) { r.xtrim("mystream", 3, limit: 10) }
+    end
+
+    def test_xtrim_with_limit
+      r.del "mystream"
+      30.times { |i| r.xadd("mystream", { "n" => i.to_s }) }
+
+      # Should run without raising; exact removed count is server-dependent
+      # under `~` semantics, so only assert on the return type.
+      removed = r.xtrim("mystream", 5, approximate: true, limit: 100)
+      assert_kind_of Integer, removed
+
+      r.del "mystream"
+    end
+
     def test_xread
       # Clean up any existing stream first
       r.del "mystream"
--- a/lib/valkey/commands/stream_commands.rb
+++ b/lib/valkey/commands/stream_commands.rb
@@ -220,19 +220,25 @@ class Valkey
       # @param [Integer] maxlen maximum length of the stream
       # @param [Hash] options trimming options
       #   - `:strategy => String`: trimming strategy (default: "MAXLEN")
-      #   - `:approximate => true`: use approximate trimming (default: true)
+      #   - `:approximate => Boolean`: use approximate trimming (default: false)
+      #   - `:limit => Integer`: cap trim work per call; only valid with `approximate: true`
       # @return [Integer] number of entries removed
       #
-      # @example Trim to maximum length
+      # @example Trim to maximum length (exact)
       #   valkey.xtrim("mystream", 1000)
-      # @example Trim with exact count
-      #   valkey.xtrim("mystream", 1000, approximate: false)
+      # @example Approximate trimming
+      #   valkey.xtrim("mystream", 1000, approximate: true)
+      # @example Approximate trimming with LIMIT
+      #   valkey.xtrim("mystream", 1000, approximate: true, limit: 100)
       #
       # @see https://valkey.io/commands/xtrim/
-      def xtrim(key, maxlen, strategy: "MAXLEN", approximate: true)
+      def xtrim(key, maxlen, strategy: "MAXLEN", approximate: false, limit: nil)
+        raise ArgumentError, "limit: requires approximate: true" if limit && !approximate
+
         args = [key, strategy]
         args << "~" if approximate
         args << maxlen.to_s
+        args << "LIMIT" << limit.to_s if limit
         send_command(RequestType::X_TRIM, args)
       end
 


xShinnRyuu (approved):

test/lint/stream_commands.rb
```diff
@@ -135,6 +135,36 @@ def test_xtrim
...
 
+    def test_xtrim_default_trims_small_streams
+      r.del "mystream"
+      10.times { |i| r.xadd("mystream", { "n" => i.to_s }) }
+
+      # Default form must actually trim on small streams — approximate should
+      # default to false so `MAXLEN ~ N` is not silently a no-op below the
+      # server's radix-tree node boundary.
+      removed = r.xtrim("mystream", 3)
+      assert_operator removed, :>=, 1
```
Aryex, line 146: Let's compare this to exact 3.

test/lint/stream_commands.rb
```diff
@@ -135,6 +135,36 @@ def test_xtrim
...
+    end
+
+    def test_xtrim_with_limit
+      r.del "mystream"
+      30.times { |i| r.xadd("mystream", { "n" => i.to_s }) }
+
+      # Should run without raising; exact removed count is server-dependent
+      # under `~` semantics, so only assert on the return type.
+      removed = r.xtrim("mystream", 5, approximate: true, limit: 100)
+      assert_kind_of Integer, removed
```
Aryex, line 163: Can we assert this equals to 5?

yipin-chen (approved):

PR #244, opened 2026-08-06
fix: remove route parameters from commands and pipeline API

Aryex:
## Summary
Two coordinated cleanups so Ruby's cluster/function routing surface matches Python.

### Base `Commands` — 8 methods no longer accept `route:`
- `cluster_info`, `cluster_links`, `cluster_myid`, `cluster_myshardid`, `cluster_nodes`, `cluster_shards` — Python's cluster client exposes no `route`-taking wrappers for these; users who need routing should use `call("CLUSTER", "<subcommand>", route: …)`.
- `fcall`, `fcall_ro` — Python's keyed `fcall`/`fcall_ro` do not take a route (only the keyless `fcall_route`/`fcall_ro_route` variants do). The `args:` kwarg is also renamed to `arguments:` for Python parity. Route support here is tracked in #249.
### Pipeline — `route:` stripped from method signatures via override adapters
The shared `Commands` module carries `route:` on ~29 remaining methods (bgsave, info, dbsize, ping, function_*, call, etc.) so callers of the base `Valkey` client can route those commands to a specific cluster node. Because `Pipeline` mixes in the same module, IDE autocomplete on a pipeline object advertises `route:` on all of those commands — where passing it does nothing. Batch-level routing is separately tracked in #137.
This PR adds signature adapters on `Pipeline` that strip the `route:` kwarg so `Pipeline.instance_method(:bgsave).parameters` no longer advertises it. Each adapter forwards positionally to `super`, so all behavior comes from the shared `Commands` implementation — no duplicated logic.
## Behavior change
- Passing `route:` inside a `pipelined { |p| ... }` or `multi { |m| ... }` block now **raises `ArgumentError`** instead of being silently ignored. Pre-GA is the correct window for this change; the previous behavior was a footgun (no signal that the kwarg didn't work).
- `Valkey#cluster_info(route: …)` and the other 5 `cluster_*` methods now raise `ArgumentError`. Use `call("CLUSTER", "<subcommand>", route: …)` instead.
- `Valkey#fcall(function, keys: […], args: […])` — the `args:` kwarg is renamed to `arguments:`. `route:` is no longer accepted; route support tracked in #249.
## Non-goals
- Does not implement batch-level `route:` on `pipelined`/`multi` — that's #137, and matches the peer clients' design (Python `ClusterBatchOptions.route`, Node `ClusterBatchOptions.route`, Java `ClusterBatchOptions.route`, Go `ClusterBatchOptions.WithRoute`).
- Does not touch the base `Valkey` client's `route:` support on methods where Python legitimately exposes it — those 29 methods still accept `route:` on the non-batch path.
- `Pipeline#send_command` still accepts `route:` because that path is internal — the underlying `Commands` methods forward `route:` to `send_command` unconditionally, and the adapters guarantee it always arrives as `nil` from user-facing entry points.

### DIFF
--- a/lib/valkey/commands/function_commands.rb
+++ b/lib/valkey/commands/function_commands.rb
@@ -182,40 +182,38 @@ class Valkey
       # Invoke a function.
       #
       # @example Call a function
-      #   valkey.fcall("myfunc", keys: ["key1"], args: ["arg1"])
+      #   valkey.fcall("myfunc", keys: ["key1"], arguments: ["arg1"])
       #     # => <function result>
       # @example Call a function without keys
-      #   valkey.fcall("myfunc", args: ["arg1", "arg2"])
+      #   valkey.fcall("myfunc", arguments: ["arg1", "arg2"])
       #     # => <function result>
       #
       # @param [String] function the function name
       # @param [Array<String>] keys the keys to pass to the function
-      # @param [Array<String>] args the arguments to pass to the function
-      # @param route [Valkey::Route, nil] cluster routing. When routed, may return a Hash of node => value.
+      # @param [Array<String>] arguments the arguments to pass to the function
       # @return [Object] the function result
       #
       # @see https://valkey.io/commands/fcall/
-      def fcall(function, keys: [], args: [], route: nil)
-        command_args = [function, keys.size] + keys + args
-        send_command(RequestType::FCALL, command_args, route: route)
+      def fcall(function, keys: [], arguments: [])
+        command_args = [function, keys.size] + keys + arguments
+        send_command(RequestType::FCALL, command_args)
       end
 
       # Invoke a read-only function.
       #
       # @example Call a read-only function
-      #   valkey.fcall_ro("myfunc", keys: ["key1"], args: ["arg1"])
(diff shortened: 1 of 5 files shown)


lib/valkey/commands/function_commands.rb
```diff
@@ -182,40 +182,38 @@ def function_stats(route: nil)
...
-      # @param [Array<String>] args the arguments to pass to the function
-      # @param route [Valkey::Route, nil] cluster routing. When routed, may return a Hash of node => value.
+      # @param [Array<String>] arguments the arguments to pass to the function
       # @return [Object] the function result
       #
       # @see https://valkey.io/commands/fcall/
-      def fcall(function, keys: [], args: [], route: nil)
-        command_args = [function, keys.size] + keys + args
-        send_command(RequestType::FCALL, command_args, route: route)
+      def fcall(function, keys: [], arguments: [])
```
Aryex, line 197: These param renames are to match `redis-rb` interface

currantw (approved):
Looks good to me. ✅

yipin-chen (approved):
LGTM

PR #245, opened 2026-08-07
fix(ruby): return stream entry fields as Hash to match redis-rb

### LINKED ISSUE #240: Stream commands (`xrange` / `xrevrange` / `xread` / `xreadgroup` / `xclaim`) return entry fields as a flat `Array`, not a `Hash` — sole outlier across all GLIDE bindings
## Summary

Every stream-read command returns each entry's field-value pairs as a flat
`Array` (`["name","alice","age","30"]`) instead of the `Hash`
(`{"name"=>"alice","age"=>"30"}`) that `redis-rb` returns and that the Java,
Python, and Go GLIDE bindings return. The Ruby `Utils::HashifyStreamEntries`
lambda explicitly calls `.flatten` on the pair-of-pairs shape glide-core
produces, discarding the structure the core already assembled.

Consequences on the caller side:

- By-name access (`entry["name"]`) raises `TypeError: no implicit conversion
  of String into Integer` — the flat array indexes by integer.
- Pair-iteration works but silently returns `nil` on odd-length data if any
  field value happens to be `nil` (the fault mode changes silently).
- Every consumer must reshape with `each_slice(2).to_h` at the call site.
- The `flatten_map: true` client option does *not* restore parity — it
  returns pair-of-pairs, not a `Hash`.

Same root cause affects `HashifyStreamAutoclaim` (adjacent finding, not in
the CSV): `xautoclaim` returns `{"next"=>"0-0","entries"=>[]}` because its
`reply[1].is_a?(Array)` check misses the raw `Hash` shape, so `xautoclaim`
silently drops all claimed entries while server-side ownership actually
transfers. Filing here as a related repro; consider covering both in one PR.

## Steps to reproduce

```ruby
require "valkey"

v = Valkey.new
v.del("s")
v.xadd("s", { "name" => "alice", "age" => "30" })
v.xadd("s", { "name" => "bob",   "age" => "40" })

entries = v.xrange("s", "-", "+")
entries
# => [
#      ["1735000000000-0", ["name", "alice", "age", "30"]],
#      ["1735000000001-0", ["name", "bob",   "age", "40"]]
#    ]                                           (BUG — flat Array)

entries.first[1]["name"]     # => TypeError: no implicit conversion of String into Integer  (BUG)

# What redis-rb 5.4.1 returns for the same call:
# [["1735000000000-0", {"name"=>"alice","age"=>"30"}],
#  ["1735000000001-0", {"name"=>"bob",  "age"=>"40"}]]

# flatten_map: true does not restore parity:
v2 = Valkey.new(flatten_map: true)
v2.xrange("s", "-", "+")
# => [["...", [["name","alice"],["age","30"]]], ...]   (pair-of-pairs, still not a Hash)
```

Same shape defect on `xrevrange`, `xread`, `xreadgroup`, and `xclaim`.

## Multi-way validation

| Source                              | One entry's fields                                | Evidence |
|-------------------------------------|---------------------------------------------------|----------|
| Raw wire (RESP2 & RESP3)            | `["name","alice","age","30"]` flat                | `valkey-cli`, both protocols |
| glide-core declared type            | `Map{{ id -> ArrayOfPairs }}`                     | `value_conversion.rs:1646` |
| What core hands to Ruby             | `{"1-1"=>[["name","alice"],["age","30"]]}`        | observed pre-conversion |
| glide **Java**                      | `Map<String, String[][]>`                         | `StreamBaseCommands:453` |
| glide **Python**                    | `Mapping[bytes, List[List[bytes]]]`               | `core.py:3445` |
| glide **Go**                        | `[]StreamEntry{ID, Fields []FieldValue}`          | `response_types.go:303` |
| redis-rb 5.4.1                      | `{"name"=>"alice","age"=>"30"}` Hash              | live run |
| **glide-ruby (default)**            | `["name","alice","age","30"]` flat                | live run — **sole outlier** |
| glide-ruby `flatten_map: true`      | `[["name","alice"],["age","30"]]` pairs           | live run — still not a Hash |

## Expected behavior

`xrange(...)` returns `[[entry_id, {field => value, ...}], ...]`, matching
redis-rb and the pair-of-pairs shape glide-core already assembled.

## Actual behavior

`xrange(...)` returns `[[entry_id, [field, value, field, value, ...]], ...]`.
By-name access raises; pair-iteration works but silently mishandles nils.

## Root cause

`lib/valkey/utils.rb:104–145`:

```ruby
HashifyStreamEntries = lambda { |reply|
  return [] if reply.nil?

  # In cluster mode, MAP responses come as Hash: {id => [fields], ...}
  if reply.is_a?(Hash)
    return reply.map { |entry_id, values|
      [entry_id, values.is_a?(Array) ? values.flatten : []]   # <-- discards structure
    }
  end
  ...
```

The `.flatten` on line 109 (and the block at line 133 that already receives a
flat array) collapses the pair-of-pairs into a flat array. Comment says
"Match redis-rb: return flat arrays" — but redis-rb does *not* return flat
arrays; it returns a Hash. Verified with `redis-rb 5.4.1` side by side.

## Suggested fix

```ruby
HashifyStreamEntries = lambda { |reply|
  return [] if reply.nil?

  entries =
    case reply
    when Hash
      reply.map { |id, pairs| [id, pairs] }
    when Array
      # Handle both [[id, pairs], ...] and [id, pairs, id, pairs, ...] shapes
      if reply.first.is_a?(Array) && reply.first.length == 2
        reply.compact
      else
        reply.compact.each_slice(2).to_a
      end
    else
      []
    end

  entries.map do |id, pairs|
    hash = {}
    pairs = pairs.to_a
    if pairs.first.is_a?(Array)
      pairs.each { |k, v| hash[k] = v }        # already [[k,v],[k,v]]
    else
      pairs.each_slice(2) { |k, v| hash[k] = v } # flat [k,v,k,v]
    end
    [id, hash]
  end
}
```

Also fix the sibling `HashifyStreamAutoclaim` in the same file: extend its
`reply[1].is_a?(Array)` guard to also accept `Hash`, and reshape via the same
helper.

Regression tests belong in `test/valkey/`:

- `xrange` / `xrevrange` / `xread` return `Hash`-valued entries.
- `xautoclaim` returns claimed entries (not empty).
- `flatten_map: true` — whatever contract we keep for it — is documented and
  asserted explicitly.

## Related

- Same conversion family as R3-17 (`xpending` argv order) — related but
  independent root causes; file R3-17 separately.
- Adjacent finding: `xautoclaim` silently drops entries. Cover here.

## Severity

`high` — common reach (every stream read), silent-nils on pair-iteration
plus loud `TypeError` on by-name access, no in-place workaround short of a
manual `.each_slice(2).to_h` at every call site.

jamesx-improving:
### Summary

Every stream-read command (`xrange`, `xrevrange`, `xread`, `xreadgroup`, `xclaim`) returned each entry's field-value pairs as a flat `Array` (`["name", "alice", "age", "30"]`) instead of the `Hash` (`{"name"=>"alice","age"=>"30"}`) that redis-rb 5.4.1 returns. By-name access (`entry[1]["name"]`) raised `TypeError`, and pair-iteration silently produced `nil` on odd-length data. The same root cause silently dropped every claimed entry from `xautoclaim` when the server had already transferred ownership.

### Issue link

Closes #240

### Features / Changes

- `lib/valkey/utils.rb` — `HashifyStreamEntries` rewritten to accept every input shape the FFI/RESP layer can produce (`Hash{id => pairs}`, `[[id, pairs], ...]`, or flat `[id, pairs, id, pairs, ...]`; and inner `pairs` as `Hash{k=>v}`, `[[k,v], [k,v]]`, or flat `[k,v,k,v]`) and normalize to redis-rb's `[[id, {field => value, ...}], ...]`. Inner conversion delegates to the existing `Utils::Hashify` helper. Duplicate field names in one entry collapse to last-write-wins, same as redis-rb.
- `lib/valkey/utils.rb` — `HashifyStreamAutoclaim` no longer discards a `Hash`-shaped `entries` slot (cluster mode / RESP3 path); it delegates to `HashifyStreamEntries`, which handles all shapes uniformly.
- `lib/valkey/commands/stream_commands.rb` — YARD docstrings updated to reflect the Hash return shape.
- `test/lint/stream_commands.rb` — every `assert_kind_of Array, entries[0][1]` flipped to `Hash` and augmented with positive by-name-access assertions (`entry[1]["field1"] == "value1"`). `xautoclaim` gains a `length >= 1` guard against the silent-drop regression.
- `test/valkey/utils_test.rb` — 12 new server-independent unit tests covering every input shape for both lambdas.

### Testing

Ran locally against a Valkey 8 standalone server on `127.0.0.1:6379`:

- `bundle exec rubocop` — clean (99 files)
- `bundle exec rake test:standalone TESTOPTS="--name=/hashify_stream/"` — 12 tests / 16 assertions / 0 failures
- `bundle exec rake test:standalone TESTOPTS="--name=/test_x/"` — 24 tests / 97 assertions / 0 failures
- Reporter's exact repro from the issue now returns `[["...", {"name"=>"alice", "age"=>"30"}], ...]` and `entries.first[1]["name"] == "alice"` — parity with redis-rb 5.4.1 confirmed live.

Cluster suite (`bundle exec rake test:cluster`) was not run locally — the same shared lint suite runs there, and no code path is cluster-specific to this change.

### Checklist

-   [x] Destination branch is correct - `release-1.0` (RC3 bug bash target).

### DIFF
--- a/lib/valkey/utils.rb
+++ b/lib/valkey/utils.rb
@@ -101,61 +101,29 @@ class Valkey
     EMPTY_STREAM_RESPONSE = [nil].freeze
     private_constant :EMPTY_STREAM_RESPONSE
 
+    # Normalizes the many shapes that stream-read replies can arrive in
+    # (Hash / [[id, pairs], ...] / flat [id, pairs, id, pairs] outer;
+    # Hash / [[k, v], ...] / flat [k, v, k, v] inner) into redis-rb's
+    # [[id, {field => value, ...}], ...] shape. Duplicate field names in one
+    # entry collapse to last-write-wins, same as redis-rb.
     HashifyStreamEntries = lambda { |reply|
-      return [] if reply.nil?
-
-      # In cluster mode, MAP responses come as Hash: {id => [fields], ...}
-      if reply.is_a?(Hash)
-        return reply.map { |entry_id, values| [entry_id, values.is_a?(Array) ? values.flatten : []] }
-      end
-
-      return [] if !reply.is_a?(Array) || reply.empty?
-
-      # Reply format: [[entry_id, [field1, value1, field2, value2, ...]], ...]
-      # Match redis-rb: return flat arrays [["id", ["field", "value", ...]], ...]
-      # Check if first element is a pair [entry_id, values_array]
-      first_elem = reply.first
-      if first_elem.is_a?(Array) && first_elem.length == 2
-        # Already in pair format: [[entry_id, [fields...]], ...]
-        reply.compact.map do |entry_id, values|
-          # Return flat array format like redis-rb, not hash
-          values_array = if values.nil?
-                           []
-                         elsif values.is_a?(Array)
-                           values
-                         else
-                           []
-                         end
-          [entry_id, values_array]
+      entries =
+        case reply
+        when Hash
+          reply
+        when Array
+          reply.first.is_a?(Array) ? reply : reply.each_slice(2)
+        else
+          return []
         end
-      else
-        # Flat array format: [entry_id1, [field1, value1, ...], entry_id2, [field2, value2, ...], ...]
-        reply.compact.each_slice(2).map do |entry_id, values|
-          # Return flat array format like redis-rb, not hash
-          values_array = if values.nil?
-                           []
-                         elsif values.is_a?(Array)
-                           values
-                         else
-                           []
-                         end
-          [entry_id, values_array]
-        end
-      end
+
+      entries.map { |entry_id, pairs| [entry_id, Hashify.call(pairs || [])] }
     }
 
     HashifyStreamAutoclaim = lambda { |reply|
       {
         'next' => reply[0],
-        'entries' => if reply[1].nil?
-                       []
-                     elsif reply[1].is_a?(Array)
-                       # Reply[1] is already an array of entries: [[id, [field, value, ...]], ...]
-                       # Use HashifyStreamEntries to convert them properly
-                       HashifyStreamEntries.call(reply[1])
-                     else
-                       []
-                     end
+        'entries' => HashifyStreamEntries.call(reply[1])
       }
     }
 
--- a/lib/valkey/commands/stream_commands.rb
+++ b/lib/valkey/commands/stream_commands.rb
@@ -102,7 +102,7 @@ class Valkey
       #
       # @example Read from a single stream
       #   valkey.xread(["mystream"], ["0"])
-      #     # => { "mystream" => [["1234567890-0", ["field1", "value1"]]] }
+      #     # => { "mystream" => [["1234567890-0", {"field1" => "value1"}]] }
       # @example Read with count and block
       #   valkey.xread(["mystream"], ["0"], count: 10, block: 1000)
       #
@@ -177,10 +177,11 @@ class Valkey
       # @param [String] end_id end ID ("-" for beginning, "+" for end)
       # @param [Hash] options optional parameters
       #   - `:count => Integer`: maximum number of entries to return
-      # @return [Array] array of [id, [field, value, ...]] entries
+      # @return [Array<Array>] array of `[id, {field => value, ...}]` entries
       #
       # @example Get all entries
       #   valkey.xrange("mystream", "-", "+")
+      #     # => [["1234567890-0", {"field1" => "value1", "field2" => "value2"}]]
       # @example Get entries with count limit
       #   valkey.xrange("mystream", "-", "+", count: 10)
       #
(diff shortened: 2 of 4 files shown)


lib/valkey/utils.rb
```diff
@@ -101,61 +101,29 @@ module Utils
     EMPTY_STREAM_RESPONSE = [nil].freeze
     private_constant :EMPTY_STREAM_RESPONSE
 
+    # Normalizes the many shapes that stream-read replies can arrive in
+    # (Hash / [[id, pairs], ...] / flat [id, pairs, id, pairs] outer;
+    # Hash / [[k, v], ...] / flat [k, v, k, v] inner) into redis-rb's
+    # [[id, {field => value, ...}], ...] shape. Duplicate field names in one
```
affonsov, line 107: Here is mentioning about duplicate keys, but I didn't see any test covering this. If possible can we add that?

jamesx-improving: Good catch — added `test_hashify_stream_entries_duplicate_field_names_last_write_wins` in `test/valkey/utils_test.rb`, covering both the flat-inner and pair-of-pairs-inner shapes so the last-write-wins contract is locked in for standalone and cluster/RESP3 paths. Addressed in 6a0b563.

yipin-chen (approved):

Aryex (approved):

affonsov (approved):

PR #246, opened 2026-08-07
feat: Add client_info_tag and lib_name config options

### LINKED ISSUE #243: [Task] Add client_info_tag (and lib_name) to client config
### Description

Follow-up from [valkey-io/valkey-glide#6429](https://github.com/valkey-io/valkey-glide/issues/6429) (which itself follows up on [#6389](https://github.com/valkey-io/valkey-glide/pull/6389), the Python reference implementation).

Add a `client_info_tag` config option to the Ruby client that appends a parenthesized tag to the default library name (e.g. `GlideRuby(SpringDataValkey)`), keeping the `GlideRuby` token visible in `CLIENT INFO`. This is preferred over a full `lib_name` override for framework attribution because it preserves GLIDE adoption visibility in dashboards while still allowing a per-framework breakdown.

While here, also expose `lib_name` as a full override (parity with the other clients), since the FFI connection request already supports it.

### Context (Ruby)

- The FFI layer already accepts a `lib_name` field in the connection-request JSON (`valkey-glide/ffi/src/lib.rs`), and the current default resolves to `GlideRuby`.
- Ruby currently exposes **neither** `lib_name` nor `client_info_tag` — connection options are assembled in `Valkey#build_connection_request` (`lib/valkey.rb`, around the `client_name` handling near line 153).

### Checklist

- [ ] Add `lib_name` option: pass through to `json_options["lib_name"]` when provided
- [ ] Add `client_info_tag` option: compose into the resolved lib name as `<lib_name>(<tag>)`, defaulting the base to `GlideRuby` when no `lib_name` override is given
- [ ] Validate that `client_info_tag` contains no whitespace (matches the `CLIENT SETINFO` value constraint); raise `ArgumentError` otherwise
- [ ] Tests: tag-only (`GlideRuby(<tag>)`), tag combined with `lib_name` override (`<lib_name>(<tag>)`), and whitespace-rejection — for both standalone and cluster
- [ ] Update README / DEVELOPER docs with the new connection options

### Additional Notes

- Python implementation in [#6389](https://github.com/valkey-io/valkey-glide/pull/6389) is the reference pattern.
- `client_info_tag` must not contain whitespace.
- Once supported, a Ruby framework integration could set `client_info_tag: "SpringDataValkey"` to produce `GlideRuby(SpringDataValkey)` in `CLIENT INFO`.

Jonathan-Improving:
## Summary

Adds two client-identification options to the Ruby client configuration:

1. **`lib_name`** — overrides the `CLIENT SETINFO LIB-NAME` value. When unset, the existing default (`GlideRuby`) is preserved.
2. **`client_info_tag`** — appends a parenthesized tag to the resolved library name while preserving the GLIDE identity, e.g. `GlideRuby(my-framework:1.2.3)`.

Closes #243.

## Motivation

Libraries and frameworks embedding GLIDE need to identify themselves for adoption tracking, observability, and debugging. `client_info_tag` keeps the `GlideRuby` token intact for GLIDE adoption dashboards while adding per-framework attribution.

## Behavior matrix

| Config | Resulting `lib-name` |
|---|---|
| _(none)_ | `GlideRuby` |
| `client_info_tag: \"lmcache:1.2\"` | `GlideRuby(lmcache:1.2)` |
| `lib_name: \"custom\"` | `custom` |
| `lib_name: \"custom\"`, `client_info_tag: \"lmcache:1.2\"` | `custom(lmcache:1.2)` |

## Usage

```ruby
# Tag only (recommended for frameworks)
client = Valkey.new(host: \"127.0.0.1\", client_info_tag: \"my-framework:1.0\")

# Full override
client = Valkey.new(host: \"127.0.0.1\", lib_name: \"CustomLib\")

# Both
client = Valkey.new(host: \"127.0.0.1\", lib_name: \"MyLib\", client_info_tag: \"v2.0\")
```

## What changed

- **`lib/valkey.rb`** — Added `lib_name` and `client_info_tag` handling in the options serialization path (after `client_name`). Validates tag has no whitespace. Composes `{base}({tag})` and passes to `json_options[\"lib_name\"]` which maps to protobuf `ConnectionRequest.lib_name` (field 19).
- **`test/valkey/client_info_tag_test.rb`** — New test module with validation tests (whitespace rejection) and integration tests (verifying CLIENT INFO output).
- **`test/standalone/valkey_test.rb`** — Include the new test module.

## Reference

Python implementation: valkey-io/valkey-glide#6389">
</invoke>


nderraugh:
1. 
**validation does not enforce printable ASCII** `valkey.rb 156 - 165` validates whitespace in client_info_tag; lib_name is not validated at all. The `/\s/` check also permits DEL, non-whitespace controls, and printable non-ASCII characters.
Examples currently accepted:
* lib_name: "invalid name"
* lib_name: "café"
* client_info_tag: "\x7F"
* client_info_tag: "框架"
* 
Both non-empty inputs should match printable ASCII from `!` through `~` (U+0021 through U+007E), inclusive before connection setup. Coverage at `client_info_tag_test.rb 15 - 42`
 needs lib_name validation for the same range.

2.
**empty values have incorrect semantics** At `valkey.rb 156 - 162`:
* `client_info_tag: ""` raises, but the requirement says empty means “not configured.”
* `lib_name: "", client_info_tag: "tag"` produces (tag) because empty strings are truthy in Ruby. It should produce `GlideRuby(tag)`.
* Empty lib_name without a tag only falls back accidentally because the FFI later converts it to None.
`client_info_tag_test.rb 15 - 20` currently codifies the empty-tag rejection.

3.
**metadata tests are not gated to server 7.2+** The server-visible assertions at 
`client_info_tag_test.rb 45 - 105` run against Redis 6.2 even though CLIENT SETINFO metadata is only available from 7.2.

This causes all five new metadata tests to fail in the five Redis 6.2 CI jobs. The logs confirm `CLIENT INFO` succeeds but contains no lib-name field. These tests should call the existing `omit_version("7.2")` helper before asserting metadata.

4.
(optional) **no binding-local pure resolver** Composition is embedded directly in `Valkey#initialize` at `valkey.rb 156 - 165`. The other clients (Java, Node, Python, Go) use a pure resolver over raw override and tag values so the full behavior matrix can be tested without a server or FFI.

Without options, Ruby also sends no runtime lib_name and depends on the native artifact being compiled with `GLIDE_NAME=GlideRuby` `(Rakefile:28-35)`. An analogous Ruby resolver would always return and propagate `GlideRuby`, while native compile-time naming remains a fallback.

5.
(optional) **cluster and recreated physical connections are unverified** `ClientInfoTag` is included only in `valkey_test.rb 15`; it is absent from `cluster_commands_test.rb`. Every integration case also explicitly skips cluster mode.

The shared constructor suggests initial cluster propagation should work, but the required behavior is not verified for:
* Connections to each cluster node
* Connections created after topology changes
* Recreated/reconnected physical connections

6.
**changelog** Does not seem to match this PR.

.gitignore
```diff
@@ -74,3 +74,6 @@ Gemfile.lock
 *.pem
 *.crt
 *.key
+
+# SDD docs
+sdd/
```
Aryex: What is this?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 * Ruby: Fix `blpop`, `brpop`, `blmove`, `rpoplpush` and `brpoplpush`, all of which were non-functional. `blpop`/`brpop` called a non-existent `send_blocking_command` helper, `blmove` leaked the command name into argv, and `rpoplpush`/`brpoplpush` dispatched `RequestType::RPOPLPUSH`/`BRPOPLPUSH`, for which glide-core has no command mapping. Since Valkey defines `RPOPLPUSH src dst` as exactly `LMOVE src dst RIGHT LEFT` (and `BRPOPLPUSH src dst timeout` as `BLMOVE src dst RIGHT LEFT timeout`), `rpoplpush`/`brpoplpush` are now fixed-argument facades over `lmove`/`blmove`. Both remain deprecated as of Redis 6.2; prefer `lmove`/`blmove` in new code. The unusable `RequestType::RPOPLPUSH`/`BRPOPLPUSH` constants were removed.
 
+### Breaking Changes
+
+* Ruby: `client` dispatcher uses `public_send` instead of `send` — subclasses that previously relied on dispatching to private `client_*` helpers via the `client(:subcommand)` interface must make those methods public. This prevents accidental invocation of unrelated private methods and aligns with Ruby best practices. ([PR #1](https://github.com/Jonathan-Improving/valkey-glide-ruby/pull/1))
+
```
Aryex, line 8: You probably need to rebase against main. We recently back ported 1.0 release fixes so the commit history different.

test/test_helper.rb
```diff
@@ -8,6 +8,7 @@
 require_relative "support/test_cluster"
 
 require "minitest/autorun"
+require "minitest/mock"
```
Aryex, line 11: Is this used anywhere here?

lib/valkey.rb
```diff
@@ -152,6 +152,18 @@ def initialize(options = {})
...
+    if options[:client_info_tag]
+      tag = options[:client_info_tag].to_s
+      raise ArgumentError, "client_info_tag must not be empty" if tag.empty?
+      raise ArgumentError, "client_info_tag must not contain whitespace, got: #{tag.inspect}" if tag.match?(/\s/)
+
+      base = options[:lib_name] ? options[:lib_name].to_s : "GlideRuby"
+      json_options["lib_name"] = "#{base}(#{tag})"
+    elsif options[:lib_name]
+      json_options["lib_name"] = options[:lib_name].to_s
+    end
```
Aryex: While `lib_name` and `client_info_tag` are validated server side, it seems GLIDE core never surfaces error to the users. We should fix this on the core's side. 

```ruby
# Ok, but the lib_name is invalid.
client = Valkey.new(host: "127.0.0.1", port: 6379, lib_name: "café")

# Throws Valkey::CommandError
client.call("CLIENT","SETINFO","lib-name","café")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    def test_client_info_tag_rejects_whitespace_newline
+      error = assert_raises(ArgumentError) do
+        Valkey.new(host: "127.0.0.1", port: PORT, timeout: TIMEOUT, client_info_tag: "has\nnewline")
+      end
+      assert_match(/client_info_tag must not contain whitespace/, error.message)
+    end
+
+    # --- Integration tests (server needed) ---
```
Aryex: These integration tests should also be gated on server versions as well, hence the CI failures.

Aryex:
Hi @Jonathan-Improving. Apologies for not reviewing this sooner. Are you still working on this?

Jonathan-Improving:
@Aryex yes, still working on this, standby...

.gitignore
Jonathan-Improving: Resolved. That was the `# SDD docs` / `sdd/` entry — a stray local spec-driven-dev scratch dir that should never have been in `.gitignore`. Removed in cbf7172; the `sdd/` directory is not tracked and the ignore entry is gone.

CHANGELOG.md
Jonathan-Improving: Fixed in 0fe8721. The breaking-change note pointed at a fork-local `PR #1` link — it now references upstream `#246`. I also added the missing behavior note: because the `client` dispatcher moved from `send` to `public_send`, intentionally-private `client_*` helpers (e.g. the internal client-side-caching commands removed from the public interface) are no longer reachable via `client(:subcommand)` and now raise `NoMethodError`. The PR is currently MERGEABLE, so a full rebase against main is not strictly required; happy to rebase if you would prefer the history aligned.

test/test_helper.rb
Jonathan-Improving: Yes — it is used. `require "minitest/mock"` provides `Object#stub`, which the suite relies on extensively, e.g. `test/valkey/call_test.rb` (`r.stub(:send_command, ...)`), `test/valkey/connection_config_test.rb`, `test/valkey/connection_lifecycle_test.rb`, `test/valkey/failover_commands_test.rb`, and the lint recorders in `test/lint/*.rb`. Without the require those stubs raise `NoMethodError`. Leaving as-is.

lib/valkey.rb
Jonathan-Improving: Agreed the proper long-term fix is core surfacing the error, but the shared-core gitlink is pinned for this PR, so I mitigated on the Ruby side in 0fe8721: both `lib_name` and `client_info_tag` are now validated against printable ASCII (U+0021 through U+007E) before the connection is opened, raising `ArgumentError` for values like `"café"`, `"框架"`, `"\x7F"`, or embedded spaces — instead of silently connecting with a value the server rejects. Added coverage in `client_info_tag_test.rb` (`test_lib_name_rejects_non_ascii`, `test_client_info_tag_rejects_control_character`, etc.).

test/integration/valkey/client_info_tag_test.rb
Jonathan-Improving: Resolved (cbf7172). Every server-visible integration test in this file now calls `omit_version("7.2")` before asserting metadata, so the five Redis 6.2 CI jobs skip them (CLIENT SETINFO lib-name is 7.2+). Cluster mode is likewise skipped via `cluster_mode?`. Verified green locally against a 7.2+ server: 13 tests, 34 assertions, 0 failures.

lib/valkey.rb
Jonathan-Improving: I'm going to see about getting at the proper fix in core and revert the https://github.com/valkey-io/valkey-glide-ruby/commit/0fe87213045c0e26c3fdcaa6af7b993f5bd62e45 short-term fix that my agent applied.

**UPDATE:** Seems this may be more gnarly than I had previously thought, worth a quick chat during stand-up.

CHANGELOG.md
Jonathan-Improving: Done — rebased onto `origin/main` (head now `06b16f6`), which picks up the back-ported 1.0 release fixes. Resolved the resulting conflicts: the test suite move to `test/integration/valkey/` (this feature's test file moved with it), plus README/CHANGELOG/lib overlaps. PR is `MERGEABLE` again.

lib/valkey.rb
Jonathan-Improving: Reverted the client-side printable-ASCII validation as discussed — the invalid-value rejection is now handled in glide-core, so this wrapper no longer duplicates it. Removed the `validate_client_metadata!` helper, its call sites, the ASCII/control-char tests, and the ASCII wording in the README. The pre-existing empty/whitespace `client_info_tag` checks remain. Verified: `client_info_tag: "café"` / `lib_name: "café"` now pass validation (no `ArgumentError`), while empty and whitespace tags still raise. RuboCop clean; unit suite 121/0.

Jonathan-Improving: Resolved properly now that the core fix has landed. The submodule was bumped to glide-core `49f2b6d21`, which contains #6891 (`fix(core): Validate client library names`, merged as `0e02ff973`), so invalid library names fail client creation as a configuration error and surface in Ruby as `Valkey::CannotConnectError` — no panic, no silent pass.

Following on from that, this branch now defers **all** library-name validation to core: the wrapper performs no character checking of `lib_name` or `client_info_tag`. The previous empty/whitespace `client_info_tag` raises were removed — whitespace is rejected by core's grammar, and an empty tag is treated as absent (matching core, whose tests assert `Some("")` is `Ok(())`).

One subtlety worth noting: the empty-tag guard was also preventing the wrapper from composing `base()`, which core rejects (empty parentheses fail the grammar). So an empty tag now skips composition entirely rather than raising — a compositional rule, not validation. Verified against a local standalone server: `client_info_tag: ""` yields `lib-name=GlideRuby` (and `CustomLib` when combined with `lib_name`) with no stray `()` and no error.

Coverage added for both the core-rejection cases (whitespace, non-ASCII, DEL, and all malformed-paren forms) and the empty-tag composition cases.

Jonathan-Improving:
@nderraugh thanks — all six addressed. Head is now `7a109f7`.

**1. Printable ASCII not enforced** — resolved, but in core rather than the wrapper. The submodule was bumped to glide-core `49f2b6d21`, which contains #6891 (`fix(core): Validate client library names`). Core validates the composed name before client creation, so `lib_name: "invalid name"`, `"café"`, `"\x7F"` and `"框架"` now all fail with `Valkey::CannotConnectError`. Per maintainer direction the wrapper deliberately does *not* duplicate character validation — core is the single authority. Coverage added for whitespace, non-ASCII, DEL and every malformed-paren form, for both `lib_name` and `client_info_tag`.

**2. Empty values have incorrect semantics** — fixed; this was a real bug and a good catch. As you noted, an empty string is truthy in Ruby, so `lib_name: ""` looked configured and `lib_name: "", client_info_tag: "tag"` composed `(tag)`. With core validation now in place that was no longer merely wrong, it was fatal — core rejects a name starting with a parenthesis, so it raised `CannotConnectError`. Empty `lib_name` and empty `client_info_tag` are now both normalized to "not configured", so that case yields `GlideRuby(tag)`. The empty-tag rejection you flagged in the tests is gone; empty tag now means absent, matching core (whose tests assert `Some("")` is `Ok(())`).

**3. Metadata tests not gated to 7.2+** — resolved. Every server-visible assertion calls `omit_version("7.2")`.

**4. No binding-local pure resolver** — done. Composition now lives in `Valkey.resolve_lib_name`, a pure function over the raw override and tag, with server-free unit tests covering the full matrix including each empty-string form (`test/unit/lib_name_resolver_test.rb`). Also per your point, the resolved value is now *always* sent, so the reported name no longer depends on the native artifact being built with `GLIDE_NAME=GlideRuby`; compile-time naming remains a fallback in core.

**5. Cluster / recreated connections unverified** — done. `ClientInfoTag` is now included in the cluster suite and the blanket "standalone only" skips are removed; the tests use `_new_client`, so they run unchanged in both suites.

This surfaced a pre-existing harness bug worth flagging: a cluster `INFO` returns `{ "host:port" => { ... } }`, but `extract_version_from_info` only looked for the version keys at the top level, so `version` fell back to `"0.0"` and *every* `omit_version`-gated cluster test silently skipped. Fixed to fall back to the first node's info — cluster metadata coverage went from 9 skips to 0.

**6. Changelog** — resolved. Added an entry for the feature and a Fixes entry for the empty-value behavior.

Verified: RuboCop 106 files / no offenses; unit 138 tests / 266 assertions / 0 failures; standalone 198 tests / 0 failures; cluster `client_info_tag` set 23 tests / 63 assertions / 0 failures / 0 skips. The 4 remaining cluster TLS errors are environmental (`TLS_CERT_DIR` unset) and reproduce without these changes.

nderraugh (approved):

lib/valkey.rb
```diff
@@ -24,6 +24,110 @@ class Valkey
...
+  # "A(x)(y)", which core rejects outright (its grammar allows at most one
+  # trailing tag). That is left to core rather than pre-empted here, since
+  # inspecting the base's characters is exactly the validation this binding does
+  # not do.
+  #
+  # @param lib_name [String, Symbol, nil] raw `lib_name` override, may be empty
+  # @param client_info_tag [String, Symbol, nil] raw `client_info_tag`, may be empty
+  # @return [String] the effective library name
+  # @raise [ArgumentError] if either value is neither a String, a Symbol, nor nil
+  def self.resolve_lib_name(lib_name: nil, client_info_tag: nil)
```
Aryex, line 41: I assume `resolve_lib_name` is made public so we can test it? Could we mark it with `_resolve_lib_name` then and document that it for internal use only?

lib/valkey.rb
```diff
@@ -24,6 +24,110 @@ class Valkey
...
+  # and is the single authority on it. The only rules applied are compositional:
+  # an empty override or tag means "not configured" (matching core's
+  # empty-means-absent semantics), so no `base()` or `(tag)` is ever composed.
+  #
+  # Note the composition rule is only meaningful for an unparenthesized base: a
+  # `lib_name` that already ends in "(...)" combined with a tag composes
+  # "A(x)(y)", which core rejects outright (its grammar allows at most one
+  # trailing tag). That is left to core rather than pre-empted here, since
+  # inspecting the base's characters is exactly the validation this binding does
+  # not do.
```
Aryex: I'm seeing a lot of these chunky comments. Could we review them and simplify? Ideally code should be self-explanatory with comments to highlight subtle behaviours.

lib/valkey.rb
```diff
@@ -24,6 +24,110 @@ class Valkey
...
+      raise ArgumentError,
+            "#{field} must not contain an invalid #{string.encoding} byte sequence"
+    end
+
+    begin
+      string = string.encode(Encoding::UTF_8)
+    rescue Encoding::UndefinedConversionError, Encoding::InvalidByteSequenceError => e
+      raise ArgumentError,
+            "#{field} must be convertible to UTF-8 (#{e.class}: #{e.message})"
+    end
```
Aryex, line 78: Same as the [above](https://github.com/valkey-io/valkey-glide-ruby/pull/246/changes#r3916425310).

For example, I don't think this comment is necessary?  These kind of chunky comment makes things more complicated.

```suggestion
    # Reject malformed bytes before conversion
    unless string.valid_encoding?
      raise ArgumentError,
            "#{field} must not contain an invalid #{string.encoding} byte sequence"
    end

    begin
      string = string.encode(Encoding::UTF_8)
    rescue Encoding::UndefinedConversionError, Encoding::InvalidByteSequenceError => e
      raise ArgumentError,
            "#{field} must be convertible to UTF-8 (#{e.class}: #{e.message})"
    end
```

test/unit/version_parseable_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    "0garbage",
+    "7.1garbage",
+    "8garbage",     # end-unanchored regex gated this to RUN: a false PASS
+    "8.1.0.",
+    "8.1.0-",
+    "-1.0",
+    "v8.1.0",
+    " 8.1.0",
+    "8.1.0 "
+  ].freeze
```
Aryex, line 33: I think `8.1.0.`, `8.1.0`, `8.1.0-` are collapsed into `test_rejects_8_1_0_ `

lib/valkey.rb
```diff
@@ -24,6 +24,110 @@ class Valkey
...
+    # This screens BYTES, not characters: "café" satisfies both and is passed
+    # through untouched for glide-core to judge.
+    unless string.valid_encoding?
+      raise ArgumentError,
+            "#{field} must not contain an invalid #{string.encoding} byte sequence"
+    end
+
+    begin
+      string = string.encode(Encoding::UTF_8)
+    rescue Encoding::UndefinedConversionError, Encoding::InvalidByteSequenceError => e
```
Aryex: Should this be just catch all instead?

```suggestion
    rescue EncodingError => e
```

test/unit/version_parseable_test.rb
Jonathan-Improving: Good catch, and it's worse than described — there are **three** colliding entries, not two, and the collision was silently deleting assertions.

`gsub(/\W/, '_')` maps `"8.1.0."`, `"8.1.0-"` **and** `"8.1.0 "` (trailing space) all onto `test_rejects_8_1_0_`. `define_method` overwrites silently, so 2 of those 3 never ran:

```
REJECTED entries:        14
distinct method names:   12
test_rejects_8_1_0_  <-  ["8.1.0.", "8.1.0-", "8.1.0 "]  (3x)
```

The file reported `25 tests, 0 failures` while quietly discarding coverage — a suite claiming more than it holds.

The guard itself was never broken: `parseable?` correctly returns `false` for all three, so nothing shipped wrong. This was lost *coverage*, not a product bug. `ACCEPTED` was unaffected (8/8 distinct).

Fixed by indexing the generated names, so two entries can no longer collapse regardless of their punctuation. Test count `25 -> 28` (the 2 recovered assertions, plus one new guard).

I also added `test_every_table_entry_generates_a_distinct_test`, which asserts generated-method count equals table size — so if someone reintroduces value-derived naming or adds a duplicate entry, it fails loudly instead of shrinking the suite. Mutation-verified: dropping the index back out produces

```
REJECTED table has 14 entries but generated 3 tests.
1 failures
```

RuboCop clean; unit suite `183 tests, 336 assertions, 0 failures`.

lib/valkey.rb
Jonathan-Improving: Fair point, and I measured it rather than guessing — you're right on volume:

```
lib/valkey.rb lines 31-131:  65 comment / 28 code  =  2.32:1
```

Trimmed to **1.0:1** in that region (−51 comment lines, +13). Cut the "parallels the Java/Node/Go resolvers" paragraph, the empty-means-absent restatement, and the `to_s`-coercion history — all either restating the code or better placed in the PR description than in the source.

**One block I deliberately kept, in reduced form** (2 lines instead of 20), and I want to flag it rather than trim it silently: the note on why *both* encoding checks are needed.

```ruby
# Both checks are required: #valid_encoding? catches a String tagged UTF-8
# holding invalid bytes (where #encode is a no-op); #encode catches other
# encodings not representable as UTF-8 (where valid_encoding? is vacuously
# true). Either alone leaves a JSON::GeneratorError vector.
```

That one earned its place: during QA an earlier round proposed replacing the pair with `#encode` alone, which **would have reopened the invalid-UTF-8 vector** — `#encode` is a no-op when source and target encodings match, so it cannot detect a String already tagged UTF-8 that holds invalid bytes. The comment exists specifically to stop that simplification being made again. Both guards are mutation-verified (neutering either fails tests).

Happy to cut it further if you'd rather it live in a test name or the PR body instead.

lib/valkey.rb
Jonathan-Improving: Applied — this block is now essentially your suggestion. The 20-line explanation is down to 4 lines covering only the non-obvious part (why *both* checks are needed, not what each one does).

Detail in the [thread above](https://github.com/valkey-io/valkey-glide-ruby/pull/246#discussion_r3916425310): the pair is load-bearing, so I kept a short note guarding against re-simplifying it to `#encode` alone. Everything else in the block went.

Also took your `rescue EncodingError` suggestion from the [other thread](https://github.com/valkey-io/valkey-glide-ruby/pull/246#discussion_r3917502652), so the `begin/rescue` here is now narrower than what you quoted.

lib/valkey.rb
Jonathan-Improving: Agreed, applied.

I checked what widening actually admits before taking it — `EncodingError` has exactly four descendants:

```
Encoding::CompatibilityError
Encoding::ConverterNotFoundError
Encoding::InvalidByteSequenceError    <- was caught
Encoding::UndefinedConversionError    <- was caught
```

So it adds `CompatibilityError` and `ConverterNotFoundError`. Neither is reachable from this call site: `ConverterNotFoundError` requires a bogus **explicit source** encoding (`"abc".encode("UTF-8", "NO-SUCH-ENCODING")`), and this call only ever passes a target of `Encoding::UTF_8`. `CompatibilityError` comes from concatenation-style operations, not `#encode`.

So the catch-all is strictly more robust with no behaviour change today, and it won't silently miss a new `EncodingError` subclass in a future Ruby. Both encoding guards remain mutation-verified after the change.

lib/valkey.rb
Jonathan-Improving: @Aryex this one is resolved upstream — summarising once more, as briefly as I can, since the thread has a lot of history.

**Your original point was right, and it was fixed in core, not here.** The proper fix you asked for landed as valkey-io/valkey-glide#6891 (`fix(core): Validate client library names`, merged as `0e02ff973`). This branch's submodule is pinned at `49f2b6d21`, which contains it.

Current behaviour for your exact example:

```ruby
Valkey.new(host: "127.0.0.1", port: 6379, lib_name: "café")
# => Valkey::CannotConnectError: configuration error: library name must contain
#    only printable ASCII characters from '!' through '~' and parentheses must
#    form a non-empty trailing '(tag)'
```

Fails at **client creation** — no silent connect with a value the server rejects, no panic. Verified against a live standalone server, not just unit tests.

The interim client-side ASCII validation I added in `0fe8721` has been **reverted**, so the wrapper no longer duplicates core's rules; it does type/encoding preconditions and name composition only.

Nothing further is pending on this thread — **could you resolve it when you get a chance?** It's currently counted as blocking review feedback on the PR.

lib/valkey.rb
Jonathan-Improving: Correct on the diagnosis — it's public for testability. I checked every reference: the only production caller is `#initialize`, and everything else pointing at it is `test/unit/lib_name_resolver_test.rb`.

I've marked it rather than renamed it:

```ruby
# @api private
def self.resolve_lib_name(lib_name: nil, client_info_tag: nil)
```

Reasoning for preferring that over `_resolve_lib_name` — happy to be overruled:

- A leading underscore isn't a Ruby convention for "internal"; it's borrowed from Python. Ruby's vocabulary for this is `@api private` (YARD, which this file already uses throughout) or `:nodoc:`.
- `@api private` is machine-readable — YARD excludes it from generated docs — whereas an underscore only communicates to a human who already knows the convention.
- It documents intent without renaming a method that has a test suite pointing at it, so no churn in the test file for a naming preference.

`coerce_client_metadata` next to it is genuinely `private_class_method`, which is the stronger form. `resolve_lib_name` can't be, since the tests need it — `@api private` is the accurate signal for "public only because it's tested".

If you'd still rather it be renamed, I'll go with `internal_resolve_lib_name` — just say so.

**One thing worth recording while this file is under review:** the whitespace-passthrough behaviour pinned in `lib_name_resolver_test.rb` (`resolve_lib_name(client_info_tag: " ") == "GlideRuby( )"`) is now an **explicit policy decision**, not an oversight. Client-side filtering is deliberately not performed: the server is the authority on library-name validity, and its rejection is the intended feedback for what is a caller mistake. Core's grammar excludes `0x20` (its ranges start at `\x21`), so that value fails validation and surfaces as `Valkey::CannotConnectError` at client creation — verified against a live server. I've expanded the comment on that test to say so, so it doesn't get "fixed" later by normalising the whitespace away.

Note empty and whitespace are intentionally different: empty means *absent* (core returns `Ok` for it, so no `()` is composed), whitespace is *invalid input*.

Jonathan-Improving: **TL;DR**: Going with the "API private" annotation rather than the rename, unless you insist otherwise @Aryex

test/lint/connection_options.rb
```diff
@@ -99,6 +99,71 @@ def test_connection_with_client_name
       client.close
     end
 
+    # --- Library identification options (F-08) ---
```
Aryex: What is F-08?

lib/valkey.rb
```diff
@@ -24,6 +24,84 @@ class Valkey
   include Commands
   include PubSubCallback
 
+  # Default value reported for `CLIENT SETINFO LIB-NAME`.
+  #
+  # The native artifact is normally compiled with `GLIDE_NAME=GlideRuby`, and
+  # glide-core falls back to that compile-time value when no runtime library name
+  # is sent. Resolving the same default here means the runtime value is always
+  # sent explicitly, so the reported name does not depend on how the native
+  # artifact happened to be built, while compile-time naming remains a fallback.
```
Aryex: The compile-time-artifact paragraph is design rationale, not API documentation.

```suggestion
  # Default value reported for `CLIENT SETINFO LIB-NAME`.
```

lib/valkey.rb
```diff
@@ -24,6 +24,84 @@ class Valkey
...
+  # authority on library-name validity and its error is the intended feedback.
+  #
+  # A `lib_name` already ending in "(...)" plus a tag composes "A(x)(y)", which
+  # core rejects — left to core, since inspecting the base is validation.
+  #
+  # @api private
+  # @param lib_name [String, Symbol, nil] raw `lib_name` override, may be empty
+  # @param client_info_tag [String, Symbol, nil] raw `client_info_tag`, may be empty
+  # @return [String] the effective library name
+  # @raise [ArgumentError] if either value is neither a String, a Symbol, nor nil
```
Aryex: Trim to the composition rule; the `@raise` also misses the invalid-encoding case.

```suggestion
  # Resolves the effective `CLIENT SETINFO LIB-NAME` value, composing `base(tag)`.
  # An empty override or tag means "not configured" and is omitted. Character
  # validity is glide-core's (see valkey-io/valkey-glide#6891).
  #
  # @api private
  # @param lib_name [String, Symbol, nil] raw `lib_name` override, may be empty
  # @param client_info_tag [String, Symbol, nil] raw `client_info_tag`, may be empty
  # @return [String] the effective library name
  # @raise [ArgumentError] if either value is neither a String, a Symbol, nor nil,
  #   or is a String with an invalid byte sequence
```

lib/valkey.rb
```diff
@@ -24,6 +24,84 @@ class Valkey
...
+  # Blanket `to_s` was harmful: `lib_name: false` silently identified the client
+  # as "false", and arbitrary objects leaked inspect output or a
+  # JSON::GeneratorError, neither of which is a Valkey error, breaking the
+  # documented `rescue Valkey::BaseError` contract around construction.
+  #
+  # @param field [String] option name, used in the error message
+  # @param value [String, Symbol, nil] the raw configured value
+  # @return [String, nil] the value, or nil when absent/empty
+  # @raise [ArgumentError] if the value is not a String/Symbol/nil, or is a
+  #   String with an invalid byte sequence
```
Aryex: The `to_s` history paragraph belongs in the PR description, not the docstring.

```suggestion
  # Normalizes a client-identification value to a non-empty String or nil.
  #
  # @param field [String] option name, used in the error message
  # @param value [String, Symbol, nil] the raw configured value
  # @return [String, nil] the value, or nil when absent/empty
  # @raise [ArgumentError] unless the value is a String, Symbol, or nil with a
  #   valid byte sequence
```

lib/valkey.rb
```diff
@@ -24,6 +24,84 @@ class Valkey
...
+      raise ArgumentError,
+            "#{field} must be a String, Symbol, or nil, got #{value.class}"
+    end
+
+    string = value.to_s
+    # Both checks are required: #valid_encoding? catches a String tagged UTF-8
+    # holding invalid bytes (where #encode is a no-op); #encode catches other
+    # encodings not representable as UTF-8 (where valid_encoding? is vacuously
+    # true). Either alone leaves a JSON::GeneratorError vector. Screens BYTES,
+    # not characters — "café" passes through for core to judge.
```
Aryex: Worth keeping — it is a real Ruby gotcha — but two lines do it, and `valid_encoding?` is only vacuous for ASCII-8BIT, not "other encodings" generally.

```suggestion
    # Both checks are needed: #valid_encoding? catches invalid bytes in a String
    # already tagged UTF-8; #encode catches ASCII-8BIT, where every byte is valid.
```

lib/valkey.rb
```diff
@@ -152,6 +230,21 @@ def initialize(options = {})
...
+    #
+    # Composition/normalization lives in the pure `Valkey.resolve_lib_name`
+    # resolver so it is testable without a server or the FFI. Character validity
+    # is glide-core's responsibility (see valkey-io/valkey-glide#6891), so no
+    # character checking is performed here.
+    #
+    # The resolved value is always sent, so the reported library name does not
+    # depend on the native artifact having been compiled with
+    # `GLIDE_NAME=GlideRuby`.
+    json_options["lib_name"] = self.class.resolve_lib_name(
```
Aryex: Duplicates the `DEFAULT_LIB_NAME` and `resolve_lib_name` docstrings.

```suggestion
    json_options["lib_name"] = self.class.resolve_lib_name(
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+# These options control the CLIENT SETINFO LIB-NAME value sent during
+# connection establishment. client_info_tag appends a parenthesized tag
+# to the default library name (GlideRuby), while lib_name provides a
+# full override.
+#
+# References:
+# - valkey-io/valkey-glide#6389 (Python reference implementation)
+# - valkey-io/valkey-glide#6891 (core-side library-name validation, which this
+#   file's rejection expectations depend on)
+module ValkeyTests
```
Aryex: File header restates what the options do; the tests and README already cover it.

```suggestion
module ValkeyTests
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+  module ClientInfoTag
+    # Asserts the EXACT lib-name reported by CLIENT INFO.
+    #
+    # CLIENT INFO fields are space-delimited, so an unanchored
+    # `assert_match(/lib-name=CustomLib/)` also passes for "CustomLibSURPRISE" or
+    # "CustomLib(unexpected)" — i.e. it cannot fail for the wrong-composition bug
+    # these tests exist to catch. Anchoring on the delimiter fixes the whole class
+    # of that bug rather than one instance.
```
Aryex: Keep the one-line docstring plus why the regex anchors; drop the argument aimed at reviewers.

```suggestion
    # Asserts the EXACT lib-name reported by CLIENT INFO. The regex anchors on the
    # field delimiter so "CustomLibSURPRISE" / "CustomLib(unexpected)" do not pass.
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+
+    # --- Tag composition / empty handling (no character validation here) ---
+    #
+    # The wrapper performs no character validation of client_info_tag or
+    # lib_name; that is glide-core's responsibility (upstream #6891). The one
+    # wrapper-side rule is compositional: an empty tag is treated as absent,
+    # matching core's empty-means-absent semantics, so it must neither raise nor
+    # compose a "GlideRuby()" that core would reject.
+
+    def test_empty_client_info_tag_is_treated_as_absent
```
Aryex: Duplicates resolve_lib_name's docstring, and empty-means-absent is a wrapper rule, not core's — core rejects "GlideRuby()" (line 127).

```suggestion
    def test_empty_client_info_tag_is_treated_as_absent
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      begin
+        info = client.call("CLIENT", "INFO")
+        assert_lib_name("GlideRuby", info)
+      ensure
+        client&.close
+      end
+    end
+
+    def test_empty_client_info_tag_with_lib_name_is_treated_as_absent
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: This trailing comment repeats verbatim on ten omit_version lines; keep only the first (line 38).

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      ensure
+        client&.close
+      end
+    end
+
+    # Regression: an empty lib_name combined with a tag previously composed
+    # "(tag)" — an empty string is truthy in Ruby, so the override looked
+    # configured — which glide-core rejects. An empty override must be treated as
+    # absent so the default base is used.
+    def test_empty_lib_name_with_tag_uses_default_base
```
Aryex: Development history; the test name already states the expected behavior.

```suggestion
    def test_empty_lib_name_with_tag_uses_default_base
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Duplicate of the note kept at line 38.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+        info = client.call("CLIENT", "INFO")
+        assert_lib_name("GlideRuby(my-framework:1.0)", info)
+      ensure
+        client&.close
+      end
+    end
+
+    # Whitespace in a tag is rejected by core (it fails the library-name
+    # grammar), not by the wrapper.
+    CORE_REJECTED_TAGS = {
```
Aryex: CORE_REJECTED_TAGS and the literal "has space" values already say this.

```suggestion
    CORE_REJECTED_TAGS = {
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+        # Core validates at Client::new, so client CREATION raises — there is no
+        # command to send. Assert the documented Valkey::CannotConnectError
+        # (README) rather than the Valkey::BaseError root.
+        #
+        # Verified caveat: a connection failure ALSO raises CannotConnectError, so
+        # narrowing the class does not by itself distinguish "rejected the name"
+        # from "could not reach the server". The message assertion below is
+        # therefore load-bearing, not decorative — do not remove it. It does couple
+        # this test to an upstream message string; that is a deliberate trade for
+        # not having a false green.
```
Aryex: Keep only the fact that makes the message assertion necessary; drop the plea to reviewers.

```suggestion
        # A connection failure also raises CannotConnectError, so the message
        # assertion is what distinguishes rejection from an unreachable server.
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+    # validation; glide-core validates the composed lib-name before client
+    # creation and surfaces a configuration error through the FFI. These tests
+    # assert that the error reaches the caller as a Valkey error at client
+    # creation, rather than panicking or silently connecting with a name the
+    # server ignores.
+    #
+    # The accepted grammar is documented canonically in README's Connection
+    # Options (the `lib_name` row); it is deliberately not restated here, since
+    # independent copies of it have already drifted apart. The cases below are
+    # transcribed from core's own regex at the pinned commit.
```
Aryex: Keep provenance for the case list; drop the "deliberately do NOT" declaration and drift history.

```suggestion
    # Cases transcribed from glide-core's library-name regex (upstream #6891);
    # the accepted grammar is documented in README's Connection Options.
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      "trailing_suffix_after_tag" => "GlideRuby(tag)suffix"
+    }.freeze
+
+    CORE_REJECTED_LIB_NAMES.each do |label, value|
+      define_method(:"test_core_rejects_lib_name_#{label}") do
+        client = nil
+        # See the note on CORE_REJECTED_TAGS above: creation raises, the documented
+        # error class is asserted, and the message assertion is load-bearing
+        # because a connection failure shares that class.
+        error = assert_raises(Valkey::CannotConnectError) do
```
Aryex: Back-reference restates the note above and the assertions below.

```suggestion
        error = assert_raises(Valkey::CannotConnectError) do
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+          client = _new_client(lib_name: value)
+        end
+        assert_match(/library name must contain only printable ASCII/, error.message)
+      ensure
+        client&.close
+      end
+    end
+
+    def test_core_accepts_valid_composed_lib_name_and_tag
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Duplicate of the note kept at line 38.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+
+      client = _new_client(lib_name: "custom-lib", client_info_tag: "framework:1.2")
+      begin
+        # Assert the composed name actually reached the server — construction not
+        # raising says nothing about what was sent.
+        assert_lib_name("custom-lib(framework:1.2)", client.call("CLIENT", "INFO"))
```
Aryex: Restates the assertion on the following line.

```suggestion
        assert_lib_name("custom-lib(framework:1.2)", client.call("CLIENT", "INFO"))
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      ensure
+        client&.close
+      end
+    end
+
+    def test_empty_lib_name_falls_back_to_default_without_error
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Duplicate of the note kept at line 38.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+        info = client.call("CLIENT", "INFO")
+        assert_lib_name("GlideRuby", info)
+      ensure
+        client&.close
+      end
+    end
+
+    # --- Integration tests (server needed) ---
+
+    def test_client_info_tag_appends_to_default_lib_name
```
Aryex: Section divider, and wrong: the tests above (lines 42, 54, 70, 155, 166) also need a live server.

```suggestion
    def test_client_info_tag_appends_to_default_lib_name
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Duplicate of the note kept at line 38.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      begin
+        info = client.call("CLIENT", "INFO")
+        assert_lib_name("GlideRuby(my-framework:1.0)", info)
+      ensure
+        client&.close
+      end
+    end
+
+    def test_lib_name_override
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Duplicate of the note kept at line 38.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      begin
+        info = client.call("CLIENT", "INFO")
+        assert_lib_name("CustomLib", info)
+      ensure
+        client&.close
+      end
+    end
+
+    def test_lib_name_with_client_info_tag
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Duplicate of the note kept at line 38.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      begin
+        info = client.call("CLIENT", "INFO")
+        assert_lib_name("MyLib(v2.0)", info)
+      ensure
+        client&.close
+      end
+    end
+
+    def test_default_lib_name_when_no_options
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Duplicate of the note kept at line 38.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      begin
+        info = client.call("CLIENT", "INFO")
+        assert_lib_name("GlideRuby", info)
+      ensure
+        client&.close
+      end
+    end
+
+    def test_client_info_tag_with_special_characters
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Duplicate of the note kept at line 38.

```suggestion
      omit_version("7.2")
```

test/unit/lib_name_resolver_test.rb
```diff
@@ -0,0 +1,172 @@
...
+#
+# They are not, however, dependency-free: requiring test_helper loads `valkey`,
+# which dlopens the native library at require time, so a checkout without a built
+# library cannot load this file.
+#
+# Character validity is intentionally NOT tested here, because it is not the
+# wrapper's concern — glide-core validates the composed name before client
+# creation (valkey-io/valkey-glide#6891). What is tested is composition and the
+# empty-means-absent normalization, which the wrapper does own.
+class TestLibNameResolverUnit < Minitest::Test
```
Aryex: File header is development narration; the test names carry this.

```suggestion
class TestLibNameResolverUnit < Minitest::Test
```

test/unit/lib_name_resolver_test.rb
```diff
@@ -0,0 +1,172 @@
...
+  def test_default_when_nothing_configured
+    assert_equal "GlideRuby", Valkey.resolve_lib_name
+  end
+
+  # Pins the constant's literal value. The previous
+  # `assert_equal Valkey::DEFAULT_LIB_NAME, Valkey.resolve_lib_name` was
+  # tautological — both sides resolve through the same constant, so it passed
+  # even if the constant were changed to something invalid.
```
Aryex: Reviewer-facing history; one line on why a literal is enough.

```suggestion
  # Pinned as a literal so a change to the constant is deliberate.
```

test/unit/lib_name_resolver_test.rb
```diff
@@ -0,0 +1,172 @@
...
+    assert_equal "GlideRuby", Valkey.resolve_lib_name(client_info_tag: "")
+  end
+
+  def test_both_empty_falls_back_to_default
+    assert_equal "GlideRuby", Valkey.resolve_lib_name(lib_name: "", client_info_tag: "")
+  end
+
+  # Regression: an empty override with a tag previously composed "(tag)", which
+  # glide-core rejects because the name must not start with a parenthesis.
+  def test_empty_lib_name_with_tag_uses_default_base
```
Aryex: The test name states the behavior; drop the regression narration.

```suggestion
  def test_empty_lib_name_with_tag_uses_default_base
```

test/unit/lib_name_resolver_test.rb
```diff
@@ -0,0 +1,172 @@
...
+  # Previously any object was accepted via a blanket `to_s`. `lib_name: false`
+  # was the sharp edge — a caller passing a predicate got a client identifying
+  # itself to the server as "false" with no error. Arbitrary objects also let
+  # non-Valkey exceptions (JSON::GeneratorError, or anything raised by the
+  # object's own #to_s) escape Valkey.new, breaking the documented
+  # `rescue Valkey::BaseError` contract.
+  #
+  # This is a type check, not character validation — character validity is still
+  # glide-core's responsibility.
+  def test_false_is_rejected_rather_than_stringified
```
Aryex: Nine lines of change history; the test names and error matchers say it.

```suggestion
  def test_false_is_rejected_rather_than_stringified
```

test/unit/lib_name_resolver_test.rb
```diff
@@ -0,0 +1,172 @@
...
+  # rejected by core. This is settled policy, not incidental: client-side
+  # filtering is deliberately not performed, because the server is the authority
+  # on library-name validity and its error is the intended feedback for what is a
+  # caller mistake. Core's grammar excludes 0x20 (its ranges start at \x21), so
+  # "GlideRuby( )" fails validation and surfaces as Valkey::CannotConnectError at
+  # client creation.
+  #
+  # Do NOT "fix" this by normalizing whitespace away. Empty and whitespace are
+  # different cases on purpose: empty means absent (core returns Ok for it),
+  # whitespace is invalid input.
```
Aryex: Keep only the external fact; drop review IDs and reviewer instructions.

```suggestion
  # Core's name grammar starts at \x21, so "GlideRuby( )" is rejected at
  # client creation rather than folded to absent here.
```

test/unit/lib_name_resolver_test.rb
```diff
@@ -0,0 +1,172 @@
...
+  #   attempt 2: valid_encoding? -> binary strings still escaped (every byte
+  #                                 sequence is "valid" ASCII-8BIT)
+  #   attempt 3: #encode alone   -> a String already tagged UTF-8 but holding
+  #                                 invalid bytes still escaped (#encode is a
+  #                                 no-op when the encoding already matches)
+  #
+  # So the decisive test is the property one at the end of this section, not the
+  # per-input cases above it.
+
+  def test_rejects_invalid_utf8_bytes
```
Aryex: Section divider plus a development diary of superseded attempts.

```suggestion
  def test_rejects_invalid_utf8_bytes
```

test/unit/lib_name_resolver_test.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  def test_accepts_binary_string_that_is_representable_as_utf8
+    assert_equal "plain", Valkey.resolve_lib_name(lib_name: "plain".dup.force_encoding("ASCII-8BIT"))
+  end
+
+  def test_accepts_other_encodings_convertible_to_utf8
+    # ISO-8859-1 "café" converts cleanly, so it passes the wrapper untouched and
+    # is left for glide-core to judge — bytes are screened here, characters are not.
+    assert_equal "café", Valkey.resolve_lib_name(lib_name: "caf\xE9".dup.force_encoding("ISO-8859-1"))
```
Aryex: Restates the test name and repeats the core-owns-characters point.

```suggestion
    assert_equal "café", Valkey.resolve_lib_name(lib_name: "caf\xE9".dup.force_encoding("ISO-8859-1"))
```

test/unit/lib_name_resolver_test.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  def test_accepts_valid_non_ascii_utf8
+    # Character validity is core's call, not ours: this must NOT raise here.
+    assert_equal "café", Valkey.resolve_lib_name(lib_name: "café")
```
Aryex: Fourth restatement of "character validity is core's job" in this file.

```suggestion
    assert_equal "café", Valkey.resolve_lib_name(lib_name: "café")
```

test/unit/version_parseable_test.rb
```diff
@@ -0,0 +1,112 @@
...
+# "8garbage" gated tests to RUN) and that it admitted "0"/"0.0" (the sentinel
+# values whose removal the guard exists to make permanent).
+#
+# Why this matters more than it looks: Version comparison is lenient by design —
+# `<=>` compares parts with #to_i, so a non-numeric part becomes 0 and an
+# unparseable string sorts BELOW every real version. Garbage therefore WINS a
+# minimum, so a single bad node can drag a whole cluster's version gates down to
+# "skip". A skipped test and a passing test look identical in suite output, so a
+# hole here is silent.
+class TestVersionParseableUnit < Minitest::Test
```
Aryex: File header is development history and reviewer argument, not contract.

```suggestion
class TestVersionParseableUnit < Minitest::Test
```

test/unit/version_parseable_test.rb
```diff
@@ -0,0 +1,112 @@
...
+  # Real-world forms that must be usable.
+  ACCEPTED = [
```
Aryex: The constant name already says these are the accepted forms.

```suggestion
  ACCEPTED = [
```

test/unit/version_parseable_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    "8.1.0",
+    "7.2",
+    "9.1.1",
+    "10.0.0",
+    "8.1.0-rc1",
+    "8.1.0-alpha.2"
+  ].freeze
+
+  # Everything that must NOT be trusted as a version.
+  REJECTED = [
```
Aryex: The constant name already says these are rejected.

```suggestion
  REJECTED = [
```

test/unit/version_parseable_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    "",             # a valueless INFO line yields this via HashifyInfo
+    "unstable",
+    "0",            # sentinel shape
+    "0.0",          # verbatim the removed sentinel
```
Aryex: Both restate the value; the old sentinel is not a live concept here.

```suggestion
    "0",
    "0.0",
```

test/unit/version_parseable_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    "00",
+    "0garbage",
+    "7.1garbage",
+    "8garbage",     # end-unanchored regex gated this to RUN: a false PASS
```
Aryex: Narrates a fixed bug rather than the case; the value speaks for itself.

```suggestion
    "8garbage",
```

lib/valkey.rb
Aryex: The compile-time-artifact paragraph is design rationale, not API documentation.

```suggestion
  # Default value reported for `CLIENT SETINFO LIB-NAME`.
```

lib/valkey.rb
Aryex: Trim to the composition rule; the `@raise` also misses the invalid-encoding case.

```suggestion
  # Resolves the effective `CLIENT SETINFO LIB-NAME` value, composing `base(tag)`.
  # An empty override or tag means "not configured" and is omitted. Character
  # validity is glide-core's (see valkey-io/valkey-glide#6891).
  #
  # @api private
  # @param lib_name [String, Symbol, nil] raw `lib_name` override, may be empty
  # @param client_info_tag [String, Symbol, nil] raw `client_info_tag`, may be empty
  # @return [String] the effective library name
  # @raise [ArgumentError] if either value is neither a String, a Symbol, nor nil,
  #   or is a String with an invalid byte sequence
```

lib/valkey.rb
Aryex: The `to_s` history paragraph belongs in the PR description, not the docstring.

```suggestion
  # Normalizes a client-identification value to a non-empty String or nil.
  #
  # @param field [String] option name, used in the error message
  # @param value [String, Symbol, nil] the raw configured value
  # @return [String, nil] the value, or nil when absent/empty
  # @raise [ArgumentError] unless the value is a String, Symbol, or nil with a
  #   valid byte sequence
```

lib/valkey.rb
Aryex: Worth keeping — it is a real Ruby gotcha — but two lines do it, and `valid_encoding?` is only vacuous for ASCII-8BIT, not "other encodings" generally.

```suggestion
    # Both checks are needed: #valid_encoding? catches invalid bytes in a String
    # already tagged UTF-8; #encode catches ASCII-8BIT, where every byte is valid.
```

lib/valkey.rb
Aryex: Duplicates the `DEFAULT_LIB_NAME` and `resolve_lib_name` docstrings.

```suggestion
    json_options["lib_name"] = self.class.resolve_lib_name(
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+# connection establishment. client_info_tag appends a parenthesized tag
+# to the default library name (GlideRuby), while lib_name provides a
+# full override.
+#
+# References:
+# - valkey-io/valkey-glide#6389 (Python reference implementation)
+# - valkey-io/valkey-glide#6891 (core-side library-name validation, which this
+#   file's rejection expectations depend on)
+module ValkeyTests
+  module ClientInfoTag
```
Aryex: File header restates what the options do; the tests and README already cover it. Stating the version floor once here lets every `omit_version("7.2")` below stand on its own.

```suggestion
module ValkeyTests
  # Every lib-name assertion needs CLIENT SETINFO, so Valkey/Redis 7.2+.
  module ClientInfoTag
```

test/integration/valkey/client_info_tag_test.rb
Aryex: Keep the one-line docstring plus why the regex anchors; drop the argument aimed at reviewers.

```suggestion
    # Asserts the EXACT lib-name reported by CLIENT INFO. The regex anchors on the
    # field delimiter so "CustomLibSURPRISE" / "CustomLib(unexpected)" do not pass.
```

test/integration/valkey/client_info_tag_test.rb
Aryex: Duplicates resolve_lib_name's docstring, and empty-means-absent is a wrapper rule, not core's — core rejects "GlideRuby()" (line 127).

```suggestion
    def test_empty_client_info_tag_is_treated_as_absent
```

test/integration/valkey/client_info_tag_test.rb
```diff
@@ -0,0 +1,235 @@
...
+    # --- Tag composition / empty handling (no character validation here) ---
+    #
+    # The wrapper performs no character validation of client_info_tag or
+    # lib_name; that is glide-core's responsibility (upstream #6891). The one
+    # wrapper-side rule is compositional: an empty tag is treated as absent,
+    # matching core's empty-means-absent semantics, so it must neither raise nor
+    # compose a "GlideRuby()" that core would reject.
+
+    def test_empty_client_info_tag_is_treated_as_absent
+      omit_version("7.2") # CLIENT SETINFO (lib-name) requires Valkey/Redis 7.2+
```
Aryex: Covered by the module-level note now.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
Aryex: This trailing comment repeats verbatim on ten `omit_version` lines; the version floor is stated once at the module level instead.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
Aryex: Development history; the test name already states the expected behavior.

```suggestion
    def test_empty_lib_name_with_tag_uses_default_base
```

test/integration/valkey/client_info_tag_test.rb
Aryex: Same trailing comment again; the module-level note covers it.

```suggestion
      omit_version("7.2")
```

test/integration/valkey/client_info_tag_test.rb
Aryex: CORE_REJECTED_TAGS and the literal "has space" values already say this.

```suggestion
    CORE_REJECTED_TAGS = {
```

test/integration/valkey/client_info_tag_test.rb
Aryex: Keep only the fact that makes the message assertion necessary; drop the plea to reviewers.

```suggestion
        # A connection failure also raises CannotConnectError, so the message
        # assertion is what distinguishes rejection from an unreachable server.
```

test/integration/valkey/client_info_tag_test.rb
Aryex: Keep provenance for the case list; drop the "deliberately do NOT" declaration and drift history.

```suggestion
    # Cases transcribed from glide-core's library-name regex (upstream #6891);
    # the accepted grammar is documented in README's Connection Options.
```

test/integration/valkey/client_info_tag_test.rb
Aryex: Back-reference restates the note above and the assertions below.

```suggestion
        error = assert_raises(Valkey::CannotConnectError) do
```

test/integration/valkey/client_info_tag_test.rb
Aryex: Same trailing comment again; the module-level note covers it.

```suggestion
      omit_version("7.2")
```

Aryex (approved):

PR #250, opened 2026-08-07
fix: remove monitor, shutdown, and debug commands

Aryex:
Remove three `Valkey` server-command methods that either dispatch to undefined helpers or expose a server surface that no glide-family client supports.

## Summary

- **`monitor`**: Monitor is supported as a separate API. See https://glide.valkey.io/how-to/monitoring/monitor-command/. 
- **`shutdown`**:  Dangerous command not implemented by other GLIDE clients.
- **`debug`**: Potentially dangerous development command. Not intended for public usage and not implemented by other GLIDE clients.
## Test plan

- [x] `rubocop` clean on touched files
- [x] Skipped `test_debug` case in `test/lint/server_commands.rb` removed
- [x] Live probes confirm `v.monitor` / `v.shutdown` / `v.debug` now raise plain `NoMethodError: undefined method '<name>' for an instance of Valkey` (not the previous `NoMethodError: synchronize` or `NameError`)
- [x] Remaining `server_commands` lint suite passes on standalone

### DIFF
--- a/lib/valkey/commands/server_commands.rb
+++ b/lib/valkey/commands/server_commands.rb
@@ -173,40 +173,12 @@ class Valkey
         send_command(RequestType::LAST_SAVE, [], route: route)
       end
 
-      # Listen for all requests received by the server in real time.
-      #
-      # There is no way to interrupt this command.
-      #
-      # @yield a block to be called for every line of output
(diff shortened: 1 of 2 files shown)


yipin-chen (approved):
LGTM. Make sure the document is updated.

Issue #133, opened 2026-07-08
Cluster Scan Support


Aryex:
## Overview

`redis-rb` support `scan()` method, allowing for scanning of a cluster. This needs to be adapted to 
GLIDE's native scanning capabilities.

## Description

redis-rb 4.x (before cluster support moved to `redis-cluster-client`) had its own cluster-aware
`scan`, confirmed directly from source (`redis-rb` v4.8.0, `lib/redis/cluster.rb`):

```ruby
def _scan(command, &block)
  input_cursor = Integer(command[1])
  client_index = input_cursor % 256
  raw_cursor = input_cursor >> 8

  clients = @node.scale_reading_clients
  client = clients[client_index]
  return ['0', []] unless client

  command[1] = raw_cursor.to_s
  result_cursor, result_keys = client.call(command, &block)

  result_cursor = Integer(result_cursor)
  client_index += 1 if result_cursor == 0

  [((result_cursor << 8) + client_index).to_s, result_keys]
end
```

## Proposal

**Target: redis-rb 4.x's `scan`/`scan_each` split — keep `scan(cursor, **options)` exactly as it
is, implement `scan_each`.** This client's `scan` already matches 4.x's `scan` (one page per call,
caller manages the cursor) — no change needed there beyond adding the cluster-mode branch below.
`scan_each` already has its full YARD doc written in this codebase
(`generic_commands.rb:33-52`), matching 4.x's `scan_each` exactly — only the method body is
missing:

```ruby
# redis-rb 4.x's scan_each, confirmed directly from source (v4.8.0, lib/redis/commands/keys.rb):
def scan_each(**options, &block)
  return to_enum(:scan_each, **options) unless block_given?
  cursor = 0
  loop do
    cursor, keys = scan(cursor, **options)
    keys.each(&block)
    break if cursor == "0"
  end
end
```

Implement this directly — no redesign of `scan` needed, since `scan_each` just calls `scan` in a
loop. Making `scan` itself branch on `cluster_mode?` (plain `SCAN` vs. `request_cluster_scan`) is
enough to make both methods cluster-correct automatically; `scan_each` doesn't need its own
separate cluster-aware logic.

### Example

```ruby
# scan: unchanged from today, one page per call, caller manages the cursor —
# now cluster-correct via the branch in task 2.
cursor, keys = client.scan("0", match: "user:*")
# => ["17", ["user:1", "user:2"]]

[...]

PR #252, opened 2026-08-07
fix: short-circuit `scan`/`scan_each` to empty in cluster mode

Aryex:
## Summary
`scan` and `scan_each` are standalone mode only. This PR adds additional comments clarifications. It also return empty results if scan is used on `cluster_mode`. 

Cluster support to be added in https://github.com/valkey-io/valkey-glide-ruby/issues/133

## Test plan

- [x] `bundle exec rubocop lib/valkey/commands/generic_commands.rb test/lint/generic_commands.rb` — clean
- [x] `ruby -c` on both files — syntax OK
- [ ] `bundle exec rake test:standalone` — deferred (no local standalone server up at the time of push); please verify in CI
- [ ] `bundle exec rake test:cluster` — deferred (no local cluster up at the time of push); the new `cluster_mode?` branches in `test_scan` / `test_scan_each` are what exercises the guard, so CI is the load-bearing check here

### DIFF
--- a/lib/valkey/commands/generic_commands.rb
+++ b/lib/valkey/commands/generic_commands.rb
@@ -9,10 +9,10 @@ class Valkey
     module GenericCommands
       # Scan the keyspace
       #
-      # @note Standalone mode only. glide-core has no defined default route for
-      #   SCAN in cluster mode, so each call may land on a different node with
-      #   no cursor continuity between them — results are undefined (missed or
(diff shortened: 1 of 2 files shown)


jamesx-improving (approved):
LGTM

yipin-chen (approved):
LGTM. nit: Should we remove `TODO` in the comment and just keep issue links?

Issue #135, opened 2026-07-08
Pub/Sub Support


Aryex:
## Overview

Pub/sub in the Ruby client is partially implemented:
- The publish and introspection commands work and are tested, 
- The FFI plumbing for message delivery is already in place
 
However, there is no way to consume messages, thus Pub/Sub is not usable in practice.

## Current State

**Working (implemented and covered by `test/lint/pubsub_commands.rb`):**

- `publish`, `spublish`
- `pubsub_channels`, `pubsub_numpat`, `pubsub_numsub`, `pubsub_shardchannels`,
  `pubsub_shardnumsub` (plus the `pubsub(:subcommand, ...)` convenience method)

**Partially working:**

- `subscribe`, `unsubscribe`, `psubscribe`, `punsubscribe`, `ssubscribe`,
  `sunsubscribe` exist in `lib/valkey/commands/pubsub_commands.rb` with request-type
  IDs (901–913) matching glide-core's protobuf enum. The core accepts these — its
  `PubSubSynchronizer` intercepts subscription commands and manages them on the
  multiplexed connection — so the subscription itself is established.
- The FFI callback is already wired: `create_client` / `create_client_from_uri`
  accept a `PubSubCallback`, and the client registers one at creation
  (`lib/valkey.rb`). Push messages do arrive in Ruby.

**Missing:**

- Message consumption. The registered callback (`lib/valkey/pubsub_callback.rb`) is a
  hardcoded debug `puts` — users cannot supply their own handler and there is no
  message queue to poll. Messages are printed to stdout and lost.
- Blocking subscribe variants. `request_type.rb` stops at 913; the core's
  `SubscribeBlocking`/`UnsubscribeBlocking`/`PSubscribeBlocking`/
  `PUnsubscribeBlocking`/`SSubscribeBlocking`/`SUnsubscribeBlocking` (914–918) —
  which wait for server confirmation — are not exposed.
- Tests and examples. The lint suite covers only publish and introspection; there is
  zero coverage of the subscribe/unsubscribe family and no pub/sub examples.

## Proposal

Follow the Python sync client (`python/glide-sync`), which sits on the same
`libglide_ffi` C API and is a direct blueprint:

1. **Delivery**: replace the debug callback with two mutually exclusive modes —
   a user-supplied callback (+ optional context) configured at client creation, or an
   internal queue with `get_pubsub_message` (blocking) / `try_get_pubsub_message`
   (non-blocking).
[...]

PR #254, opened 2026-08-10
fix: remove pubsub public API and skip upstream client KILL bug

Aryex:
### Summary

Removes PubSub from the public API and works around an upstream `CLIENT KILL` bug.

### Changes

- **PubSub:** stop including `PubSubCommands` in the public client, so the unsupported PubSub API is no longer exposed. Tracked in #135.
- **CLIENT KILL tests:** the address form of `CLIENT KILL` currently raises `expected array of integers as response`, because glide-core routes it to all nodes with Sum aggregation and that rejects the `OK` reply. `test_client_kill` and `test_client_kill_simple` now assert the error with `assert_raises` and carry a TODO to restore `assert_equal "OK"` once it's fixed. Tracked in #256 (upstream valkey-io/valkey-glide#6750).
- **connection_commands.rb:** move `build_client_kill_args` into the shared `private` block. No behavior change.

### Testing

`bundle exec rubocop` is clean. The CLIENT KILL tests run in the standalone and cluster lint suites in CI.

### DIFF
--- a/lib/valkey/commands.rb
+++ b/lib/valkey/commands.rb
@@ -47,7 +47,8 @@ class Valkey
     include SetCommands
     include ScriptingCommands
     include FunctionCommands
-    include PubSubCommands
+    # TODO: https://github.com/valkey-io/valkey-glide-ruby/issues/135
+    # include PubSubCommands
     include ClusterCommands
     include TransactionCommands
     include StreamCommands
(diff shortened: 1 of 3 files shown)


yipin-chen (approved):

jamesx-improving (approved):
LGTM

Issue #290, opened 2026-08-17
[Epic] Ruby 1.1 Release


yipin-chen:
### Description

This is the epic issue to track Ruby 1.1 release alongside Valkey-GLIDE 2.6.
- Scheduled for September 22
- Code freeze on September 4

### Support and Bugfixes
- [ ] [close() frees the client while a command is in-flight](https://github.com/valkey-io/valkey-glide-ruby/issues/284)
- [ ] [Client crashes child process on fork()](https://github.com/valkey-io/valkey-glide-ruby/issues/255)
- [ ] [Pub/Sub Support](https://github.com/valkey-io/valkey-glide-ruby/issues/135)
- [ ] [Code Cov reporting](https://github.com/valkey-io/valkey-glide-ruby/issues/288)

### Additional Bugfixes (issue #275)
- [x] [`xpending(idle:)` places IDLE in the consumer-name slot; silently returns zero results](https://github.com/valkey-io/valkey-glide-ruby/issues/270)
- [ ] [`Valkey.new(cluster_mode: true)` with no seeds silently uses `127.0.0.1:6379`](https://github.com/valkey-io/valkey-glide-ruby/issues/265)
- [ ] [`Valkey.new(nodes: [...])` only uses the first entry; every other seed is silently discarded](https://github.com/valkey-io/valkey-glide-ruby/issues/264)
- [x] [`client.multi` with no block silently returns self with `@in_multi = true`](https://github.com/valkey-io/valkey-glide-ruby/issues/263)
- [x] [`watch(*keys) do ... end` block form doesn't clean up on exception](https://github.com/valkey-io/valkey-glide-ruby/issues/262)
- [x] [`pipelined` / `multi` drop the entire reply array when any queued command errors](https://github.com/valkey-io/valkey-glide-ruby/issues/260)

### Reference
From the [GA issue](https://github.com/valkey-io/valkey-glide/issues/5798), the following feature-parities will be added post-GA, based on priority and available bandwidth:

Compression support
Inflight request limit
Password update API
AZ affinity API
Lazy connect
IAM Authentication
Refresh topology
JSON module
FT module
Client Side Caching

Issue #234, opened 2026-08-05
FT Search Module Support


Aryex:
Implement support for FT Search modules.

jbrinkman:
This is currently being worked by @Jonathan-Improving as part of our effort to synchronize Valkey-Search across the languages.  The initial focus is on getting Valkey-Search 1.2 compatibility (see Valkey-Glide issue #[5627](https://github.com/valkey-io/valkey-glide/issues/5627)). We will have a second set of issues to implement Search Module support across all languages for Valkey-Search 1.3 and expect to begin that work in September with the RC release of VS 1.3.

Jonathan-Improving:
## High-level design — Ruby structured Valkey Search (FT.*) API

Posting the design up front (per the ask to avoid the kind of post-GA churn the C# client saw). Feedback welcome **before** the remaining phases land. Tracking PR: #295 (draft, builds up in exactly 4 commits, one per phase). Merge target: before the 2026-09-04 release window.

### Goals & constraints

- **Parity** with the other language clients for Valkey-Search 1.2, using the **Java client as the reference authority** (it is the most mature).
- **Backward compatible**: every existing raw `ft_*(*args)` call keeps working unchanged. The builder API is purely additive and selected by argument type.
- **Ruby-idiomatic**: snake_case, symbols, keyword arguments; builders only where tree-shaped structure justifies them.
- **Lean by design — hard PR budget of < 600 lines.** The C# effort ran ~4K lines and generated significant post-GA rework; this PR is deliberately scoped and compact to avoid that. Anything over 600 is a red flag to be trimmed (consolidating files, tightening YARD, and keeping tests proportionate).

### Approach (hybrid)

Typed builder objects for tree-shaped structures (field definitions, aggregate pipelines); plain keyword arguments for flat option sets (search/create). Structured result object for FT.SEARCH (Go/C# shape) — the raw path still returns the unwrapped reply.

Namespace: `Valkey::Search`. Dispatch rule: when the first argument is a builder/collection type, take the (fully validated) builder path; otherwise forward raw tokens verbatim.

### Phases (one commit each)

1. **Field types + CreateOptions** — `TextField`, `TagField`, `NumericField`, `VectorField` (`.flat`/`.hnsw`), `CreateOptions`; `ft_create` builder path. ✅ landed
2. **SearchOptions + SearchResult** — query options + structured result; `ft_search` returns `SearchResult` on the builder path. ✅ landed
[...]

PR #65, opened 2025-12-01
Implementing vector search commands

pratheep-kumar:
Resolves #62

## Add Full Vector Search (RediSearch) Commands

This PR adds complete support for RediSearch/Vector Search commands to `Valkey::Commands::VectorSearchCommands`. All 13 FT (Full-Text Search) commands are implemented in a modular, redis-rb-compatible way with full backend support.

#### Index Management Commands

| Redis CLI Command | Ruby Method | Description |
|-------------------|-------------|-------------|
| `FT.LIST` | `ft_list` | Lists all existing indexes |
| `FT.CREATE` | `ft_create(index, *args)` | Creates an index with schema definition |
| `FT.DROPINDEX` | `ft_drop_index(index, dd: false)` | Drops an index, optionally with documents |
| `FT.INFO` | `ft_info(index)` | Returns information and statistics about an index |

#### Search & Query Commands

| Redis CLI Command | Ruby Method | Description |
|-------------------|-------------|-------------|
| `FT.SEARCH` | `ft_search(index, query, *args)` | Searches the index with a query string |
| `FT.AGGREGATE` | `ft_aggregate(index, query, *args)` | Executes aggregation queries on an index |
| `FT.EXPLAIN` | `ft_explain(index, query)` | Returns the execution plan for a query |
| `FT.EXPLAINCLI` | `ft_explain_cli(index, query)` | Returns CLI-formatted execution plan |
| `FT.PROFILE` | `ft_profile(index, type, *args)` | Profiles query execution (SEARCH or AGGREGATE) |

#### Alias Management Commands

| Redis CLI Command | Ruby Method | Description |
|-------------------|-------------|-------------|
| `FT.ALIASADD` | `ft_alias_add(alias_name, index)` | Adds an alias to an index |
| `FT.ALIASDEL` | `ft_alias_del(alias_name)` | Deletes an alias from an index |
| `FT.ALIASLIST` | `ft_alias_list` | Lists all aliases |
| `FT.ALIASUPDATE` | `ft_alias_update(alias_name, index)` | Updates an alias to point to a different index |

### ✅ Key Features

- **Redis-rb Compatible**: All methods follow redis-rb conventions for developer familiarity
- **Idiomatic Ruby**: Returns data in natural Ruby types (arrays, strings, hashes)
- **Flexible Index Creation**: `ft_create` accepts variadic arguments for complex schema definitions (ON, PREFIX, SCHEMA, etc.)
[...]

PR #295, opened 2026-08-21
feat(ruby): structured Valkey Search (FT.*) builder API

Jonathan-Improving:
### Summary

Structured, Ruby-idiomatic builder API for Valkey Search (`FT.*`) commands, bringing the Ruby GLIDE client toward parity with the other language clients (modeled on Java GLIDE's field/option builders, adapted to Ruby idiom: snake_case, symbols, kwargs). Tracks #234 (FT Search Module Support); part of the Ruby 1.1 epic #290.

All four planned phases are complete: field/index builders, the query builder with structured results, the aggregate pipeline builders, and info/cluster options. During review, the original redis-rb-style raw-token pass-through was removed entirely in favor of the builders as the sole supported path (breaking change — see below), and several commands present in the initial port were removed because the server does not implement them.

### Issue link

- Related #234 (FT Search Module Support)
- Sub task of #290 (Ruby 1.1 epic)

### Changes

- **Field/index builders (`FT.CREATE`)**: `TextField`, `TagField`, `NumericField`, `VectorField.flat`/`.hnsw`, and `CreateOptions`, each serializing to the FT.CREATE token array.
- **Query builder (`FT.SEARCH`)**: `SearchOptions` plus a structured `Valkey::Search::SearchResult` (`total_results` + `documents`, each a `Document` with `key`/`fields`/`sort_key`). `ft_search` always returns a `SearchResult`.
- **Aggregate pipeline builders (`FT.AGGREGATE`)**: `AggregateOptions` with ordered pipeline clauses — `GroupBy`/`Reducer`, `SortBy`, `Filter`, `Apply`, `Limit`.
- **Info/cluster options (`FT.INFO`)**: `InfoOptions`, plus `shard_scope:`/`consistency:` flags where the server actually accepts them (`ft_search`/`ft_info`, not `ft_aggregate`).
- **Breaking: raw-token pass-through removed.** The builders are now the only supported way to pass options; the old redis-rb-style `ft_search("idx", "@title:hello", "LIMIT", "0", "10")` form and the `ft(:subcommand, ...)` dispatcher are gone. Use `Valkey#call("FT.SEARCH", ...)` to send FT tokens by hand. Collapsed the dual builder-vs-raw dispatch into a single options resolver (`vector_search_commands.rb` 406 → 287 lines).
- **Removed commands the server doesn't implement.** Verified against valkey-search source rather than inferred from errors: only `FT._LIST`, `FT.CREATE`, `FT.DROPINDEX`, `FT.INFO`, `FT.SEARCH`, `FT.AGGREGATE` exist. Removed the `FT.ALIASADD`/`ALIASDEL`/`ALIASUPDATE`/`_ALIASLIST`, `FT.EXPLAIN`, `FT.EXPLAINCLI`, `FT.PROFILE` wrappers carried over from the original port. Also removed `ft_drop_index`'s `dd:` flag — `FT.DROPINDEX` has arity 2, so it was producing a wrong-arity error.
- **Cluster support, client-side half.** Index definitions are per-node, so `ft_create`/`ft_drop_index` now broadcast to all primaries (`Route.all_primaries`), collapsing a uniform `"OK"` reply from every node back to `"OK"` for the caller.
- **Fail loudly instead of returning garbage**, in three cases the review surfaced: `no_content:` + `with_sort_keys:` is rejected at construction (the server drops sort keys under `NOCONTENT`, desyncing the reply); `ft_search` raises `ArgumentError` inside `pipelined`/`multi` and with `flatten_map: true` instead of a `NoMethodError` or a corrupt `SearchResult`; a malformed `limit:` Hash raises the documented `ArgumentError` instead of `KeyError`.
- Re-enabled the `module-tests` CI job (previously gated behind `if: false`), so the search suite now runs against a live search module in CI.
- `CHANGELOG.md` updated under `## 1.1.0 (Pending)`.

### Limitations

- **Cross-shard result aggregation is not supported.** Measured on a 3-primary cluster (Valkey 9.1.1, search 1.2.1): documents are indexed on the shard owning their key, and `FT.SEARCH`/`FT.AGGREGATE` return shard-local matches only. `shard_scope: :all_shards` is accepted by the server but does not fan out — a query against the shard holding 1 of 6 seeded documents returned 1, not 6. No client-side change can fix this; it needs server support. The cluster test suite is left disabled with the measurement and a reproduction command recorded in `test/integration/cluster/cluster_commands_test.rb`, rather than made to pass by hash-tagging keys into one slot.
- `RequestType::FT_*` constants for the removed commands are kept (AGENTS.md requires `RequestType` to mirror the glide-core enum); open to removing them if preferred.
- The `module-tests` CI job is standalone-only — it does not cover cluster, so a cluster regression would not fail CI. The docker action already provisions module cluster nodes (ports 8001-8006) that nothing currently connects to; wiring them in is a follow-up, not blocked on anything here.
- `DIALECT` is validated to `2`, the only dialect Valkey Search supports.

### Testing

- Unit tests: 136 runs / 0 failures across field, query, aggregate, and info builder serialization + dispatch guards.
- `rubocop` clean (119 files, 0 offenses).
- Live standalone integration tests against a real search module (valkey-search 1.2.1): 29 tests, 101 assertions, 0 failures/errors/skips.
- **Mutation-tested the integration suite** by deliberately injecting faults into the builders (wrong `LIMIT` count/offset, dropped `SORTBY`/`NOCONTENT`, swapped vector algorithm, dropped `LOAD`) and confirming the suite fails; 18/18 injected faults are now caught. This found and fixed several previously-weak assertions, including every `GROUPBY` test asserting against an ungrouped result (the server silently collapses groups when the grouped/reduced field isn't in `LOAD`, rather than erroring).
- Cluster: measured directly on a 3-primary cluster with a live search module (see Limitations); the suite itself stays disabled with the measurement recorded for the next person.

### Checklist


Jonathan-Improving:
Shortened / made more terse (excluding tests)

Jonathan-Improving:
Build pipeline failing with `note: error: unsupported linker arg: --fix-cortex-a53-843419`

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -25,18 +25,46 @@ def ft_list
...
+      #        valkey.ft_aggregate("idx", "@price:[0 +inf]",
+      #          clauses: [
+      #            Valkey::Search::GroupBy.new(["@category"],
+      #              reducers: [Valkey::Search::Reducer.count(as: "count")]),
+      #            Valkey::Search::SortBy.new("@count", :desc),
+      #            Valkey::Search::Limit.new(0, 10),
+      #          ], dialect: 2)
+      #
+      # 2. **Raw args** (backward compatible) — pass FT.AGGREGATE tokens directly:
+      #
```
Aryex: Not required. The existing implementation is an old port from redis-rb that we removed before GA. 

Can we remove support for these and simplify the implementations?

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -258,14 +343,129 @@ def ft_search(index, query, *args)
       def ft(subcommand, *args, **options)
```
Aryex: Similarly, I think this could be removed? This is a `redis-rb` styled command that we don't need to support.

yipin-chen (approved):
Conditional approval, pending Alex' final blessing.
Please update changelog.

test/integration/standalone/commands_test.rb
```diff
@@ -54,7 +54,7 @@ class TestStandaloneGeoCommands < Minitest::Test
 #     include Lint::ModuleCommands
 #   end
 #
-#   class TestStandaloneVectorSearchCommands < Minitest::Test
-#     include Helper::Client
-#     include Lint::VectorSearchCommands
-#   end
+class TestStandaloneVectorSearchCommands < Minitest::Test
```
Aryex, line 57: I'm guessing cluster mode is not supported for this?

test/lint/vector_search_commands.rb
Aryex, line 1: Module tests were previously disabled in ci/cd. Will need to re-enable them.

lib/valkey/search/query.rb
```diff
@@ -0,0 +1,214 @@
...
+      # @param params [Hash, nil] query parameters (PARAMS)
+      # @param sort_by [String, nil] field to sort by
+      # @param sort_order [Symbol] :asc (default) or :desc
+      # @param shard_scope [Symbol, nil] :all_shards or :some_shards
+      # @param consistency [Symbol, nil] :consistent or :inconsistent
+      def initialize(limit: nil, return_fields: nil, params: nil, sort_by: nil,
+                     sort_order: :asc, with_sort_keys: false, no_content: false,
+                     verbatim: false, in_order: false, slop: nil, dialect: nil,
+                     timeout: nil, shard_scope: nil, consistency: nil)
+        raise ArgumentError, "with_sort_keys requires sort_by" if with_sort_keys && sort_by.nil?
```
Aryex, line 34: Should `no_content: true` + `with_sort_keys: true` be allowed here? The [FT.SEARCH docs](https://valkey.io/commands/ft.search/) list both options but say nothing about combining them.

Tested against Valkey 9.1.1, search 1.2.1. The server discards the sort keys when `NOCONTENT` is set.

```
FT.SEARCH a1_idx hello NOCONTENT SORTBY price ASC WITHSORTKEYS
1) (integer) 2
2) "a1:1"
3) "a1:2"

FT.SEARCH a1_idx hello NOCONTENT
1) (integer) 2
2) "a1:1"
3) "a1:2"
```

glide-core still pairs the elements as `key, sortkey`, so what comes back depends on how many documents matched. With 5 matching docs:

```
n=1  Valkey::CommandError: WITHSORTKEYS NOCONTENT response: missing sort key after document key
n=2  [5, {"wsk:4" => ["wsk:5", {}]}]
n=3  CommandError
n=4  [5, {"wsk:4" => ["wsk:5", {}], "wsk:2" => ["wsk:3", {}]}]
n=5  CommandError
```

Perhaps worth it to document this behavior?

lib/valkey/search/query.rb
```diff
@@ -0,0 +1,214 @@
...
+        @documents = documents
+      end
+
+      # Parse a raw FT.SEARCH reply. Handles both the GLIDE normalized map form
+      # ([count, { key => { field => value } }]) and the flat RESP2 array form,
+      # honoring no_content and with_sort_keys.
+      #
+      # @return [SearchResult]
+      def self.from_raw(raw, no_content: false, with_sort_keys: false)
+        return new(0, []) if raw.nil? || raw.empty?
```
Aryex, line 151: `ft_search` raises inside `pipelined` / `multi`. 

Valkey 9.1.1, search 1.2.1:

```ruby
r.pipelined { |p| p.ft_search("idx", "hello", dialect: 2) }
# NoMethodError: undefined method 'empty?' for an instance of Valkey::Future

r.pipelined { |p| p.ft_search("idx", "hello") }        # raw path
# => [[1, {"pl:1" => {"title" => "hello world"}}]]

r.pipelined { |p| p.hgetall("pl:1") }
# => [{"title" => "hello world"}]
```

The `Valkey::Pipeline` does `include Commands` so every `ft_*` method is callable in a batch.

Note that this is the redis-rb style batch API. We plan on adding GLIDE batch API so incompatibilities here would be ok.

lib/valkey/search/query.rb
```diff
@@ -0,0 +1,214 @@
...
+
+        total = coerce_count(raw[0])
+        rest = raw[1..] || []
+        documents =
+          if rest.length == 1 && rest[0].is_a?(Hash)
+            parse_map(rest[0], no_content: no_content, with_sort_keys: with_sort_keys)
+          else
+            parse_flat(rest, no_content: no_content, with_sort_keys: with_sort_keys)
```
Aryex, line 159: With `flatten_map: true`, builder `ft_search` returns a corrupt `SearchResult` rather than failing.

Note that `flatten_map` is purely a backward compat options. I think we just need to document this, but bonus if we can make it work without complicating the implementations.

#### More Details

`flatten_map` flattens every map reply, so `rest[0]` is an Array and `from_raw` routes to `parse_flat`. Valkey 9.1.1, search 1.2.1, four matching documents:

```ruby
fm = Valkey.new(host: h, port: p, flatten_map: true)

fm.ft_search("fm_idx", "hello", dialect: 2)
# TypeError: FT.SEARCH reply truncated (missing field payload)

fm.ft_search("fm_idx", "hello", no_content: true, dialect: 2)
# => total_results = 4, documents.length = 1
#    documents.first.key == ["fm:2", [], "fm:3", [], "fm:4", [], "fm:1", []]

fm.ft_search("fm_idx", "hello")     # raw path
# => [4, ["fm:2", ["title", "hello n2"], ...]]
```

The `no_content` branch skips the field-payload bounds check at line 197, so nothing raises and the caller gets a single Document whose `key` holds the whole flattened reply.

`flatten_map` is described at `lib/valkey.rb:284-286` as "Compatibility for redis-rb 4.x", so this is probably a docs change rather than a code one: state that the option covers the raw-args path only and is not supported alongside the builder API.

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,301 @@
...
+      # @param verbatim [Boolean] emit VERBATIM
+      # @param in_order [Boolean] emit INORDER
+      # @param slop [Integer, nil] emit `SLOP <n>`
+      # @param dialect [Integer, nil] emit `DIALECT <n>` (only 2 is valid)
+      # @param timeout [Integer, nil] emit `TIMEOUT <ms>`
+      # @param shard_scope [Symbol, nil] :all_shards or :some_shards
+      # @param consistency [Symbol, nil] :consistent or :inconsistent
+      def initialize(clauses: [], load: nil, params: nil, verbatim: false,
+                     in_order: false, slop: nil, dialect: nil, timeout: nil,
+                     shard_scope: nil, consistency: nil)
```
Aryex: Does FT.AGGREGATE accepts shard_scopes? Per [docs](https://valkey.io/commands/ft.aggregate/) there's no mention of it. 

Also tested against Valkey 9.1, search 1.2.1:

```ruby
r.ft_aggregate("idx", "@price:[0 100]", load: ["@category"], clauses: grp,
               shard_scope: :all_shards)
# Valkey::CommandError: Unexpected: argument `ALLSHARDS`
```

All four values behave the same way:

```
shard_scope: :all_shards    ->  Unexpected: argument `ALLSHARDS`
shard_scope: :some_shards   ->  Unexpected: argument `SOMESHARDS`
consistency: :consistent    ->  Unexpected: argument `CONSISTENT`
consistency: :inconsistent  ->  Unexpected: argument `INCONSISTENT`
```

lib/valkey/search/query.rb
```diff
@@ -0,0 +1,214 @@
...
+      end
+
+      def normalize_limit(limit)
+        case limit
+        when nil then nil
+        when Array
+          raise ArgumentError, "limit Array must be [offset, count], got #{limit.inspect}" unless limit.length == 2
+
+          limit
+        when Hash then [limit.fetch(:offset), limit.fetch(:count)]
```
Aryex: Should a malformed `limit:` Hash raise `ArgumentError`? `ft_search` documents `@raise [ArgumentError] on a malformed builder invocation`, but `limit.fetch(:offset)` and `limit.fetch(:count)` raise `KeyError` when a key is missing.

The other rejecting branches do raise `ArgumentError`:

```
limit: [0]          ->  ArgumentError: limit Array must be [offset, count], got [0]
limit: "x"          ->  ArgumentError: limit must be a { offset:, count: } Hash or [offset, count] Array
limit: {offset: 0}  ->  KeyError: key not found: :count
```

`KeyError < IndexError`, so a rescue on the documented class won't catch it.

Perhaps something like this could fix it
```ruby
when Hash
  offset = limit[:offset] or raise ArgumentError, "limit Hash requires :offset"
  count  = limit[:count]  or raise ArgumentError, "limit Hash requires :count"
  [offset, count]
```

lib/valkey/commands.rb
```diff
@@ -51,6 +51,7 @@ module Commands
     # include PubSubCommands
     include ClusterCommands
     include TransactionCommands
+    include VectorSearchCommands
```
Aryex, line 54: This exposes 7 commands that `VectorSearchCommands` implemented but the server could not found (Valkey 9.1, search 1.2.1): 

`FT.ALIASADD`, `FT.ALIASDEL`, `FT.ALIASUPDATE`, `FT._ALIASLIST`, `FT.EXPLAIN`, `FT.EXPLAINCLI`, `FT.PROFILE`

All fail with `unknown command`. Do you know if these existed in previous versions? Otherwise we should remove them.

Aryex (commented):
Reviewed up to [d80aa64](https://github.com/valkey-io/valkey-glide-ruby/pull/295/commits/d80aa64a2abeac7a95f344f4fa2fdbb6f24d6a0a)

Referenced against: Valkey 9.1, search 1.2.1.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Removed. Taking this as "remove the raw-args passthrough and simplify" — the whole redis-rb-style pass-through is gone, so the builders are now the only way to pass options. `Valkey#call` is the escape hatch for anyone who wants to send FT tokens by hand.

Collapsing the dual builder-vs-raw dispatch let a lot go with it: `vector_search_commands.rb` went **406 → 287 lines**, and the three per-command option resolvers plus the `resolve_search_options` state machine (with its `allow_raw` branching) became one small `ft_resolve_options`.

New signatures:
```ruby
ft_create(index, fields, options = nil, route: nil, **kwargs)
ft_search(index, query, options = nil, **kwargs)   # always returns a SearchResult
ft_aggregate(index, query, options = nil, **kwargs)
ft_info(index, options = nil, **kwargs)
```
This is a breaking change and is called out as such in the CHANGELOG.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Removed — the `ft(:subcommand, ...)` dispatcher is gone, along with its SEC-001 regression test (which only existed to prove `public_send` could not reach private helpers through it).

lib/valkey/search/aggregate.rb
Jonathan-Improving: Confirmed and removed — you are right, and I checked it against the module source rather than just taking the error at face value.

`ALLSHARDS`/`SOMESHARDS` appear only in the **ft_info** and **ft.search** parsers in valkey-search; there is no such token in the aggregate parser or `ft.aggregate.json`. So `shard_scope:`/`consistency:` are valid on `FT.SEARCH` and `FT.INFO` but genuinely invalid on `FT.AGGREGATE`, exactly as your `Unexpected: argument ALLSHARDS` shows.

`AggregateOptions` no longer accepts either keyword, and there is now a test asserting they are rejected so this cannot silently come back:
```ruby
def test_aggregate_options_rejects_cluster_flags
  assert_raises(ArgumentError) { S::AggregateOptions.new(shard_scope: :all_shards) }
  assert_raises(ArgumentError) { S::AggregateOptions.new(consistency: :consistent) }
end
```
Related bug this surfaced: `AggregateOptions#to_args` still called `Search.append_cluster_flags(args, @shard_scope, @consistency)` after I removed the params. It referenced ivars that no longer existed and only *appeared* harmless because Ruby reads undefined ivars as `nil` — a silent no-op rather than a crash. Also removed.

lib/valkey/search/query.rb
Jonathan-Improving: Fixed — and thanks for spotting that `KeyError < IndexError`, which is the part that makes it a real bug rather than a cosmetic one: a caller rescuing the documented `ArgumentError` would not catch it.

I used explicit key checks rather than the `or raise` form, so a legitimately falsey value can never be mistaken for a missing key:
```ruby
when Hash
  raise ArgumentError, "limit Hash requires :offset, got #{limit.inspect}" unless limit.key?(:offset)
  raise ArgumentError, "limit Hash requires :count, got #{limit.inspect}"  unless limit.key?(:count)

  [limit[:offset], limit[:count]]
```
Tests cover both missing keys and `{ offset: 0, count: 0 }` staying valid.

lib/valkey/search/query.rb
Jonathan-Improving: Went further than documenting it — the combination is now **rejected up front**, because your `n=1..5` table shows the result is not merely undocumented but unusable: the reply either mispairs documents or raises, depending on how many matched.

The server honours `NOCONTENT` and drops the sort keys, but glide-core still pairs elements as `key, sortkey`, so the parser desynchronises. There is no reading of that reply we can trust, so returning *something* would be worse than failing:
```ruby
if with_sort_keys && no_content
  raise ArgumentError,
        "with_sort_keys cannot be combined with no_content: the server omits sort keys " \
        "under NOCONTENT, so the reply cannot be parsed reliably"
end
```
`with_sort_keys` on its own (with `sort_by`) is unaffected and still emits `WITHSORTKEYS`. I kept the defensive branch in `SearchResult.parse_map` for that sub-case, since `from_raw` is public and can still be handed such a reply directly.

lib/valkey/search/query.rb
Jonathan-Improving: Fixed, and noted on the redis-rb-batch point — but a `NoMethodError` from inside the parser is not an acceptable way to express "unsupported", so it now fails deliberately:
```
ArgumentError: ft_search is not supported inside pipelined/multi;
use Valkey#call to issue FT.SEARCH and handle the raw reply
```
Detection is duck-typed on the receiver queueing commands (`@futures`/`@commands`, i.e. a `Valkey::Pipeline`) or `@in_multi`, rather than `is_a?(Valkey::Pipeline)` — a unit test caught that the constant is not necessarily loaded in every context, which would have swapped one `NameError` for another.

Since the raw-args path is now gone (see the thread on `:42`), `Valkey#call` is the documented alternative rather than raw tokens. Both limitations are in the `ft_search` YARD.

lib/valkey/search/query.rb
Jonathan-Improving: Fixed — and your root-cause analysis was exactly right, including the subtle part: the `no_content` branch skips the field-payload bounds check, which is why nothing raised and the caller got a single `Document` whose `key` held the entire flattened reply.

Rather than document a corrupt result as expected behaviour, the builder path now rejects it:
```
ArgumentError: ft_search is not supported with flatten_map: true;
use Valkey#call to issue FT.SEARCH and handle the raw reply
```
`flatten_map` flattens every map reply, which destroys the structure `SearchResult` parses, so there is no honest structured result to return. Documented on `ft_search` alongside the batch limitation.

lib/valkey/commands.rb
Jonathan-Improving: All 7 removed. To answer "do you know if these existed in previous versions" — I checked the module source rather than guessing, and valkey-search registers only:

```
FT._LIST   FT.AGGREGATE   FT.CREATE   FT.DROP   FT.DROPINDEX   FT.INFO   FT.SEARCH
(+ internal FT._DEBUG / FT.INTERNAL_UPDATE)
```

No `FT.ALIASADD`/`ALIASDEL`/`ALIASUPDATE`/`_ALIASLIST`, `FT.EXPLAIN`, `FT.EXPLAINCLI` or `FT.PROFILE` anywhere — they are RediSearch commands that were carried over in the original port, not things that regressed. So they are gone rather than version-gated.

**One more found by extension:** `ft_drop_index` accepted a `dd:` flag, but `FT.DROPINDEX` has **arity 2** (`ft.dropindex.json`, and `ft_dropindex.cc` errors on `argc != 2`), so passing `DD` was producing a wrong-arity error. The parameter is removed and the docs now state documents are always preserved.

Nice side effect: the standalone search suite went from **9 skips to 0**, because those 9 skips were exactly these nonexistent commands. `0 skips` is now the correct assertion there, which makes a module-less server detectable instead of looking like a pass.

I left the `RequestType::FT_*` constants in place, since AGENTS.md requires `RequestType` to mirror the glide-core enum — happy to strip them too if you would rather they go.

test/lint/vector_search_commands.rb
Jonathan-Improving: Done — removed the `if: false` gate on the `module-tests` job (it was commented "Disabled for now until Module support is added", which this PR adds).

**It is green on this PR: Module Tests passed in 51s**, so the job is genuinely exercising the search API against a live module server in CI rather than just locally. All 18 checks pass.

One caveat I want to state rather than let the green tick imply more than it should: that job runs **standalone only** —
```
run: bundle exec ruby -Itest -Ilib test/integration/standalone/commands_test.rb -n "/TestStandalone(Module|Json|VectorSearch)Commands/"
env: STANDALONE_ENDPOINTS: ...
```
There are no `CLUSTER_ENDPOINTS`/`cluster_commands_test` references in the job, so **CI cannot catch a cluster regression** (see the cluster thread). Interestingly the job already tears down module *cluster* containers on 8001-8006 that nothing connects to, so the infrastructure exists and is provisioned every run — wiring it up would need a run step, not new infrastructure. Happy to do that separately if you want it.

test/integration/standalone/commands_test.rb
Jonathan-Improving: Good question, and rather than guess I stood up a 3-primary cluster with the search module and measured it. Short answer: **partially supported**, and the two halves have different owners.

Worth noting first that cluster was *expected-but-untested* rather than out of scope — the builder API mirrors Java GLIDE, ships `shard_scope`/`consistency` for `FT.SEARCH`/`FT.INFO`, and this mixin was pre-written and merely commented out since #219.

**Cause A — index creation was single-node. Client-side, and fixed here.**
After a builder `ft_create`, per-node `FT._LIST` showed the index on **one** of three nodes; the others answered `Index with name ... not found in database 0`. `ft_create`/`ft_drop_index` now broadcast to all primaries via `Route.all_primaries`:
```ruby
# => {"127.0.0.1:44058"=>"OK", "127.0.0.1:10278"=>"OK", "127.0.0.1:32586"=>"OK"}
```
Replies are collapsed back to `"OK"` when every node agrees, so callers see the same value as on standalone, while a genuine per-node disagreement still surfaces as the Hash instead of being hidden. Index creation/removal now works cluster-wide, and the index-creation tests pass on cluster.

**Cause B — cross-shard aggregation is absent on search 1.2.1. Engine-side, not fixable from the client.**
Six documents seeded, distributed by slot:

| Node | `FT.INFO` num_docs | per-node `FT.SEARCH` |
|---|:--:|:--:|
| 10278 | 1 | 1 |
| 32586 | 3 | 3 |
| 44058 | 2 | 2 |

`FT.SEARCH idx hello NOCONTENT ALLSHARDS` run on the node holding **1** document returned **1**, not 6. Via the client, `shard_scope: :all_shards` gives a well-formed `SearchResult` with `total_results=1` of 6. So `ALLSHARDS` is *accepted but does not fan out* on this version — no client change can synthesise an aggregation the module does not perform.

Measured against valkey-search **1.2.1** (`MODULE LIST` ver 66049), built from `main` at `8fc3bbd`.

**Consequence:** 8 of 29 cluster tests fail on result completeness regardless of client behaviour, so I have left the mixin commented out — but with the measurement and a reproduction command recorded in the file, so the next person inherits the evidence instead of rediscovering it:
```
env -u CLUSTER_ENDPOINTS CLUSTER_MODULES=/abs/path/libsearch.dylib \
  bundle exec ruby -Itest -Ilib test/integration/cluster/cluster_commands_test.rb \
    -n "/TestClusterVectorSearchCommands/"
```
(`CLUSTER_ENDPOINTS` must be unset or the harness short-circuits and never loads the module.)

I deliberately did **not** hash-tag the seed keys into a single slot to make it green — that would test standalone behaviour in a cluster costume and convert a measured limitation into an invisible one.

If you know of a newer module build where cross-shard aggregation works, I am happy to re-run — it is one command now.

Jonathan-Improving:
Thanks both — all review comments are addressed and I have replied in each thread with what changed. @yipin-chen the CHANGELOG is updated.

Rebased onto current `main` (picking up #302 and #300), so the merge conflict is resolved and the PR is mergeable again. **All 18 checks pass, including Module Tests** — the job that was gated behind `if: false`, now genuinely running the search suite against a live module server.

Summary of changes since the review:

**Removed things the server does not implement.** Verified against the valkey-search source rather than inferring from error messages: only `FT._LIST`, `FT.AGGREGATE`, `FT.CREATE`, `FT.DROP`, `FT.DROPINDEX`, `FT.INFO`, `FT.SEARCH` are registered. So `FT.ALIAS*`, `FT.EXPLAIN`, `FT.EXPLAINCLI` and `FT.PROFILE` are gone, as is `shard_scope:`/`consistency:` on `ft_aggregate` (those tokens exist only in the ft_info and ft.search parsers). Found by extension: `FT.DROPINDEX` has arity 2, so the `dd:` flag was producing a wrong-arity error — removed.

**Dropped the redis-rb raw-args passthrough** (breaking). The builders are now the only way to pass options; `Valkey#call` is the escape hatch. Collapsing the dual dispatch took `vector_search_commands.rb` from 406 → 287 lines, folding three per-command resolvers into one.

**Failed loudly instead of returning garbage** in three places rather than just documenting them: `NOCONTENT` + `WITHSORTKEYS` is now rejected (the server drops sort keys while the reply is still parsed as pairs); `ft_search` raises a clear `ArgumentError` inside `pipelined`/`multi` and with `flatten_map: true` instead of a `NoMethodError` or a corrupt `SearchResult`; a malformed `limit:` Hash raises the documented `ArgumentError` rather than `KeyError`.

**Cluster mode: partially supported, measured not assumed.** Index creation was single-node — fixed client-side by broadcasting `ft_create`/`ft_drop_index` to all primaries. Cross-shard result aggregation is absent on search 1.2.1 (`ALLSHARDS` is accepted but returns shard-local results only), which no client change can fix, so the cluster suite stays disabled with the measurement and a reproduction command recorded in the file. Details and numbers in that thread.

Two things I did **not** do, and flag deliberately:
- Kept the `RequestType::FT_*` constants for the removed commands, since AGENTS.md requires `RequestType` to mirror the glide-core enum. Happy to strip them if you would prefer.
- Did not wire cluster into the Module Tests job. That job is standalone-only, so **green CI does not cover cluster** — worth knowing when reading the tick. The docker action already provisions module cluster nodes on 8001-8006 that nothing connects to, so closing that gap needs a run step rather than new infrastructure. Happy to do it separately.

Verified: 136 unit tests / 0 failures, rubocop clean, and a live standalone run of 29 tests / 74 assertions / 0 failures / 0 errors / **0 skips** (the 9 previous skips were the nonexistent commands, so 0 skips is now the correct assertion and a module-less server is detectable).

Jonathan-Improving:
Follow-up on test strength — worth flagging because it changes what the green Module Tests tick actually means.

The question was whether the E2E suite genuinely covers the new builder API, given the harness itself predates this PR (`test/lint/vector_search_commands.rb` came in with #65). I checked by mutation testing: injecting deliberate faults into the builders and confirming the suite goes red. **It did not, for 5 of 18 mutants.** All now caught, in 41d4a8e.

Mutants that survived the original suite:

| Injected fault | Why it slipped through |
|---|---|
| `LIMIT` count → `0`; offset → `0` | `assert_operator length, :<=, 2` is satisfied by 0 |
| `SORTBY` clause dropped | `assert_equal prices.sort, prices` passed by luck on two already-ordered docs |
| `NOCONTENT` never emitted | parsed result is identical either way (see below) |
| `VectorField` FLAT↔HNSW swapped; `DIM` dropped | `assert_equal "OK"` + index existence holds for *any* schema |

A silently broken `LIMIT` would truncate every paginated result in production, and CI would have stayed green.

`NOCONTENT` needed a different fix rather than a stronger assertion: `SearchResult.from_raw` receives `no_content:` as a separate argument and discards the field payload client-side, so a builder that never emits the token produces a byte-identical `SearchResult`. No assertion on the parsed result can detect it — that test now asserts the emitted args directly.

**The more substantive finding: every `GROUPBY` test was asserting against an ungrouped result.** `FT.AGGREGATE` requires grouped and reduced fields to be named in `LOAD`; without it the server collapses all documents into a single row instead of erroring. So the reducer assertions were being read off one bogus row, and `SUM(@price)` was returning 0. One of my own test comments had rationalized that as "grouped total is 0 in this engine build" — it was a missing `LOAD`, not an engine quirk. Verified directly:

```
FT.AGGREGATE probe ... GROUPBY 1 @category REDUCE SUM 1 @price AS total
  -> 1 row,  total=0                      # no LOAD
LOAD 1 @category  GROUPBY ...             -> 2 rows, total=0
LOAD 2 @category @price GROUPBY ...       -> electronics=300, books=80   # correct
```

The tests now pass `load:` and assert the per-group breakdown. Note this means `AggregateOptions#load` had **no** real E2E coverage before — dropping the `LOAD` emission entirely used to be invisible; it now fails 6 tests.

Same 29 tests, **74 → 101 assertions**, no library changes: the builders were correct throughout, only the tests were weak. Module Tests still green (44s), 18/18 checks pass.

Two things I am explicitly *not* claiming: this is standalone-only (the cluster gap in the thread above is unchanged), and mutation coverage is not exhaustive — I probed the arg-emitting paths, not every branch.

Jonathan-Improving:
> Conditional approval, pending Alex' final blessing. Please update changelog.

Done, changelog updated in [aa31fa1](https://github.com/valkey-io/valkey-glide-ruby/pull/295/commits/aa31fa17782f83378e59a36ff6f5b32e07653db8)

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -2,13 +2,15 @@
...
+    # full-text search, aggregation, and vector similarity search. They require
+    # the search module to be loaded on the server.
     #
-    # RediSearch provides secondary indexing, full-text search, and vector similarity search
-    # capabilities on top of Redis/Valkey. These commands require the RediSearch module to be loaded.
-    #
-    # @see https://redis.io/docs/stack/search/
+    # Options are always expressed with the {Valkey::Search} builders — either as
+    # an options object or as the equivalent keyword arguments. There is no
+    # raw-token pass-through form; to send FT.* tokens by hand use {Valkey#call}.
```
currantw: Is this comment necessary? It seems to me like this is more a design decision, and is already documented by the API itself? 🤷

lib/valkey/search/aggregate.rb
currantw, line 1: Similar to other comments, I think that the documentation here is unnecessarily verbose. To my eye, much of it duplicates information already available in https://valkey.io/commands/ft.aggregate/. My general preference would be to use very short, top-level descriptions only, and to leave all further documentation to Valkey. This is really only a pass-through to a Valkey server, after all. For example, see `AggregateGroupBy` at https://github.com/valkey-io/valkey-glide-csharp/blob/main/sources/Valkey.Glide/ServerModules/Ft.Aggregate.cs

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
-        command_args = [index, query] + args
-        send_command(RequestType::FT_AGGREGATE, command_args)
+      # Pass an Array of {Valkey::Search::Field} objects and, optionally, a
+      # {Valkey::Search::CreateOptions} (or the same index-level options as
+      # keyword arguments):
+      #
+      #     valkey.ft_create("idx",
+      #       [Valkey::Search::TextField.new("title", sortable: true),
+      #        Valkey::Search::VectorField.hnsw("embedding", dim: 128, metric: :cosine)],
+      #       on: :hash, prefixes: ["doc:"])
```
currantw, line 33: @Aryex is there any standard way to document examples in Ruby that we are using?

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
+      #     valkey.ft_search("idx", "@title:hello", limit: { offset: 0, count: 10 })
+      #
+      # ### Limitations
+      #
+      # * **`flatten_map: true` is not supported.** That client option is a
+      #   redis-rb 4.x compatibility shim that flattens every map reply, which
+      #   destroys the structure {Valkey::Search::SearchResult} parses.
+      # * **Not supported inside `pipelined` / `multi`.** Commands queued in a
+      #   batch return a {Valkey::Future} rather than a reply, so there is nothing
+      #   to parse into a SearchResult.
```
currantw, line 126: @Aryex is this true in Ruby client for _all_ methods that return a structured result? 🤔

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
+      #     result.total_results  # => Integer
+      #     result.documents      # => [#<struct Document key=..., fields={...}>, ...]
+      #
+      #     valkey.ft_search("idx", "@title:hello", limit: { offset: 0, count: 10 })
+      #
+      # ### Limitations
+      #
+      # * **`flatten_map: true` is not supported.** That client option is a
+      #   redis-rb 4.x compatibility shim that flattens every map reply, which
+      #   destroys the structure {Valkey::Search::SearchResult} parses.
```
currantw: It is not clear to me what this comment applies to? Where is the `flatten_map` option coming fromn?

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
+      # arguments, rejecting both at once. Returns nil when neither is given.
+      #
+      # @param klass [Class] the options class
+      # @param options [Object, nil] a positional options instance
+      # @param kwargs [Hash] keyword options
+      # @param label [String] command name, for error messages
+      # @return [Object, nil] an instance of klass, or nil
+      # @raise [ArgumentError] on a wrong-typed positional, or options given both ways
+      def ft_resolve_options(klass, options, kwargs, label:)
+        short = klass.name.split("::").last
```
currantw: Should we include the full class name in error message? Could be a more descriptive error message?

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
+      #     valkey.ft_aggregate("idx", "@price:[0 +inf]",
+      #       clauses: [
+      #         Valkey::Search::GroupBy.new(["@category"],
+      #           reducers: [Valkey::Search::Reducer.count(as: "count")]),
+      #         Valkey::Search::SortBy.new("@count", :desc),
+      #         Valkey::Search::Limit.new(0, 10),
+      #       ], dialect: 2)
+      #
+      # Note that Valkey Search rejects the `"*"` wildcard here — use a filter
+      # expression such as `"@price:[0 +inf]"` as the query.
```
currantw: Is this specific to our client, or something that is true for this command more generally?

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
+      # Resolve an options object from either a positional instance or keyword
+      # arguments, rejecting both at once. Returns nil when neither is given.
+      #
+      # @param klass [Class] the options class
+      # @param options [Object, nil] a positional options instance
+      # @param kwargs [Hash] keyword options
+      # @param label [String] command name, for error messages
+      # @return [Object, nil] an instance of klass, or nil
+      # @raise [ArgumentError] on a wrong-typed positional, or options given both ways
+      def ft_resolve_options(klass, options, kwargs, label:)
```
currantw, line 218: @Aryex Is this a common pattern in the Ruby client? If not, I think this potentially adds more complexity than it is worth. In particular, it means we bypass any validation performed by the options class.

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
+        short = klass.name.split("::").last
+        if options.nil?
+          kwargs.empty? ? nil : klass.new(**kwargs)
```
currantw, line 220: What would be the behaviour here is the user calls this with invalid options? Would this fail, or would it just proceed silently? If this latter, that seems like behaviour that might conceal a lot of bugs 😬

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
-      # @see https://redis.io/commands/ft.explaincli/
-      def ft_explain_cli(index, query, *args)
-        command_args = [index, query] + args
-        send_command(RequestType::FT_EXPLAIN_CLI, command_args)
+      # @raise [ArgumentError] on an empty/invalid schema, a non-CreateOptions
+      #   options object, or options passed both ways
+      def ft_create_args(index, fields, options, kwargs)
+        unless fields.is_a?(Array)
+          raise ArgumentError, "ft_create schema must be an Array of Valkey::Search::Field, got #{fields.class}"
+        end
```
currantw, line 240: Is this necessary? We have already specified that `fields` must be an array, right? Do we typically include additional checks to verify that a parameter conforms to the specified type, or is any failure expected when the user violates the method's contract?

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
+      # every node replied identically (the normal case for FT.CREATE /
+      # FT.DROPINDEX) collapse it back to the single value, so cluster and
+      # standalone callers see the same "OK". A genuine per-node disagreement is
+      # returned as-is rather than hidden.
+      def ft_collapse_broadcast(reply)
+        return reply unless reply.is_a?(Hash) && !reply.empty?
+
+        values = reply.values.uniq
+        values.length == 1 ? values.first : reply
       end
```
currantw, line 267: @Aryex Do we not have a common pattern for this sort of response aggregation in the Ruby client? This is something that is quite common across Valkey commands.

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
-      # @see https://redis.io/commands/ft.profile/
-      def ft_profile(index, query_type, *args)
-        command_args = [index, query_type] + args
-        send_command(RequestType::FT_PROFILE, command_args)
+      # Index definitions are per-node in Valkey Search, so FT.CREATE /
+      # FT.DROPINDEX must reach every primary in cluster mode — a single-node
+      # create leaves other shards answering "Index with name '...' not found".
+      # Returns nil on standalone (and wherever routing is unavailable, e.g. a
+      # queued batch), which leaves the command's default routing untouched.
+      def ft_all_primaries_route
```
currantw, line 276: This also seems like something for which there should be a common pattern or helper method? 🤔

lib/valkey/commands/vector_search_commands.rb
```diff
@@ -18,253 +20,265 @@ module VectorSearchCommands
...
-      def ft_search(index, query, *args)
-        command_args = [index, query] + args
-        send_command(RequestType::FT_SEARCH, command_args)
+      # ft_search parses the reply into a structured result, which is incompatible
+      # with two redis-rb compatibility behaviors: a queued batch yields a
+      # {Valkey::Future} instead of a reply, and `flatten_map: true` flattens the
+      # map structure the parser depends on. Fail with a clear ArgumentError
+      # instead of returning a corrupt result (or a NoMethodError from deep inside
+      # the parser).
+      def ft_assert_supported!(method_name)
```
currantw: Do we not already have a standard way to handle this in the Ruby client, if this is a redis-rb compatibility issue? (@Aryex)

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+    # A single REDUCE within a GROUPBY. Emits `REDUCE <function> <nargs> <arg...>
+    # [AS <name>]`. The nargs count is the number of argument tokens that follow
+    # the function name, matching the wire format and Java.
+    #
+    # @example
+    #   Valkey::Search::Reducer.count(as: "total")
+    #   Valkey::Search::Reducer.sum("@price", as: "revenue")
+    class Reducer
+      # @return [String] the reducer function (e.g. "COUNT", "SUM")
+      attr_reader :function
```
currantw, line 21: Should we define an enum for the allowed values here?

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+    # Filter, Limit, Apply) serialize in the order the user supplies them, since
+    # FT.AGGREGATE is an ordered pipeline. Flat top-level options (verbatim, load,
+    # params, dialect, timeout) are keyword arguments on {AggregateOptions}.
+    # Mirrors the Java FTAggregateOptions surface, adapted to Ruby idiom.
+    #
+    # @see https://valkey.io/commands/ft.aggregate/
+
+    # A single REDUCE within a GROUPBY. Emits `REDUCE <function> <nargs> <arg...>
+    # [AS <name>]`. The nargs count is the number of argument tokens that follow
+    # the function name, matching the wire format and Java.
```
currantw: I don't think we should reference other clients. @Aryex let me know if there is any different guidance. In general, I think we shouldn't expect the user to know anything about the other clients. 🤷

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+      def initialize(function, *args, as: nil)
+        @function = function.to_s.upcase
+        @args = args
+        @alias_name = as
+      end
+
+      # COUNT — number of records in the group. Takes no property.
+      def self.count(as: nil)
+        new("COUNT", as: as)
+      end
```
currantw, line 39: As enum for functions (see C# for an example) would eliminate the need for all these separate builder methods.

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+      def self.stddev(property, as: nil)
+        new("STDDEV", property, as: as)
+      end
+
+      # @return [Array] `REDUCE <function> <nargs> <arg...> [AS <name>]`
+      def to_args
+        tokens = ["REDUCE", @function, @args.length, *@args]
+        tokens.push("AS", @alias_name) unless @alias_name.nil?
+        tokens
+      end
```
currantw, line 57: It seems to me like this could be an internal method. Do we have any established pattern for this, such as prefixing the method with `_` or labelled it with `@api private`?

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+    #   Valkey::Search::GroupBy.new(["@category"],
+    #     reducers: [Valkey::Search::Reducer.count(as: "total")])
+    class GroupBy < AggregateClause
+      # @param properties [Array<String>] grouping properties (may be empty for a
+      #   global aggregation, matching `GROUPBY 0`)
+      # @param reducers [Array<Reducer>] reducers applied to each group
+      def initialize(properties = [], reducers: [])
+        super()
+        @properties = Array(properties)
+        @reducers = reducers
```
currantw: Why do we call `Array()` here for `properties` but not for `reducers`? Aren't both already passed to this method as `Array`s?

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+    # wire format and Java.
+    #
+    # @example
+    #   Valkey::Search::SortBy.new({ "@total" => :desc }, max: 10)
+    #   Valkey::Search::SortBy.new("@total", :desc)
+    class SortBy < AggregateClause
+      # Accepts either a single `(property, order)` pair or an ordered Hash /
+      # Array-of-pairs of `property => order`.
+      #
+      # @param keys [String, Hash, Array] a property, or `{ property => order }`
```
currantw, line 109: I don't think a hash works here, because the _order_ is important, no (i.e. because hashes have no order)?

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+        @pairs = normalize_keys(keys, order)
+        @max = max
+      end
+
+      def to_args
+        flat = @pairs.flat_map do |property, ord|
+          [property, Search.lookup_token(Search::SORT_ORDERS, ord, "sort order")]
+        end
+        tokens = ["SORTBY", flat.length, *flat]
+        tokens.push("MAX", @max) unless @max.nil?
```
currantw, line 123: Does this client have a file that defines all the constants like this? If not, should it?

lib/valkey/commands/vector_search_commands.rb
Aryex, line 1: I believe other clients don't support FT search at all in Batch? Can you confirm? If so we should also disable it in pipeline/multi for now. 

Just want to keep the scope for this tight. Support for batch can be implemented in a later date.

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+    end
+
+    # APPLY clause: `APPLY <expression> AS <name>`. Adds a computed field to each
+    # row.
+    #
+    # @example
+    #   Valkey::Search::Apply.new("@price * @qty", as: "line_total")
+    class Apply < AggregateClause
+      # @param expression [String] the projection expression
+      # @param as [String] result field name (required by the wire format)
```
currantw: I don 't think we should use the term "wire format". To me, it is confusing and unnecessarily technical. 🤷

lib/valkey/search/aggregate.rb
```diff
@@ -0,0 +1,298 @@
...
+    # pipeline clauses are emitted first (in supplied order), followed by the flat
+    # top-level options, matching Java's builder assembly.
+    #
+    # Options for FT.AGGREGATE, serialized after `index query`. The flat
+    # top-level options are emitted first, then the ordered pipeline clauses (in
+    # supplied order).
+    #
+    # Note: Java emits DIALECT after the pipeline clauses; this builder emits it
+    # before them. The server is order-independent for these tokens, so the
+    # divergence is wire-benign — kept for a simpler assembly.
```
currantw: As mentioned in another comment, I think that documentation like this that references other clients isn't useful.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Checked before agreeing to narrow scope, rather than assuming: Java has no FT.* support at all in this codebase; Python's `ft.*` module functions take a client directly (`create(client, ...)`), and `Batch`/`Transaction` defines no `ft_*` methods; C#'s `Ft.Aggregate` has no `IBatch`/`ITransaction` overload either.

Confirmed — none of them support FT.* in a batch. `ft_aggregate` now raises the same `ArgumentError` as `ft_search` inside `pipelined`/`multi`, in [0d6048d](https://github.com/valkey-io/valkey-glide-ruby/pull/295/commits/0d6048dcf812bb1954b1d75586082c321ce23ab3). Agreed on keeping scope tight — batch support for FT.* can follow once the client's own GLIDE-batch story lands.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Fair — removed in 0d6048d. It restated a design decision the method docs already make evident (`options = nil, **kwargs`, no raw-token param anywhere in the signatures).

lib/valkey/search/aggregate.rb
Jonathan-Improving: Trimmed the Java/C# references and a stale duplicate paragraph in 0d6048d. Kept the short structural description (clause order, flat-options-first) since that is a Ruby-specific serialization detail the linked Valkey docs do not cover, but agreed the cross-client comparison does not belong here.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: No special convention beyond standard YARD `@example` — checked the rest of the client (`connection_commands.rb`, `function_commands.rb`) and that is what is already used there, so this file matches house style rather than inventing its own.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: That's a real gap — thanks for catching it. `flatten_map:` is not a search-builder option at all; it is `Valkey.new(..., flatten_map: true)`, a client-wide constructor option defined in `lib/valkey.rb`, unrelated to anything in this file. Clarified the doc to say so explicitly in 0d6048d, since there was nothing on this page that would tell you where it comes from.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Reasonable ask, but this one is intentionally short: `label` is already the command name (e.g. `"ft_search"`), so the message reads as `"ft_search options must be a Valkey::Search::SearchOptions, got String"` the full class name is already in there via `#{klass}`. Let me know if you had a different expansion in mind and I will take another pass.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Correction: my earlier reply on this thread answered the wrong question entirely (it talked about NOCONTENT/WITHSORTKEYS, an unrelated part of this file). Sorry about that, here's the actual answer.

To your real question: server-side, not specific to this client. There is no client-side check anywhere in this file for `"*"`, the query string is passed straight through to FT.AGGREGATE. But checking it properly turned up something more important: the comment was wrong. Measured directly against a live server (valkey-search 1.2.1), with the exact GROUPBY/REDUCE/LOAD shape this test suite actually uses:

```
FT.AGGREGATE idx "*" LOAD 1 @category GROUPBY 1 @category REDUCE COUNT 0 AS count
=> 2 rows: electronics/count=2, books/count=1   (matches every document, same as a real filter)
```

`"*"` is accepted and works identically to an equivalent filter expression. It is not rejected. My best guess at how the false claim got in: an earlier no-`LOAD` test of `"*"` returned a bare row with no fields and no error, which looks like "rejected" until you compare it against a real filter expression under the same no-`LOAD` conditions, which returns the identical bare row. Neither is a rejection, `FT.AGGREGATE` just doesn't project fields unless `LOAD`/`APPLY` names them.

Removed the false claim from the doc comment and the two lint-suite comments repeating it, and added `test_ft_aggregate_builder_wildcard_query_matches_all` in b35c657, which asserts `"*"` produces identical grouped output to a filter matching the same documents, so a regression here would actually be caught rather than resting on an untested comment.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: To answer directly: this is *not* bypassing the options class' validation, `klass.new(**kwargs)` on this line calls the exact same constructor as the object form, e.g. `SearchOptions.new(**kwargs)`. That constructor is where validation actually lives (see `SearchOptions#initialize`, which raises on `with_sort_keys && no_content` at construction time). So kwargs and an explicit options object go through identical validation; this line only decides which path constructs the instance, not whether it gets validated.

I do not think this is a common Ruby-client pattern elsewhere in the repo beyond this file, if that was the concern, I can point at it differently if the dual object-or-kwargs form itself is what feels like too much surface, but the validation is intact either way.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: To close the loop on this given the clarification above: invalid options, whether passed as kwargs or as an object, raise `ArgumentError` at construction, before any command is sent. Nothing proceeds silently; there is no path where a malformed option set reaches the server. Let me know if you want the resolver itself to short-circuit earlier for a specific case and I can take a look.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: It is intentionally a second check: `fields.is_a?(Array)` on its own would let a non-Array happily flow into `fields.each` and blow up with a `NoMethodError` deep inside FT.CREATE construction. This raises `ArgumentError` up front with a message naming exactly what was wrong (empty schema vs. wrong element type), rather than a generic exception from whatever iterated over it first. Whether that is "typical" here — the client does raise `ArgumentError` for other malformed builder inputs throughout this file (see `ft_resolve_options`), so this keeps the same contract for the schema argument specifically.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: I checked and there is no existing helper for this in the client. `ft_collapse_broadcast` looks to be the first place a per-node broadcast reply gets collapsed back to a single value. It exists specifically because `ft_create`/`ft_drop_index` now broadcast to all primaries (index definitions are per-node in Valkey Search), so glide-core returns `{ "host:port" => reply }` even in the common case where every node agreed. If there is a more general pattern elsewhere in the client I should be reusing instead, point me at it and I will switch, I did not find one.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Same answer as the collapse-broadcast comment above: checked, did not find an existing helper for per-node routing in cluster mode (`ft_all_primaries_route`). If `@cloud_mode`/`Route` handling has a common home elsewhere that this should be calling into, let me know and I will consolidate.  This was written as the narrowest thing that made `ft_create`/`ft_drop_index` broadcast correctly.

lib/valkey/search/aggregate.rb
Jonathan-Improving: Removed the C# reference — agreed a Ruby user should not need to know anything about other clients to read this. Kept the short rationale for why the design differs (dialect-order divergence is wire-benign) since that stands on its own without the comparison.

lib/valkey/search/aggregate.rb
Jonathan-Improving: That's a fair ask given `SORT_ORDERS` / `SHARD_SCOPES` / `CONSISTENCY` already exist as lookup tables in `lib/valkey/search.rb` — `sort_order: :asc`/`:desc` on `SortBy` goes through exactly that lookup (`Search.lookup_token(Search::SORT_ORDERS, ord, ...)`), so it already has the same validation and error message an enum would give you.

I read the original comment as being about the reducer *function* names (COUNT/SUM/MIN/...) instead — those are intentionally open (`Reducer.new(function, *args)`) so a future reducer the server adds doesn't need a client release to use via the raw form, though the common ones each have a class method (`.count`, `.sum`, `.min`, etc.) that reads like an enum member without literally being one.

Let me know if that's not what you meant and I'll take another look.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Checked. There is no existing client-wide pattern for "this command is unsupported inside a batch" to point to. `transaction_commands.rb` reads `@in_multi` too, for a different purpose though: preventing nested `MULTI`, not rejecting a specific command inside one. `ft_queued_in_batch?`/`ft_assert_supported!` looks to be the first case of a command family needing to say "not supported here" rather than just letting the batch queue it and fail later (or, for `ft_search`, silently return a broken result).

If a general mechanism for this gets added later, `ft_search`/`ft_aggregate` would be natural first callers of it. I don't think it's worth inventing a new client-wide abstraction from a single case right now though. Happy to reconsider if you think otherwise.

lib/valkey/search/aggregate.rb
Jonathan-Improving: Checked across the whole client rather than just this file. Both conventions exist, though the underscore prefix is the minority: `list_commands.rb` and `sorted_set_commands.rb` use `_bpop`/`_zsets_operation`-style names, while `connection_commands.rb`, `generic_commands.rb`, `scripting_commands.rb`, `server_commands.rb`, and `stream_commands.rb` all have plain-named `private` helpers (`build_client_kill_args`, `flatten_call_args`, `split_keys_and_args`, `parse_info_reply`, ...) with no prefix.

So `ft_resolve_options`/`ft_create_args`/etc. under `private` here actually matches the more common pattern in this codebase. No `@api private` YARD tag anywhere either. I don't see a real inconsistency to fix, but happy to rename if you'd prefer this file specifically follow the underscore style.

lib/valkey/search/aggregate.rb
Jonathan-Improving: Took another look at this given the follow-up. It's a fair point separate from the SORT_ORDERS-style lookup I mentioned before: the real ask here is fewer hand-written builder methods, beyond just validation.

Added `Search::REDUCER_FUNCTIONS` in the commit just pushed and generate `Reducer.count_distinct`/`.sum`/`.min`/`.max`/`.avg`/`.stddev` from it via `define_singleton_method`, so there's one list instead of seven near-identical method bodies that could drift out of sync. Kept `Reducer.new(function, *args)` itself open rather than validating against a closed enum, so a server-added reducer the client doesn't know about yet still works via the raw form; it just won't have its own `.method` shortcut until the table is updated. That seemed like the right tradeoff given how easy it is for a module to add a new REDUCE function between client releases.

lib/valkey/search/aggregate.rb
Jonathan-Improving: Addressed in the commit just pushed, see the reply on the sibling comment below. `Search::REDUCER_FUNCTIONS` is now the single source of truth for `Reducer.count`/`.sum`/`.min`/etc., generated via `define_singleton_method` instead of seven hand-written methods.

lib/valkey/search/aggregate.rb
Jonathan-Improving: Ruby Hash has guaranteed insertion order since 1.9 (2009). This is not an implementation detail that happens to hold today, it is documented language behavior: https://docs.ruby-lang.org/en/3.0/Hash.html states "Hashes enumerate their values in the order that the corresponding keys were inserted."

I don't think there is a real risk here, so no code change.

lib/valkey/search/aggregate.rb
Jonathan-Improving: Checked, and yes: this is `lib/valkey/search.rb`, and it already exists for exactly this purpose, following the same pattern as `lib/valkey/read_from.rb`, `lib/valkey/request_type.rb`, and `lib/valkey/response_type.rb`, one small dedicated file per family of wire-value constants, rather than one shared constants file for the whole client.

It currently holds `SORT_ORDERS`, `SHARD_SCOPES`, `CONSISTENCY`, `DIALECTS`, and now `REDUCER_FUNCTIONS` (added in 92cbb13 for the enum-for-reducers request above), plus the shared `lookup_token`/`normalize_dialect`/`params_tokens`/`append_cluster_flags` helpers that use them. So this file is the answer to "should there be one" for the search builders specifically; there's no separate client-wide constants file, and based on the existing `read_from.rb`/`request_type.rb` precedent I don't think there should be, keeping each command family's constants next to that family's code seems to be the established convention here.

lib/valkey/search/aggregate.rb
Jonathan-Improving: Went with `@api private` in 39c9953, tagging all six private helpers (`ft_resolve_options`, `ft_create_args`, `ft_collapse_broadcast`, `ft_all_primaries_route`, `ft_assert_supported!`, `ft_queued_in_batch?`). That tag already exists elsewhere in the client (`lib/valkey/future.rb`, `lib/valkey/pipeline.rb`, `lib/valkey/opentelemetry.rb`, and `lib/valkey/search/fields.rb` in this same PR), so this brings the file in line with the established YARD convention rather than defending the plain-name status quo.

lib/valkey/search/aggregate.rb
Jonathan-Improving: Follow-up on my earlier reply: went further per additional feedback and actually consolidated. `DATA_TYPES` (was in `create_options.rb`), `DISTANCE_METRICS`/`VECTOR_TYPES` (were in `fields.rb`), and `SCOPES` (was in `info.rb`) are now all in `lib/valkey/search.rb` alongside `SORT_ORDERS`/`SHARD_SCOPES`/`CONSISTENCY`/`DIALECTS`/`REDUCER_FUNCTIONS`, in bdc7b15.

So every enum-like wire-value lookup for this feature now lives in one file instead of five. Each class's `normalize_*` validation method stayed put since that's behavior, not data, and now references `Search::DATA_TYPES` etc. No behavior change, confirmed by direct comparison before committing.

lib/valkey/commands/vector_search_commands.rb
Aryex: Nothing exists currently.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Checked, since this hadn't been answered yet. No other command family in this client parses its reply into a custom structured class. `Valkey::Search::SearchResult`/`Document` are the only such objects anywhere in the client; everything else (cluster commands, XINFO, CLIENT INFO, etc.) returns Ruby's native Hash/Array/String/Integer straight from the reply.

So this restriction is specific to `ft_search`, not a general rule about structured-result methods, because there is nothing else in the client that fits that description to generalize from. `ft_aggregate` sits differently: it returns the raw reply unparsed (an Array, not a custom class), so it only needed the batch guard, not the flatten_map one, added in 0d6048d after a separate scope question from Aryex on this same PR.

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Checked further rather than just re-stating the earlier search. `Valkey::Route` itself documents the exact shape this deals with (multi-node routes return `Hash` of `"host:port" => value`, per the comment in `route.rb`), but every other reference to it in the client is in a doc example only (`dbsize`, `ping`, `info` in comments), not in an actual command implementation. `ft_create`/`ft_drop_index` look to be the first commands that actually call a multi-node route in real code, so there is nothing to consolidate into. `ft_collapse_broadcast` is genuinely new territory rather than a duplicate of something already there.

On `@cluster_mode` specifically: found a real inconsistency while checking this, and fixed it rather than leaving it. `generic_commands.rb` reads `@cluster_mode` directly (`if @cluster_mode`), with no `instance_variable_defined?`/`instance_variable_get` guard. My version in `ft_all_primaries_route` added that guard, but it turns out to be unnecessary: reading an undefined instance variable in Ruby returns `nil` rather than raising, so the plain form degrades safely too, including on a `Pipeline` (which never sets `@cluster_mode` at all). Simplified to match the existing pattern in [6b3e4c9](https://github.com/valkey-io/valkey-glide-ruby/pull/295/commits/6b3e4c9c).

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Checked further rather than just re-stating the earlier search. `Valkey::Route` itself documents the exact shape this deals with (multi-node routes return `Hash` of `"host:port" => value`, per the comment in `route.rb`), but every other reference to it in the client is in a doc example only (`dbsize`, `ping`, `info` in comments), not in an actual command implementation. `ft_create`/`ft_drop_index` look to be the first commands that actually call a multi-node route in real code, so there is nothing to consolidate into. `ft_collapse_broadcast` is genuinely new territory rather than a duplicate of something already there.

On `@cluster_mode` specifically: found a real inconsistency while checking this, and fixed it rather than leaving it. `generic_commands.rb` reads `@cluster_mode` directly (`if @cluster_mode`), with no `instance_variable_defined?`/`instance_variable_get` guard. My version in `ft_all_primaries_route` added that guard, but it turns out to be unnecessary: reading an undefined instance variable in Ruby returns `nil` rather than raising, so the plain form degrades safely too, including on a `Pipeline` (which never sets `@cluster_mode` at all). Simplified to match the existing pattern in [6b3e4c9](https://github.com/valkey-io/valkey-glide-ruby/pull/295/commits/6b3e4c9c).

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Done in 85deff1. Both `ArgumentError` messages in `ft_resolve_options` now use the full class name; one already did (`#{klass}`), the other used a shortened form (`klass.name.split("::").last`). That's gone now, along with the now-unused local.

Before: `pass options to ft_search either as a SearchOptions object or as keyword arguments, not both`
After: `pass options to ft_search either as a Valkey::Search::SearchOptions object or as keyword arguments, not both`

lib/valkey/commands/vector_search_commands.rb
Jonathan-Improving: Checked this properly rather than just defending the existing code. You were right that it's not a common pattern here: I looked across the whole client and no other command with structured options (`client_kill`, `client_tracking`, `zadd`, ...) accepts an equivalent pre-built options object, they all take kwargs only. So the dual object-or-kwargs form is genuinely specific to the FT.* builders.

Documented why the deviation is intentional in 1cc779d, rather than leaving it unexplained: `AggregateOptions` already needs composable clause objects (`GroupBy`, `SortBy`, `Filter`, `Apply`, `Limit`) regardless of how the top-level options are passed, so letting the options themselves also be a pre-built object is consistent with that, and lets a caller build one options instance and reuse it across calls. Kept the form itself rather than removing it: it's used across 78 call sites in the existing test suite, so dropping it would be a real breaking change disproportionate to what started as a documentation gap.

On the validation question specifically, since that's what this thread's follow-up asked: nothing changes there. Both paths construct the exact same options class (`klass.new(**kwargs)` calls the identical constructor the object form would), so kwargs never bypass whatever the class validates in `#initialize`. That part of my earlier reply holds; what was missing was acknowledging the pattern itself is new here, not something already established.

lib/valkey/commands/vector_search_commands.rb
Aryex: Ruby is a sync client. That Future is used only in batch APIs, a hold over from the reference redis-rb client. Most APIs return single values.

Aryex (approved):

PR #299, opened 2026-08-27
Enhancing multi/pipeline errors but defaults to redis-rb for parity

### LINKED ISSUE #260: `pipelined` / `multi` drop the entire reply array when any queued command errors
### Summary

When a command queued inside `pipelined { ... }` or `multi { ... }` errors at execution time, the client raises `Valkey::CommandError` instead of returning the reply array with the error embedded at the failing index. All other replies from the same batch are unreachable, even though the server did apply the successful writes.

This breaks per-command retry / idempotency logic and diverges from redis-rb 5.x semantics.

### Expected behavior

`pipelined` and `multi` return an `Array` when the batch executes, with `Valkey::CommandError` instances placed at the positions of failing commands. Successful commands' replies remain accessible. This matches Valkey/Redis server-side semantics: runtime errors inside `EXEC` do **not** roll back other commands; `EXEC` returns the per-command reply array.

Concretely:

```ruby
result = client.pipelined do |p|
  p.set("z", "1")
  p.incr("notint")   # server rejects: value is not an integer
  p.get("z")
end

result[0]              # => "OK"
result[1].class        # => Valkey::CommandError
result[1].message      # => "value is not an integer or out of range"
result[2]              # => "1"
```

`multi { ... }` follows the same shape.

### Actual behavior

The block raises `Valkey::CommandError` at the failing queued command; the block's return value is never produced. The caller cannot tell:
- which commands committed (breaks idempotency, retrying the whole batch double-writes the ones that succeeded on the first attempt)
- whether the failure was a queue-time `EXECABORT` (nothing applied, safe to retry) or a runtime error mid-`EXEC` (partial commit, unsafe to retry blindly)

Related observation, likely the same fix path: `incrbyfloat` and `geodist` return `String` from imperative `multi/exec` but `Float` when called directly, indicating type decoding also runs through a different path inside a batch reply.

### Reproduce

```ruby
require "valkey"

client = Valkey.new(host: "127.0.0.1", port: 6379)
client.flushall
client.set("notint", "hello")     # will make INCR fail at EXEC

# --- pipelined ---
begin
  result = client.pipelined do |p|
    p.set("z", "1")
    p.incr("notint")     # runtime error mid-EXEC
    p.get("z")
  end
  puts "pipelined returned: #{result.inspect}"
rescue => e
  puts "pipelined RAISED: #{e.class}: #{e.message}"
end
puts "z after pipelined: #{client.get('z').inspect}   # SET was still applied server-side"

# --- multi ---
client.set("z", "before")
begin
  result = client.multi do |m|
    m.set("z", "1")
    m.incr("notint")
  end
  puts "multi returned: #{result.inspect}"
rescue => e
  puts "multi RAISED: #{e.class}: #{e.message}"
end
puts "z after multi: #{client.get('z').inspect}       # SET was still applied server-side"
```

Output on rc5:

```
pipelined RAISED: Valkey::CommandError: value is not an integer or out of range
z after pipelined: "1"   # SET was still applied server-side
multi RAISED: Valkey::CommandError: value is not an integer or out of range
z after multi: "1"       # SET was still applied server-side
```

Neither block returned. Follow-up `client.get("z")` confirms the SET was applied, proving the per-command reply information exists on the wire but is discarded by the client.

Reproduces the same way on a cluster client using a hashtag to keep the keys in one slot:

```ruby
cluster = Valkey.new(
  nodes: (7000..7002).map { |p| { host: "127.0.0.1", port: p } },
  cluster_mode: true,
)
cluster.set("{tag}:notint", "hello")

begin
  result = cluster.pipelined do |p|
    p.set("{tag}:z", "1")
    p.incr("{tag}:notint")
    p.get("{tag}:z")
  end
  puts "cluster pipelined returned: #{result.inspect}"
rescue => e
  puts "cluster pipelined RAISED: #{e.class}: #{e.message}"
end
```

### Acceptance criteria

- `pipelined { ... }` returns an `Array` when the batch runs, with `Valkey::CommandError` instances at the positions of failing commands.
- `multi { ... }` follows the same shape.
- Type coercion (`Float` for `incrbyfloat` / `geodist`, `Integer` for counter commands, etc.) is consistent between imperative-exec, block-form, and direct-call code paths.
- Existing behavior on queue-time `EXECABORT` (server-side abort, nothing applied) may continue to raise, but the exception class should be distinguishable from mid-`EXEC` runtime errors so callers can differentiate.
- Works on standalone and cluster.

---
_Filed from a blackbox bug bash of valkey-glide-rb 1.0.0.pre.rc5._

Sasidharan3094:
### Summary

`pipelined` and the block form of `multi` raised `Valkey::CommandError` and discarded the entire
reply array whenever a queued command failed at runtime (e.g. `INCR` on a non-integer) - even
though the server had already committed the other queued commands. This broke per-command
retry/idempotency logic, since callers had no way to see which commands succeeded.

This PR adds an opt-in `exception: false` mode to both `pipelined` and `multi` that returns the
reply array with a `Valkey::CommandError` embedded at the failing command's slot instead of
raising, matching the server's "no rollback on a runtime error" semantics. The **default**
behavior of both methods is left as `exception: true` (raise, array discarded) to stay in parity
with redis-rb 5.x, which was verified to behave identically to valkey-glide-ruby's pre-fix
default for both `pipelined` and `multi` (redis-rb's `multi` has no opt-out at all).

Also fixes two related bugs found while investigating:
- `incrbyfloat` / `geodist` returned a `String` instead of a `Float` when called inside
  `multi`/`exec`'s imperative (non-block) form, unlike a direct call.
- A runtime error embedded in that imperative form's `EXEC` reply could be silently coerced to
  `true` by a boolean-typed command's reply reconversion (`Utils::Boolify`), hiding the error.

### Issue link

Closes #260

### Features / Changes

- `pipelined`'s existing `exception:` kwarg is unchanged in shape; `multi`'s block form gains a
  matching `exception:` kwarg (default `true`) - a valkey-glide-ruby-specific extension, since
  redis-rb's `multi` has no such option.
- `exception: false` on either method returns the reply Array with `Valkey::CommandError`
  embedded at the failing command's position; other commands' replies (and their `Future`s) stay
  reachable.
- Added `Valkey::ExecAbortError < Valkey::CommandError`, raised specifically for a genuine
  queue-time abort (`RequestErrorType::EXECABORT`) - so callers can distinguish "nothing ran, safe
  to retry" from a mid-EXEC runtime error, matching the Python/Node/Java/Go GLIDE clients.
- `multi`'s `exception:` kwarg only affects the block form; passing it to the imperative
  `multi` (no block) / `#exec` pair now raises `ArgumentError` instead of silently being ignored,
  since that path dispatches a single literal `EXEC` command with no equivalent raise/embed
  control (redis-rb has no imperative `multi`/`exec` form at all to compare against - calling
  `multi` without a block raises `LocalJumpError` there).
- `incrbyfloat` (`string_commands.rb`) and `geodist` (`geo_commands.rb`) now apply
  `Utils::Floatify`, matching the existing convention used by `hincrbyfloat`/`zincrby`/`zscore`.
- `reconvert_queued_replies` (imperative `exec`'s reply reconversion) now skips re-coercing an
  already-embedded `CommandError`, mirroring the existing guard in `send_batch_commands`.

### Testing

- Reproduced the reported behavior against unmodified `main` on a local standalone server
  (`127.0.0.1:6379`): both `pipelined` and `multi` raised and discarded the array, while a
  follow-up `GET` confirmed the successful command had committed server-side.
- Cross-checked against real `redis-rb` 5.4.1 (same server, same scenario) to confirm this fix's
  defaults match redis-rb's own `pipelined`/`multi` defaults exactly, with `exception: false` as
  the only divergence (an explicit, deliberate opt-in enhancement).
- Added/updated tests:
  - `test/lint/string_commands.rb` - default (`exception: true`) raises; `exception: false`
    returns the array with the error embedded.
  - `test/lint/transaction_commands.rb` - same coverage for `multi`'s block form, plus float
    coercion parity between `multi`/`exec` (block and imperative) and direct calls, plus
    `multi(exception: ...)` without a block raising `ArgumentError`.
  - `test/unit/errors_test.rb` (new) - `ExecAbortError` class hierarchy.
- `bundle exec rubocop lib test` - clean.
- `bundle exec rake test:unit` - 123 tests, 0 failures.
- Targeted `test/integration/standalone/commands_test.rb` run (multi/pipelined/geodist/incrbyfloat
  filter) - 49 tests, 0 failures, 5 unrelated skips.

### Checklist


jamesx-improving (approved):
Reviewed the multi/pipelined error-handling changes, the ExecAbortError split, and the incrbyfloat/geodist float-coercion fixes — the CommandError-embedding guards are consistent across both the batch and imperative-exec paths, and ExecAbortError < CommandError keeps all existing rescues working. LGTM.

lib/valkey/commands/geo_commands.rb
```diff
@@ -47,9 +47,9 @@ def geohash(key, *member)
       # @param [String] key
       # @param [Array<String>] members
       # @param ['m', 'km', 'mi', 'ft'] unit
-      # @return [String, nil] returns distance in specified unit if both members present, nil otherwise.
+      # @return [Float, nil] returns distance in specified unit if both members present, nil otherwise.
       def geodist(key, member1, member2, unit = 'm')
-        send_command(RequestType::GEO_DIST, [key, member1, member2, unit])
+        send_command(RequestType::GEO_DIST, [key, member1, member2, unit], &Utils::Floatify)
```
Aryex, line 52: This would be a breaking change. Is there a strong need for this? I believe we don't have an explicit requirement for redis-rb compatibility.

lib/valkey/commands/string_commands.rb
```diff
@@ -67,7 +67,7 @@ def incrby(key, increment)
       # @param [Float] increment
       # @return [Float] value after incrementing it
       def incrbyfloat(key, increment)
-        send_command(RequestType::INCR_BY_FLOAT, [key, increment])
+        send_command(RequestType::INCR_BY_FLOAT, [key, increment], &Utils::Floatify)
```
Aryex, line 70: Same as [above](https://github.com/valkey-io/valkey-glide-ruby/pull/299/changes#r3908363563),

lib/valkey/commands/transaction_commands.rb
```diff
@@ -31,30 +36,61 @@ module TransactionCommands
...
+      # @param exception [Boolean] see above; a valkey-glide-ruby-specific
+      #   extension - redis-rb's `multi` has no equivalent kwarg.
       #
       # @return [Array<...>]
       #   - an array with replies
       #
       # @see #watch
       # @see #unwatch
-      def multi
+      def multi(exception: NO_EXCEPTION_KWARG)
```
Aryex: Why not use `exception: true`?

test/unit/errors_test.rb
Aryex, line 1: I don't think these tests add any value. Instead can we add at least 1 integration tests that throws `ExecAbortError`?

test/lint/transaction_commands.rb
```diff
@@ -15,6 +15,17 @@ def test_multi_discard
...
 
+    def test_multi_exception_kwarg_without_a_block_raises
+      # In cluster mode, MULTI/EXEC transactions require all keys in same slot
+      # and behave differently with connection routing
+      skip("MULTI/EXEC not supported in cluster mode") if cluster_mode?
+
+      # exception: only applies to the block form; imperative multi/#exec
+      # has no equivalent control.
+      assert_raises(ArgumentError) { r.multi(exception: false) }
+      assert_raises(ArgumentError) { r.multi(exception: true) }
```
Aryex: These types of tests are better as unit tests. It may need a lot of mocking/stubs tho.

test/lint/transaction_commands.rb
```diff
@@ -625,6 +652,26 @@ def test_multi_future_is_aborted_when_batch_raises_a_command_error
       assert_raises(Valkey::FutureAborted) { future.value }
     end
 
+    def test_multi_future_resolves_to_command_error_with_exception_false
+      skip("MULTI/EXEC not supported in cluster mode") if cluster_mode?
+
+      # exception: false leaves the CommandError in place instead of
+      # raising - other queued commands' replies stay reachable.
```
Aryex: Not needed

test/lint/transaction_commands.rb
```diff
@@ -654,5 +701,38 @@ def test_multi_future_reflects_boolean_coercion
       assert_equal false, setnx_future.value
       assert_equal false, hexists_future.value
     end
+
+    def test_multi_matches_pipelined_reproduce_from_gh_issue
```
Aryex: Let's have a better test name pls

test/lint/transaction_commands.rb
```diff
@@ -607,11 +636,9 @@ def test_multi_future_is_aborted_when_block_raises
...
       skip("MULTI/EXEC not supported in cluster mode") if cluster_mode?
 
-      # multi's block form hardcodes exception: true, so a real per-command
-      # error (as opposed to the user's block raising Ruby-level) makes the
-      # whole batch call raise before resolve_futures! ever runs - same
-      # abort path as test_multi_future_is_aborted_when_block_raises, just
-      # triggered from send_batch_commands itself instead of user code.
+      # A real per-command error (unlike the user's block raising
+      # Ruby-level, see test_multi_future_is_aborted_when_block_raises)
+      # raises via multi's default exception: true.
```
Aryex: Let's remove this comment.

test/lint/string_commands.rb
```diff
@@ -313,6 +313,35 @@ def test_pipelined_future_raises_command_error_for_inline_error_slot
       assert_equal 'not_a_list', result[1]
     end
 
+    def test_pipelined_raises_on_inline_error_by_default
```
Aryex, line 316: missing pipelined(exception: true) tests. Can you add them as well?

lib/valkey.rb
```diff
@@ -24,6 +24,13 @@ class Valkey
   include Commands
   include PubSubCallback
 
+  # @param exception [Boolean] when `true` (the default, matching redis-rb),
```
Aryex: Can we remove mentions of redis-rb?

Sasidharan3094:
@Aryex I will address the review comments tomorrow

lib/valkey/commands/geo_commands.rb
Sasidharan3094: good catch but i don't think this is actually breaking tbh.
geodist/incrbyfloat already return Float today when called directly - glide-core auto coerces it for single commands, same way it does for boolean commands like EXPIRE/SISMEMBER. so Floatify on an already-Float value is just a no-op, verified this on a live server before and after the change.
the real inconsistency is only inside multi/exec's imperative form (no block) - that sends a literal EXEC command outside the batch path so it never gets that same coercion, comes back as raw String instead. so today incrbyfloat gives you Float when called directly but String when wrapped in multi/exec, same command. thats what am fixing here.
not about redis-rb at all, my bad if the framing gave that impression. just making the client keep its own promise (Float) consistent no matter how you call it.
happy to add a test showing direct call vs multi/exec giving the same type if that helps confirm it

lib/valkey/commands/string_commands.rb
Sasidharan3094: Addressed above

test/unit/errors_test.rb
Sasidharan3094: Agreed. new version builds a fake FFI CommandResult directly and exercises convert_response's actual dispatch logic - checks EXECABORT maps to ExecAbortError and UNSPECIFIED stays a plain CommandError. so its actually testing the wiring now instead of just asserting class < class.
couldn't get a real integration test for EXECABORT tho, turns out a genuine wire-level EXECABORT is basically unreachable through this client right now - any queue-time arity/syntax error just short-circuits the read before it ever gets to the real EXEC reply, doesnt matter what exception: is set to. verified this empirically against a live server.

Sasidharan3094:
@Aryex I addressed the comments, can you review again

lib/valkey/errors.rb
```diff
@@ -15,6 +15,11 @@ def initialize(reply_type)
 
   class CommandError < BaseError; end
 
+  # Raised when EXEC aborts a transaction before any queued command ran
+  # (e.g. a queued command had a syntax/arity error) - distinct from a
+  # plain CommandError so callers know nothing was applied.
```
Aryex: Let's leave this out for now.

test/unit/errors_test.rb
Aryex: Seems like glide core only throws the first error it sees, hence in block form it never raises EXECABORT

```
client.multi # Sends MULTI
  begin
    client.call('GET') # Send GET, return error.
  rescue Valkey::CommandError
  end

  replies = client.exec # Send EXEC, throws EXECABORT
  puts replies.first.class  # Valkey::ExecAbortError
  puts replies.first.message 
```

For block form:
```
begin
  client.multi do |multi|     
    multi.call('GET')       
  end                          # block exits -> MULTI + GET + EXEC sent as ONE request
rescue Valkey::ExecAbortError => error
  puts "ExecAbortError: #{error.message}"
rescue Valkey::CommandError => error
  puts "#{error.class}: #{error.message}"
end
```

I think we can add 1 integration test for `EXECABORT` like above to cover the behavior, then the unit tests can be removed.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
...
... `exception:` kwarg (default `true`, matching `pipelined`), so `multi(exception: false) { ... }` returns an Array with `Valkey::CommandError` embedded at the position of any queued command that fails at runtime (e.g. `INCR` on a non-integer), instead of raising and discarding every other command's reply ([#260](https://github.com/valkey-io/valkey-glide-ruby/issues/260)). This matches the server's "no rollback on a runtime error" semantics — other queued commands still commit and their replies stay reachable even though the default is still to raise and discard them. This kwarg only affects the block form; the imperative `multi` (no block) / `#exec` pair has no equivalent control. A genuine queue-time abort (nothing ran) now raises the new `Valkey::ExecAbortError` (a `CommandError` subclass) instead of a plain `CommandError`, so callers can tell the two apart from a mid-EXEC runtime error. Also fixed `incrbyfloat` and `geodist` returning a `String` instead of a `Float` when called inside `multi`/`exec`'s imperative (non-block) form, and a latent bug where a runtime error embedded in that form's `EXEC` reply could be silently coerced to `true` by boolean-typed commands' reconversion.
```
Aryex: Let's fix this format as well. I believe AGENTS.md should have it defined.

Aryex (commented):
Reviewed up to [f7a88c2](https://github.com/valkey-io/valkey-glide-ruby/pull/299/commits/f7a88c234c0ccac1ff65347e581093a34a1a23f5)

Just a few comments. Other than that lgtm.

Sasidharan3094:
@Aryex Fixed review comments.

Aryex (approved):

PR #300, opened 2026-08-27
Fixing xpending org reordering

### LINKED ISSUE #241: `xpending` places the `IDLE` argument after `start`/`end`/`count` — server mis-interprets positional args, silently returning wrong results
## Summary

`Valkey#xpending(key, group, *args, idle: nil)` builds the outgoing argv as
`[key, group, *args, "IDLE", idle]` when `idle:` is given. The
[XPENDING protocol](https://valkey.io/commands/xpending/) is:

```
XPENDING key group [[IDLE min-idle-time] start end count [consumer]]
```

`IDLE min-idle-time` must appear **before** `start`, not after `count`. With
the current ordering the server treats `start` as the idle threshold,
`end` as `start`, `count` as `end`, and `"IDLE"` as `count` (which then fails
argument parsing) — or, on some argv shapes, drops the filter entirely and
returns unfiltered results. Either way, the caller sees a result set that
does not match the requested idle filter, with no error to warn them.

## Steps to reproduce

```ruby
require "valkey"

v = Valkey.new
v.del("s")
v.xadd("s", { "n" => "1" })
v.xadd("s", { "n" => "2" })
v.xgroup(:create, "s", "g", "0")
v.xreadgroup("g", "c1", "s", ">")           # claim two entries under c1

# Wait long enough for both entries to be older than 100ms:
sleep(0.15)

# Ask: what's pending on group g that's been idle > 100ms?
v.xpending("s", "g", "-", "+", 10, idle: 100)
# Outgoing argv (verify with MONITOR):
#   XPENDING s g - + 10 IDLE 100                 (BUG — wrong order)
#
# Server-side reply is either wrong-count or an ERR (server-dependent);
# neither reflects the requested IDLE filter.

# What the server needs:
v.call("XPENDING", "s", "g", "IDLE", 100, "-", "+", 10)
# => correctly filtered results
```

## Expected behavior

`xpending("s", "g", "-", "+", 10, idle: 100)` sends
`XPENDING s g IDLE 100 - + 10` and returns only pending entries idle for at
least 100 ms.

## Actual behavior

Sends `XPENDING s g - + 10 IDLE 100`. The idle filter is silently ignored or
raises a server-side error, depending on server version and argv parse path;
in neither case do the returned entries reflect the requested filter.

## Root cause

`lib/valkey/commands/stream_commands.rb:427–441`:

```ruby
def xpending(key, group, *args, idle: nil)
  cmd_args = [key, group]
  cmd_args.concat(args)              # <-- args (start/end/count) added first
  cmd_args << "IDLE" << idle.to_s if idle   # <-- IDLE appended after (wrong)

  send_command(RequestType::X_PENDING, cmd_args) do |reply|
    if args.length >= 2
      Utils::HashifyStreamPendingDetails.call(reply)
    else
      Utils::HashifyStreamPendings.call(reply)
    end
  end
end
```

## Suggested fix

Move the `IDLE` insertion before `*args` (and keep it out entirely on the
summary form, where `args.empty?`):

```ruby
def xpending(key, group, *args, idle: nil)
  raise ArgumentError, "idle: requires start/end/count" if idle && args.empty?

  cmd_args = [key, group]
  cmd_args << "IDLE" << idle.to_s if idle
  cmd_args.concat(args)

  send_command(RequestType::X_PENDING, cmd_args) do |reply|
    if args.length >= 2
      Utils::HashifyStreamPendingDetails.call(reply)
    else
      Utils::HashifyStreamPendings.call(reply)
    end
  end
end
```

Regression tests in `test/valkey/`:

- `xpending` with `idle:` produces argv `[..., "IDLE", n, start, end, count]`
  (verify with a monitor / mock).
- `xpending("s", "g", "-", "+", 10, idle: X)` returns only entries older than
  `X` ms; increasing `X` past the entry age returns `[]`.
- `xpending("s", "g", idle: 1)` raises `ArgumentError` (summary form does not
  accept an idle filter).

## Related

- Sibling to R3-7 (`xrange`/`xrevrange` response shape). Different root cause
  (argv order vs. reply shape), same file. Fix separately.
- Not tracked in any existing issue or in #216.

## Severity

`high` — silent wrong-result on any `xpending` call that uses `idle:`
(a common filter in stream-consumer health checks). Fix is a two-line
mechanical change plus a regression test.

### LINKED ISSUE #270: `xpending(idle:)` places IDLE in the consumer-name slot; silently returns zero results
### Status

Tracked upstream at [valkey-io/valkey-glide-ruby#241](https://github.com/valkey-io/valkey-glide-ruby/issues/241). File this issue for visibility, or link the reproduction into #241.

### Summary

The `idle:` option to `xpending` is appended after the `start` / `end` / `count` arguments. The Valkey `XPENDING` grammar requires `IDLE <ms>` **before** `start`. The trailing `"IDLE" <ms>` therefore gets parsed by the server as the optional `consumer` argument, so the server filters for a consumer literally named `IDLE`, finds none, and returns an empty list. No error is raised.

This is the standard consumer-group recovery pattern, find stalled pending entries by minimum idle time, and it is silently broken.

### Expected behavior

`client.xpending(key, group, "-", "+", count, idle: ms)` filters pending entries whose idle time is at least `ms` milliseconds. `idle: 0` matches every pending entry. Return shape matches `xpending` without the option.

Concretely, given a group with 2 pending entries:

```ruby
client.xpending("s", "g", "-", "+", 10, idle: 0).size
# => 2
```

### Actual behavior

```ruby
client.xpending("s", "g", "-", "+", 10, idle: 0).size
# => 0   (should be 2)
```

The server silently interprets `"IDLE" "0"` as the consumer name and returns an empty list because no consumer named `IDLE` exists.

The bare-summary form is also broken:

```ruby
client.xpending("s", "g", idle: 0)
# => Valkey::CommandError: syntax error
```

### Reproduce

```ruby
require "valkey"

c = Valkey.new(host: "127.0.0.1", port: 6379)
c.del("s")
c.xadd("s", { "field" => "v1" }, id: "1-1")
c.xadd("s", { "field" => "v2" }, id: "2-2")
c.xgroup(:create, "s", "g", "0")

# Read both entries into consumer "a" so they are pending.
c.xreadgroup("g", "a", "s", ">", count: 2)
sleep 0.05   # give them a tiny idle time

# Both entries should be returned since they are older than 0 ms idle.
detailed = c.xpending("s", "g", "-", "+", 10, idle: 0)
puts "with idle:      #{detailed.size}"

# Same call without the option: baseline, should also be 2.
baseline = c.xpending("s", "g", "-", "+", 10)
puts "without idle:   #{baseline.size}"

# Raw dispatch with the ARGS in the correct order: works.
raw = c.call("XPENDING", "s", "g", "IDLE", "0", "-", "+", "10")
puts "raw XPENDING:   #{raw.size}"
```

Output on rc5:

```
with idle:      0
without idle:   2
raw XPENDING:   2
```

### Acceptance criteria

- `client.xpending(key, group, start, stop, count, idle: ms)` returns the entries whose idle time ≥ `ms`, matching the server semantics.
- `idle:` is emitted in the correct grammar position (`XPENDING key group [IDLE ms] start end count [consumer]`).
- The summary form `xpending(key, group, idle: 0)` either works or raises a helpful `ArgumentError` instead of the current opaque server `syntax error`.

### Note on scope

Fix is a one-line reorder in `lib/valkey/commands/stream_commands.rb`: emit `["IDLE", idle.to_s]` immediately after `[key, group]`, before the `start`/`end`/`count` args. Sibling pattern in the same file (`blpop`/`brpop`) already emits its optional args in the correct grammar position.

---
_Filed from a blackbox bug bash of valkey-glide-rb 1.0.0.pre.rc5._

Sasidharan3094:
### Summary

`Valkey#xpending` emitted the `IDLE` filter argument *after* `start`/`end`/`count`
instead of before, so the server parsed `IDLE <ms>` as the optional `consumer`
name. Since no consumer is literally named `IDLE`, the server silently returned
an empty result set instead of raising an error or filtering correctly. The
summary form (`xpending(key, group, idle: ms)`) hit the same misordering and
raised an opaque server-side `syntax error`.

Fixed by inserting `IDLE <ms>` immediately after `key`/`group`, before
`start`/`end`/`count`, matching the `XPENDING key group [IDLE ms] start end
count [consumer]` grammar. Since the summary form has no `start`/`end`/`count`
slot for `IDLE` to precede, combining it with `idle:` now raises a clear
`ArgumentError` instead of an opaque server error.

### Issue link

Closes #270
Closes #241

### Features / Changes

- Reordered `xpending`'s outgoing argv in `lib/valkey/commands/stream_commands.rb`
  so `IDLE <ms>` precedes `start`/`end`/`count` instead of following them.
- Added a guard that raises `ArgumentError` when `idle:` is passed to the
  summary form (`xpending(key, group, idle: ms)`), since that combination is
  not valid per the `XPENDING` grammar.

### Testing

- Added `test_xpending_with_idle_filter` to `test/lint/stream_commands.rb`:
  verifies `xpending(key, group, "-", "+", count, idle: 0)` returns all pending
  entries, and that a large `idle:` threshold (well past the entries' actual
  age) filters them out entirely.
- Added `test_xpending_summary_with_idle_raises_argument_error` to
  `test/lint/stream_commands.rb`: verifies `xpending(key, group, idle: ms)`
  raises `ArgumentError` instead of surfacing a server-side syntax error.
- Ran `bundle exec rubocop` on both changed files — no offenses.
- Ran the `xpending`-related tests against both standalone and cluster
  integration suites (`test/integration/standalone/commands_test.rb`,
  `test/integration/cluster/cluster_commands_test.rb`) — all passing (0
  failures, 19 assertions each).
- Manually reproduced the original bug report's repro script before and after
  the fix to confirm the before/after behavior matches the issue exactly.

### Checklist


test/lint/stream_commands.rb
```diff
@@ -336,6 +336,40 @@ def test_xpending
       r.del "mystream"
     end
 
+    def test_xpending_with_idle_filter
```
Aryex, line 343: We recently merged in changes to reorganize the tests into intergrations and unit tests. Could you rebase and add unit tests to ensure the arg constructions is correct? E2E for this would be just a pass through

lib/valkey/commands/stream_commands.rb
```diff
@@ -432,9 +432,11 @@ def xack(key, group, *ids)
       #
       # @see https://valkey.io/commands/xpending/
       def xpending(key, group, *args, idle: nil)
+        raise ArgumentError, "idle: requires start, end, and count arguments" if idle && args.empty?
```
Aryex: We probably don't need this. The server handles input validation so this is just repeated code.

lib/valkey/commands/stream_commands.rb
```diff
@@ -432,9 +432,11 @@ def xack(key, group, *ids)
...
+
         cmd_args = [key, group]
-        cmd_args.concat(args)
         cmd_args << "IDLE" << idle.to_s if idle
```
Aryex: We should probably explicitly nil check the idle.  When `idle = false` it would also hit this line causing `IDLE` to silently dropped making the call pass where it should've failed.

```
r.xpending("s", "g", "-", "+", 10, idle: false) ## OK. It should've rejected with syntax error from the server.
```

Consider using this instead:
```
cmd_args << "IDLE" << idle.to_s unless idle.nil?
```

Sasidharan3094: Updated

lib/valkey/commands/stream_commands.rb
Sasidharan3094: Agreed, removed.

test/lint/stream_commands.rb
Sasidharan3094: Yes added unit test as per new reorg structure.

jamesx-improving (approved):
Verified the reordered argv against the XPENDING grammar (IDLE now correctly precedes start/end/count), and the ArgumentError guard cleanly rejects the invalid summary-form-with-idle case. Tests cover both paths and CI is green. LGTM.

Aryex (approved):

PR #302, opened 2026-08-28
fix(ruby): raise Valkey::InheritedError for clients used after fork()

### LINKED ISSUE #255: Client crashes child process on fork()
## Overview

FFI-based sync clients crash the **child** process with `SIGTRAP` when a `Valkey` client instance created before `fork()` is used after the fork. The abort happens below the Ruby VM — `rescue Exception` does **not** catch it, so user code has no way to recover. The parent process is unaffected.

This is the Ruby side of upstream **valkey-io/valkey-glide#6672**. A maintainer confirmed there that the async-only fork work (valkey-io/valkey-glide#6679, valkey-io/valkey-glide#6716) "must be extended to all clients" — the Ruby binding and Python `glide-sync` share the same FFI sync path and neither is covered yet.

## Root cause

`glide-core` embeds per-client tokio runtimes inside the Rust `ClientAdapter` — a `current_thread` runtime that `block_on`s on the calling Ruby thread, plus a `multi_thread` background runtime (1 worker thread) that owns the TCP I/O reactor. On `fork()`, POSIX copies only the calling thread; the background worker thread and its epoll/kqueue fds are inherited in an unusable state. The first FFI command in the child pokes the dead reactor and panics with:

```
WARN glide-core::client - failed to wake I/O driver: Bad file descriptor
```

which unwinds to `abort()` → `SIGTRAP` (signal 5). Ruby prints `[BUG] Aborted` with a truncated backtrace.

## User impact

Every mainstream Ruby web/worker deployment model forks after `require`: **Puma (clustered), Unicorn, Passenger, Sidekiq, Resque**. Any user following the natural "boot once, share the client" pattern (connect in the parent / preload, then fork workers) will hit this the moment a worker issues its first command.

## Reproduction

Requires a Valkey/Redis server on `127.0.0.1:6379`.

```ruby
require "valkey"

client = Valkey.new
puts "parent ping: #{client.ping}"   # PONG

pid = fork do
  begin
    client.ping                        # aborts below the VM
  rescue Exception => e
    puts "caught #{e.class}"; exit 2
  end
  exit 0
end

_, status = Process.wait2(pid)
puts "child signaled: #{status.signaled?}, signal: #{status.termsig}"  # signaled: true, signal: 5 (SIGTRAP)
puts "parent still works: #{client.ping}"                              # PONG
```

## Expected behavior

Mirror the redis-rb / redis-client idiom: **transparently recover after fork** rather than crash. redis-rb detects a PID change on each command (`@pid != Process.pid`) and reconnects the inherited connection automatically; the newer redis-client does the same via `PIDCache`. The Python `glide-sync` client already solves this by re-creating the core client in the child (`os.register_at_fork(after_in_child=_create_core_client)`).

## Proposed fix (Ruby-only — no core/FFI rebuild required)

Because each sync client owns its own tokio runtime, the child can simply build a **fresh** client handle (fresh runtime) via the existing `create_client_from_uri` FFI — no new native entrypoint is needed. This is exactly what Python `glide-sync` does.

1. Stash the connection config (`uri_str`, `json_str`, `client_type`, pubsub callback) and `Process.pid` on the client during `#initialize` (currently discarded locals). Extract the creation block into a private `establish_connection` helper.
2. In `connection!` (the single chokepoint for `command` / `command_with_route_info` / `batch`), detect a PID change and lazily rebuild the handle before dispatching. **Never call `close_client` on the inherited handle in the child** — that would poke the dead runtime and trigger the very crash we're avoiding; the stale parent handle must be leaked in the child (as Python sync does).
3. Guard `close` similarly (`@pid != Process.pid` → skip the native close).
4. Optionally expose a public `after_fork!` method so users can eagerly recreate the client from a Puma/Sidekiq/Unicorn fork hook, matching redis-rb ergonomics.

Estimated size: ~20 lines for the minimal working fix, ~30–40 lines production-hardened (thread-safety under `@close_lock`, `close` guard, `after_fork!`), plus a fork-safety integration test (~100 lines) modeled on Python's `test_fork_safety.py` — asserting the child issues a command without `SIGTRAP` and the parent remains functional.

## Open item to verify

`glide-core`'s `TimeoutWatchdog` is a process-global thread that also dies on fork and is currently only revived on the async path (`reinit_async_pipe` → `TimeoutWatchdog::reinit_global`). It does **not** cause the `SIGTRAP`, but we should confirm whether the sync command path depends on it in the child; if so, a follow-up FFI change may be needed to respawn it.

## References

- Upstream tracking issue: valkey-io/valkey-glide#6672
- Python async fork-safety (out of scope for sync path): valkey-io/valkey-glide#6679, valkey-io/valkey-glide#6716
- Reproduces on `valkey-glide-rb` 1.0.0.pre.rc3

Aryex:
### Summary

This PR add fork safter for Ruby client by ensuring users must recreate the client on the child process. This is handled by associating the pid to a connection.

This assumes [valkey-glide#6912](https://github.com/valkey-io/valkey-glide/pull/6912) is accepted and merged. Otherwise the Ruby client will also need to restart the watchdog manually as well. 

### Reviewers

Actual code changes are quite small, the bulk of the changes are test code. 

### Issue link

Closes [#255](https://github.com/valkey-io/valkey-glide-ruby/issues/255)
Related [valkey-glide#6912](https://github.com/valkey-io/valkey-glide/pull/6912) 

### Changes

| File | Change |
|---|---|
| `lib/valkey.rb` | Stamp `@pid` in `initialize`; `connection!` raises `InheritedError` on mismatch; `close` skips the native free on mismatch but still clears `@connection` |
| `lib/valkey/errors.rb` | Document the existing, previously unused `InheritedError` |
| `README.md` | Post-fork guidance and the known limitation |

`@connection` is read in exactly two places, so the one check in `connection!` covers `send_command`, `send_batch_commands` and `invoke_script`. `get_statistics` takes no handle and stays unguarded.

The message text matches Python async's `ClosingError`, so the same bug reads the same way across bindings.

### Limitations

- A client created **in the child** still crashes if the parent issued any command — the process-global watchdog isn't repaired here. That fix belongs in glide-core ([#6912](https://github.com/valkey-io/valkey-glide/pull/6912)); one integration test is skipped until the submodule is bumped past it. So: an unrescuable signal becomes a catchable error, not yet a working post-fork path.
- No transparent recovery. Nothing regresses, since there was no working post-fork path to preserve, and adding recovery later is a compatible change.
- `cd.yml` runs no server, so fork tests can't run against the published gem. CI covers every matrix combination.

### Testing

| Suite | Coverage |
|---|---|
| `test/unit/fork_safety_test.rb` | 16 server-free tests |
| `test/integration/valkey/fork_safety_test.rb` | 2 real-`fork()` tests, standalone and cluster |
| `test/isolated/fork_no_parent_command_test.rb` | Own process from `ci.yml`; needs a process that has issued no command, which a shared suite can't offer |

Tests assert on `Process::Status`, not exit status alone — the bug is a signal, and an exit-code-only check would let a regression look like an ordinary failure.

```
137 unit / 741 standalone / 731 cluster / 1 isolated — 0 failures
rubocop: 107 files, no offenses
```

### Checklist


jamesx-improving (approved):
LGTM

yipin-chen (approved):

jamesx-improving (approved):

Issue #129, opened 2026-07-07
Connection config parity: pubsub_subscriptions


Aryex:
## Overview

Support pubsub configuration through the client configuration.

## Description

This is a different mechanism from the client's existing pub/sub delivery path. Message delivery
(receiving anything you're subscribed to, regardless of how the subscription was made) already
works via `method(:pubsub_callback)` (`lib/valkey/pubsub_callback.rb`), registered once at
connection time and unrelated to protobuf as well. `pubsub_subscriptions` only changes *when* the
subscription itself is established — at connect time instead of via an explicit `subscribe` call
afterward — the delivery mechanism once subscribed is identical.

## Proposal

```ruby
client = Valkey.new(
  host: "localhost",
  port: 6379,
  pubsub_subscriptions: {
    exact: ["channel1", "channel2"],
    pattern: ["news.*"]
  }
)
# client is already subscribed to channel1, channel2, and news.* the moment
# it connects — no subscribe call needed, delivery still goes through the
# existing pubsub_callback path.
```

Accept symbol channel-type keys (`:exact`, `:pattern`, `:sharded`) in Ruby, map to the integer JSON
keys (`0`/`1`/`2`) the FFI layer expects internally — the documented `0`/`1`/`2` shape is not a
friendly Ruby API to expose directly.

PR #304, opened 2026-09-01
feat(pubsub): introduce PubSub handler

Aryex:
### Summary

This PR introduces `PubSub` as the abstraction to handle pubsub functionalities for the Ruby client.

Initial implementation of #135 

### Changes

- Added `PubSub` class.
- Added Unit tests.

### Limitations

This is only the initial implementation. PubSub functionalities is incomplete. More PR is required.

### Testing

Unit tests added.

### Checklist

- [ ] CHANGELOG.md 


lib/valkey/glide/pubsub.rb
```diff
@@ -0,0 +1,167 @@
...
+          { "pubsub_subscriptions" => mapped }
+        end
+
+        def validate!(subscriptions, protocol:)
+          return if subscriptions.empty?
+
+          unknown_modes = subscriptions.keys - MODE_CODES.keys
+          raise ArgumentError, unknown_mode_message(unknown_modes) if unknown_modes.any?
+
+          raise ArgumentError, "Pub/Sub requires RESP3 protocol. Found #{protocol}" if RESP2_OPTIONS.include?(protocol)
```
jamesx-improving: This only rejects an explicitly-passed resp2; when `protocol:` is omitted the connection defaults to RESP2 (lib/valkey.rb:128-129), so `Valkey.new(pubsub: {subscriptions: {...}})` with no protocol still builds an unsupported RESP2 pub/sub connection and skips this guard.

**Suggested fix:** validate the effective protocol (treat unset as RESP2 and reject), or default to RESP3 when subscriptions are present.

Aryex: Addressed in [c1586ad](https://github.com/valkey-io/valkey-glide-ruby/pull/304/commits/c1586ad1a41c4f543862170d6d584278b3823b33)

Aryex:
[53cfd41](https://github.com/valkey-io/valkey-glide-ruby/pull/304/commits/53cfd41972e728f7486db155637114176a90585b) added stub api methods. 

One question, our abstractions currently involves Valkey > PubSubCommands > PubSub. 

As is, PubSubCommands is just purely an "interface" that delegate to PubSub. I'm questioning if this is needed? Perhaps making `PubSub` accessible would be enough?

For example:
```
client = Valkey.new(..) # Valkey implement PubSubCommands

client.try_get_message() # Using the PubSubCommands interface.

client.pubsub.try_get_message() # Making PubSub public instead.
```

lib/valkey/glide/pubsub.rb
```diff
@@ -0,0 +1,503 @@
...
+      #
+      # TODO: unfinished -- the handler still has to:
+      #   - return early unless PushKind::MESSAGE_KINDS covers kind
+      #   - log DISCONNECTION, drop the confirmation kinds silently
+      #   - read_string(len) for message and channel; pattern is nil when
+      #     pat_ptr is null (exact and sharded pushes)
+      #   - deliver(Message.new(...))
+      #   - rescue StandardError and swallow: an exception must never cross the
+      #     FFI boundary
+      def build_ffi_handler
```
affonsov, line 488: Delivery is understood to be a follow-up (#135), but this PR already wires `parse_config` into the live connection (`lib/valkey.rb:251-265`) and ships a working `get_pubsub_message`, while this handler stays a no-op. So on this commit a connect-time subscription silently drops every message and `get_pubsub_message` blocks forever — a reachable hang with no error and no way to recover short of `close`.

Since delivery lands in a later PR, consider gating the intermediate state so this commit can't hang a user: either hold the `parse_config`→`create_client` wiring (connect-time subscriptions are tracked separately in #129), or have `get_message`/`try_get_message` raise `NotImplementedError` like the rest of the surface until the handler exists.

lib/valkey/glide/pubsub.rb
```diff
@@ -0,0 +1,503 @@
...
+
+          validate!(subscriptions, protocol: protocol)
+
+          to_ffi(pubsub_configs)
+        end
+
+        private
+
+        # Converts the configuration to what the ffi expects.
+        def to_ffi(pubsub_configs)
```
affonsov, line 117: `to_ffi` silently drops `:callback`/`:context` even though `parse_config`'s doc documents them. Since the callback path isn't supported yet, raising `NotImplementedError` when `:callback` is present (or ignoring it with an explicit note) would be more consistent with the rest of the unfinished surface than a silent no-op.

affonsov (commented):
Some suggestions

affonsov (approved):

lib/valkey/glide/pubsub.rb
Aryex: Will defer this to follow up PRs.

lib/valkey/glide/pubsub.rb
Aryex: Intended. Will defer this to follow up PRs.

jamesx-improving, line 117: Agreed, and I'd argue this one is worth doing here rather than deferring: `pubsub:` reaches `parse_config` straight from `Valkey.new` (lib/valkey.rb:251), so `callback:` is the only part of this surface a user can actually reach today — everything else sits behind the still-commented-out `include PubSubCommands` (lib/valkey/commands.rb:51). A `raise NotImplementedError if pubsub_configs[:callback]` in `validate!` closes it in one line. 👍

lib/valkey/glide/pubsub.rb
jamesx-improving, line 488: Agreed on deferring. One note for the follow-up: the hang isn't reachable yet — `get_pubsub_message` is in `PubSubCommands`, whose `include` is still commented out (lib/valkey/commands.rb:51), and `@pubsub` has no reader on `Valkey`, so nothing public can block on the queue.

jamesx-improving (approved):
Of @affonsov's two comments, I'd fix the silently-dropped `pubsub: { callback: }` in this PR — it's the only part of this surface a user can reach today via `Valkey.new`. The no-op handler one is fine to defer.

lib/valkey/glide/pubsub.rb
Aryex: Callback is currently a no op so it found break anything. This is only intended as stubs. Will defer to the next PR to finish this.

jamesx-improving (approved):

PR #308, opened 2026-09-08
feat(pubsub): inline Pub/Sub APIs for exact channels

Aryex:
### Summary

This adds the part 2 of PubSub APIs implementation [plan](https://github.com/valkey-io/valkey-glide-ruby/issues/135#issuecomment-5485707606). 

### Issue link

Related #135 

### Changes

- `PubSub` implemented to support the APIs, including the ffi callback.
- RESP3 is enforced client side. 
- unit and integration tests added.

### Limitations

- `publish` takes the message first — `publish(message, channel)` — matching the other GLIDE clients rather than redis-rb's `publish(channel, message)`.
- `timeout:` is documented and treated as **milliseconds**, not seconds. Worth confirming that's the intended public unit, since the rest of the client's timeouts are in seconds.

### Testing

- `bundle exec rubocop` — clean.
- `bundle exec rake test:unit` — 227 tests, 480 assertions, 0 failures, 0 errors, 0 skips. `test/unit/glide/pubsub_test.rb` grows by ~234 lines covering the RESP3 guard, `timeout_argument` conversion and rejection, and the push handler's kind filtering / NUL-safe reads / exception swallowing.
- `test/integration/valkey/pubsub_test.rb` — new, ~277 lines, wired into the standalone suite. Covers subscribe/publish/receive round trips, blocking vs non-blocking consume, and the negative paths.
- Integration suite has not been run against a live server in this pass; that's the main gap before this leaves draft.

### Checklist

- [ ] CHANGELOG.md and documentation files are updated.


lib/valkey/commands.rb
```diff
@@ -47,8 +47,7 @@ module Commands
     include SetCommands
     include ScriptingCommands
     include FunctionCommands
-    # TODO: https://github.com/valkey-io/valkey-glide-ruby/issues/135
-    # include PubSubCommands
+    include PubSubCommands
```
jamesx-improving, line 50: This also mixes `PubSubCommands` into `Valkey::Pipeline` (`pipeline.rb:7` includes `Commands`), which never assigns `@pubsub` — so `r.pipelined { |p| p.publish("m", "c") }` raises `NoMethodError: undefined method 'publish' for nil` and aborts the whole batch, where PUBLISH would otherwise pipeline fine.

**Suggested fix:** have `PubSubCommands#publish` call `send_command(RequestType::PUBLISH, [channel.to_s, message.to_s])` on `self` (keeping the `sharded:` guard in the mixin) so it works for both the client and a pipeline; the subscribe/consume methods aren't pipelineable and are worth an explicit `Pipeline` override that raises something readable.

lib/valkey/glide/pubsub.rb
```diff
@@ -477,19 +489,36 @@ def pubsub(subcommand, *args) = raise(NotImplementedError, "#{__method__} is not
...
         end
       end
 
+      def validate_resp3!
+        raise Resp3RequiredError, @protocol unless RESP3_VALUES.include?(@protocol)
+      end
+
+      # glide-core takes the timeout as the last command argument, in whole
+      # milliseconds, and reads a zero as "no deadline".
+      def timeout_argument(timeout)
```
jamesx-improving: `timeout: nil` → `"0"`, and glide-core reads 0 as *no deadline*: `wait_for_sync` sets `deadline = None` and falls through to a bare `notified.await`, and the pubsub interception bypasses `request_timeout`. Since `command` is `blocking: true`, a subscribe that never reconciles (server restarted right after connect) parks the thread permanently — `Timeout.timeout` and `Thread#raise` can't interrupt a thread inside a blocking FFI call — and this PR removed the "`nil` blocks indefinitely" line from every public doc comment.

**Suggested fix:** restore that wording on the blocking methods, or default `timeout:` to the client's `request_timeout` so the failure is a `Valkey::TimeoutError` instead of a hang.

AGENTS.md
```diff
@@ -259,11 +259,11 @@ cargo fmt --manifest-path ./Cargo.toml --all
 
 - **Ruby 3.0+ Required:** Minimum per `valkey.gemspec`
 - **FFI dependency:** `ffi ~> 1.17.0`; do not break ABI without rebuilding native lib
+- **FFI bindings MUST pass `blocking: true`:** Every `attach_function` in `lib/valkey/bindings.rb` releases the GVL. Ruby-FFI defaults to `blocking: false`, which holds the GVL for the entire native call — that stalls every other Ruby thread, and deadlocks outright whenever the native call waits on a thread that needs the GVL to make progress.
```
jamesx-improving: "Every `attach_function`" isn't accurate — 16 of the 21 in `bindings.rb` still default to `blocking: false`, including `create_client` and `create_client_from_uri` (`bindings.rb:238`, `:245`), the two that install this Pub/Sub callback and that hold the GVL for the entire connect-and-retry window.

**Suggested fix:** add `blocking: true` to both `create_client*` functions, and scope the rule to calls that can actually block (connect, command, batch, script, close) — the `free_*` and OTel span helpers don't need it.

jamesx-improving (commented):
CHANGELOG.md still needs a 1.1.0 entry for this — it's the largest user-facing change in the release and the checklist box is unchecked.

On your open question: milliseconds is the right wire unit — glide-core parses the last argument as a u64 ms count and Go's peer API names the parameter `timeoutMs` — but `timeout:` reads as seconds next to every other timeout in this client, so consider renaming the kwarg `timeout_ms:` while the API is still unreleased.

AGENTS.md
Aryex: Good point. updated in [587e665](https://github.com/valkey-io/valkey-glide-ruby/pull/308/commits/587e66594e5d8c674975d2926d5942252a245eef)

I have also raised #309 to fix the create_client methods.

lib/valkey/glide/pubsub.rb
Aryex: Fixed in [730dfaa](https://github.com/valkey-io/valkey-glide-ruby/pull/308/commits/730dfaaff9510d2c7d1b21872814547a2482d0d7)

lib/valkey/commands.rb
Aryex: This is a good point and raised an oversight in my design of the PubSub class. In order to support PubSub in Batch, we will need to refactor `PubSub` into a mixin. I will handle this in a follow up PR once this is merged.

Aryex: I guess it is less of an incompatibility and more that it is an anti-pattern with how the current client work. See my [comment](https://github.com/valkey-io/valkey-glide-ruby/issues/135#issuecomment-5593569418).

My suggestions is to perform this refactor after this is merged. Then proceed with support for other APIs

test/integration/standalone/valkey_test.rb
```diff
@@ -16,6 +16,7 @@ class TestStandaloneValkey < Minitest::Test
   include ValkeyTests::FailoverCommands
   include ValkeyTests::FunctionCommands
   include ValkeyTests::GenericCommands
+  include ValkeyTests::PubSub
```
affonsov, line 19: `ValkeyTests::PubSub` is wired into the standalone suite here but not into the cluster suite (`test/integration/cluster/cluster_commands_test.rb`), which already mixes in most of the shared command modules (`GenericCommands`, `Scanning`, `ScriptingCommands`, `Sorting`, …). Exact-channel subscribe/publish/receive works in cluster mode too — only *sharded* is cluster-only and out of scope for this PR — and the cluster path is meaningfully different: the message is delivered on the node the request routed to, and `publish` returns the count on that node only. A green standalone run doesn't exercise that. Consider adding `include ValkeyTests::PubSub` to the cluster suite, or leaving a one-line note deferring exact-channel cluster coverage to a tracking issue.

lib/valkey/glide/pubsub.rb
```diff
@@ -477,19 +495,35 @@ def pubsub(subcommand, *args) = raise(NotImplementedError, "#{__method__} is not
...
+
+      # glide-core takes the timeout as the last command argument, in whole
+      # milliseconds, and reads a zero as "no deadline".
+      def timeout_argument(timeout_ms)
+        valid = timeout_ms.is_a?(Numeric) && !timeout_ms.negative?
+        raise ArgumentError, "Timeout must be a non-negative number, got: #{timeout_ms.inspect}" unless valid
+        return "0" if timeout_ms.zero?
+
+        # Handling floats.
+        [timeout_ms.to_i, 1].max.to_s
```
affonsov, line 524: `timeout_argument(Float::INFINITY)` (and `Float::NAN`) passes the `is_a?(Numeric) && !negative?` check, isn't `zero?`, then reaches `timeout_ms.to_i` on this line — which raises `FloatDomainError`, a bare Ruby error outside this method's documented `@raise [ArgumentError]` contract. `Float::INFINITY` is a plausible "block forever" input a caller might try. Consider rejecting non-finite values in the validity check (e.g. `&& timeout_ms.finite?`) so they surface as the documented `ArgumentError`. Minor.

test/integration/valkey/pubsub_test.rb
affonsov, line 1: I know that you are adding tests as you go, but worth it to start parameterization the tests, like the other languages, pubsub has a lot of different combinations. You don't need to do it now, but it worth it consider in the next parts.

lib/valkey/commands.rb
jamesx-improving: Correcting my own comment above: I said this PR causes the batch to abort, and that's inaccurate — at the base `include PubSubCommands` was commented out, so the method was absent and the call already raised `NoMethodError` into `pipelined`'s `rescue StandardError` → `abort_futures!`. Same abort before and after; the only real delta is `pipeline.respond_to?(:publish)` now returning `true`.

#135 is in the same 1.1.0 milestone as this PR, so nothing needs doing here — no released version is affected.

lib/valkey/glide/pubsub.rb
jamesx-improving: `finite?` wouldn't be enough — the guard has no upper bound either, and glide-core coerces an unparseable timeout to `0` (= no deadline), so a finite `1e20` silently becomes an unbounded wait. Unlikely magnitude, but a `0..u64::MAX` range check covers both cases.

jamesx-improving (approved):
Both notes below are non-blocking.

Needs a rebase before merge — behind main since #299. The CHANGELOG entry's trailing "Requires `protocol: :resp3`" also reads as covering the whole list, but `publish` is deliberately ungated (`test_publish_works_without_resp3`).

lib/valkey/glide/pubsub.rb
Aryex: honestly I think this should be handled at the core, not at the client... 

That said, I think this is ok. I would consider infinity to be an unsupported values which would result in undocumented behavior, in this cases raising FloatDomainError. 

People wanting infinite timeout should just use blocking calls instead.

test/integration/standalone/valkey_test.rb
Aryex: Good point. Fixed in [0bd36d4](https://github.com/valkey-io/valkey-glide-ruby/pull/308/commits/0bd36d403a345ff95897091d88a8a90c2d7249f0)

test/integration/valkey/pubsub_test.rb
Aryex: Will do. Note that the current test framework does not have parameterized tests, but I will take a look at existing PubSub tests as reference.

affonsov (approved):

PR #311, opened 2026-09-09
refactor(pubsub): mixin refactor

Aryex:
### Summary

This PR refactor the `PubSub` handler class into a mixin module, per the comment [here](https://github.com/valkey-io/valkey-glide-ruby/issues/135#issuecomment-5593569418).

### Issue link

Related #135 

### Changes

* `PubSub` handler class removed
* Pubsub handlings moved into the mixin module `pubsub_commands.rb`
* Plumbing code for unit tests

### Limitations

- No functional changes. 

### Testing

All integration tests passed locally. Unit tests assertions were not touched, only plumbing code.

### Checklist

- [ ] Destination branch is correct.


lib/valkey/pipeline.rb
```diff
@@ -185,5 +185,85 @@ def call_v(argv)
       super
     end
     # rubocop:enable Lint/UselessMethodDefinition
+
+    def subscribe(*_channels, timeout_ms: 0)
+      raise CommandError, "subscribe is not supported in a pipeline or transaction"
```
nderraugh: Valkey::CommandError is the base for server-side command errors (WrongTypeError, NoScriptError, etc.). Calling subscribe on a pipeline is a caller mistake, and the existing precedent in this repo is ArgumentError: Pipeline#call with route: and ft_search/ft_aggregate inside a batch both raise ArgumentError with a "not supported inside pipelined/multi" message

.rubocop.yml
```diff
@@ -45,6 +45,7 @@ Metrics/MethodLength:
 Metrics/ClassLength:
   Exclude:
     - 'lib/valkey.rb'
+    - 'lib/valkey/pipeline.rb'
```
nderraugh: Suggest removing this, and the Lint/UnusuedMethodArgument lint, and replacing the implementation in pipeline.rb with 
```
%i[subscribe unsubscribe psubscribe ... pubsub_shardnumsub].each do |name|
  define_method(name) do |*, **|
    raise ArgumentError, "#{name} is not supported inside pipelined/multi"
  end
end
```
Future unused-arg bugs in pipeline.rb ought to be caught.

lib/valkey/commands/pubsub_commands.rb
```diff
@@ -436,8 +420,55 @@ def pubsub_shardchannels(pattern = nil)
...
+      # Get the next Pub/Sub message if one is already queued. Never blocks.
+      #
+      # @example Poll for a message
+      #   valkey.try_get_pubsub_message
+      #     # => #<struct Valkey::Commands::PubSubCommands::Message message="hi", channel="channel1", pattern=nil>
+      # @example Poll when nothing is queued
+      #   valkey.try_get_pubsub_message
+      #     # => nil
+      #
+      # @return [Valkey::Commands::PubSubCommands::Message, nil] the message, or `nil` when the queue is empty or the
```
nderraugh: Naming seems redundant, and pretty verbose.

Valkey::Commands::PubSubCommands::Message for what a user gets back form every get_pubsub_message is pretty long.

nderraugh:
Are you planning to beef up the testing around pubsub alongside pipelined, multi, and Pipeline.new?

Aryex:
> Are you planning to beef up the testing around pubsub alongside pipelined, multi, and Pipeline.new?

Yes, pipeline pubsub support would be handled in a separate PR.

.rubocop.yml
Aryex: Addressed in [b7f9bc9](https://github.com/valkey-io/valkey-glide-ruby/pull/311/commits/b7f9bc909f387b1c11018dc6a235c4a8272a34d8)

lib/valkey/pipeline.rb
Aryex: addressed in [b7f9bc9](https://github.com/valkey-io/valkey-glide-ruby/pull/311/commits/b7f9bc909f387b1c11018dc6a235c4a8272a34d8)

lib/valkey/commands/pubsub_commands.rb
Aryex: Addressed in [e3948eb](https://github.com/valkey-io/valkey-glide-ruby/pull/311/commits/e3948eb7e6f1c8b183643b17495b91c25b5a7f1c).

Moved PushKind to internal class since users won't need to use it. 

The other two types are now `Valkey::Glide::PubSubState` and `Valkey::Glide::PubSubMessage`.

nderraugh (approved):

lib/valkey/pipeline.rb
```diff
@@ -185,5 +185,22 @@ def call_v(argv)
...
     # rubocop:enable Lint/UselessMethodDefinition
+
+    # Subscriptions outlive a batch and pushes arrive out of band, so neither
+    # can be expressed as one queued reply.
+    PUBSUB_UNSUPPORTED = %i[
+      subscribe unsubscribe psubscribe punsubscribe ssubscribe sunsubscribe
+      subscribe_lazy unsubscribe_lazy psubscribe_lazy punsubscribe_lazy
+      ssubscribe_lazy sunsubscribe_lazy
+      get_subscriptions get_pubsub_message try_get_pubsub_message
+      pubsub_channels pubsub_numpat pubsub_numsub
```
jamesx-improving, line 196: `pubsub_channels`, `pubsub_numpat`, `pubsub_numsub`, `pubsub_shardchannels` and `pubsub_shardnumsub` are plain request/response `PUBSUB` subcommands that pipeline fine — the comment's rationale (subscriptions outlive a batch, pushes arrive out of band) doesn't cover them.

Nothing breaks today because all five still raise `NotImplementedError`, but the list lives in a different file from the mixin, so #135 will land them pre-blocked.

**Suggested fix:** drop those five, and declare the list in `PubSubCommands` so it can't drift from the methods it mirrors.

jamesx-improving (commented):
`test_ffi_handler_is_retained` was dropped and not carried into `pubsub_receiver_test.rb`, so nothing pins `ffi_handler` to a single memoized proc any more — if it ever returns a fresh lambda, ruby-ffi builds a new `FFI::Function` per call and the one glide-core holds can be collected, segfaulting on the next push instead of failing a test.

Worth restoring next to the three handler tests that did move; "Unit tests assertions were not touched" in the PR body needs the same correction. One more non-blocking note inline.

lib/valkey/pipeline.rb
Aryex: This list is specific to pipeline so not appropriate to move it elsewhere.

Aryex:
> `test_ffi_handler_is_retained` was dropped and not carried into `pubsub_receiver_test.rb`, so nothing pins `ffi_handler` to a single memoized proc any more — if it ever returns a fresh lambda, ruby-ffi builds a new `FFI::Function` per call and the one glide-core holds can be collected, segfaulting on the next push instead of failing a test.
> 
> Worth restoring next to the three handler tests that did move; "Unit tests assertions were not touched" in the PR body needs the same correction. One more non-blocking note inline.

Fixed in [44fc545](https://github.com/valkey-io/valkey-glide-ruby/pull/311/commits/44fc5456ad1d7140b5e13cefcbfb3f65a6297932)

lib/valkey/pipeline.rb
jamesx-improving: Agreed on placement — it's pipeline's own constraint, and moving it into `PubSubCommands` would invert the dependency. Withdrawing that half.

Separate question though: those five `pubsub_*` are single-reply wire calls, so trimming them is independent of where the list lives.

jamesx-improving (approved):
LGTM.

`test_ffi_handler_is_retained` is back and does its job — I mutated the reader to build a fresh lambda per call and the assertion goes red, so the retention invariant is genuinely pinned now.

The deny-list note stays open as a non-blocking latent item for #135; it's in the thread above, nothing to do here.

Issue #309, opened 2026-09-08
(ffi): Set blocking: true on create_client_from_uri ffi


Aryex:
## Overview

`create_client_from_uri` ffi method is attached without `blocking: true`, so Ruby-FFI holds the GVL for the entire native call. The native call blocks until the connection is established or the retry strategy gives up, so every other Ruby thread is frozen for that whole window.

`create_client` is also currently dead code and can be removed.

## Expected Behavior

`Valkey.new` releases the GVL while it waits on the network, so unrelated threads keep running.

## Current Behavior

Unrelated threads make no progress until the connect returns or gives up. Against an unreachable host that is the full retry window (seconds), during which timers, health checks and request handlers all stop.

## Reproduction Steps

N/A.

## Possible Solution

* Add `blocking: true` to `create_client_from_uri`. 
* Remove `create_client` as it is dead code.

## Additional Information/Context

Raised in review [comment](https://github.com/valkey-io/valkey-glide-ruby/pull/308#discussion_r3960699330). 

Related but distinct: the GVL-holding-with-callback pattern can deadlock outright, not just stall, when the native call waits on something that needs the GVL. That is not reachable here — the Pub/Sub push handler is spawned after `block_on` returns (`ffi/src/lib.rs:1477`), so no callback can fire mid-connect. It is one refactor away from being live, which is part of the argument for fixing the flag now.

## Client version used

valkey-glide-rb 1.0.0

## Engine type and version

Valkey 8.0

## Engine Environment

Local (no server needed — the bug reproduces against an unreachable address)

## OS

macOS 26.6.2 (arm64)

## Ruby Version

Ruby 3.4.9

## Cluster information

N/A — standalone.

## Logs

N/A.

## Other information

The 16 bindings without `blocking: true` are not all wrong. Only the `create_client*` pair blocks; the `free_*`/`drop_*` helpers, the OTel span constructors and `get_statistics` never wait, so annotating them would add handoff cost for no benefit.

Issue #257, opened 2026-08-10
Subscriber segfaults (SIGSEGV) after ~15 delivered pub/sub messages — unrescuable, kills the whole process


jamesx-improving:
## Summary

A client that calls `subscribe` and then receives push messages crashes the
entire Ruby process with `SIGSEGV` (exit status 139) once roughly 15 messages
have been delivered. The crash happens below the Ruby VM: `rescue Exception`
does not catch it, and no Ruby-level error is surfaced — the process simply
dies.

The effect is cumulative, not rate-dependent: delivering 5 messages survives
reliably, while 15 or 30 messages crash reproducibly regardless of how fast
they are published. Once a subscription is active, any process is on a countdown
to a segfault as messages arrive.

Pub/sub is documented as experimental (see #135), but since this is a crash /
memory-safety issue rather than a missing feature, filing it as a bug.

## Steps to reproduce

Requires a Valkey server on `127.0.0.1:6379`.

```ruby
require "valkey"

sub = Valkey.new(protocol: :resp3)
pub = Valkey.new(protocol: :resp3)

sub.subscribe("segv:ch")
sleep 0.3

30.times { |i| pub.publish("segv:ch", "m#{i}") }
# Process receives each message ("PubSub received ..." is printed),
# then dies with SIGSEGV partway through delivery.

sleep 1.0
puts "SURVIVED"   # never reached with 30 messages; reached with 5
```

Observed:

```
$ ruby repro.rb; echo "exit: $?"
PubSub received on channel 'segv:ch': m0
PubSub received on channel 'segv:ch': m1
...
[1]  segmentation fault  ruby repro.rb
exit: 139
```

- `MESSAGES=5` → survives, prints `SURVIVED`
- `MESSAGES=15` → crashes reproducibly
- `MESSAGES=30` → crashes reproducibly
- Wrapping everything in `begin ... rescue Exception` changes nothing — the
  fault is not visible to Ruby code.

The macOS crash report shows the crashing thread is the native GLIDE
background thread delivering the push notification, not the Ruby main thread.

## Expected behavior

Receiving pub/sub messages never terminates the process. Either messages are
delivered safely for as long as the subscription lives, or — if pub/sub is not
ready for use — `subscribe` fails with a Ruby exception instead of arming a
delayed crash.

## Actual behavior

The process is killed by `SIGSEGV` after roughly 15 delivered messages.
Nothing is rescuable, no error callback fires, and the parent shell only sees
[...]

PR #316, opened 2026-09-11
feat(pubsub): pattern subscriptions, lazy variants and callback delivery

Aryex:
### Summary

This implements the advanced pubsub APIs per part 3 of the [implementation plan](https://github.com/valkey-io/valkey-glide-ruby/issues/135#issuecomment-5485707606).

### Issue link

Related #135

### Changes

The following APIs are added

| Method | Request type |
|---|---|
| `psubscribe(*patterns, timeout_ms: 0)` | `PSUBSCRIBE_BLOCKING` (916) |
| `punsubscribe(*patterns, timeout_ms: 0)` | `PUNSUBSCRIBE_BLOCKING` (917) |
| `subscribe_lazy(*channels)` | `SUBSCRIBE` (911) |
| `unsubscribe_lazy(*channels)` | `UNSUBSCRIBE` (913) |
| `psubscribe_lazy(*patterns)` | `PSUBSCRIBE` (901) |
| `punsubscribe_lazy(*patterns)` | `PUNSUBSCRIBE` (908) |

Callback subscriptions is also added:

```ruby
Valkey.new(protocol: :resp3, pubsub: {
  subscriptions: { exact: ["news"] },
  callback: ->(message, context) { context.handle(message) },
  context: my_app_state
})
```

A callback whose arity is exactly `1` receives `(message)`; every other arity, including variadic, receives `(message, context)`.

Beyond the new verbs:

* `PubSubReceiver.make` validates the Pub/Sub options at connect time. A `callback:` that does not answer `#call` or `#arity`, and a `context:` without a `callback:`, now raise `InvalidClientOptionError` instead of being accepted and then silently dropping every message inside the FFI handler's rescue.
* `get_pubsub_message` / `try_get_pubsub_message` raise `InvalidClientOptionError` in callback mode rather than blocking on a queue nothing writes to.
* Delivery stops once the receiver is closed, so a push arriving after `close` no longer pushes onto a closed queue.
* `create_client` and `create_client_from_uri` are declared `blocking: true`. They previously held the GVL for the whole connect call, stalling every other Ruby thread until it returned. This affects all clients, not only Pub/Sub ones, and has its own `### Fixes` entry in the changelog.

### Limitations

* The six sharded verbs and `publish(sharded: true)` remain `NotImplementedError` (Part 4).
* `get_subscriptions`, the five `pubsub_*` introspection commands, and `pubsub(subcommand, *args)` remain `NotImplementedError` (Part 5), so the lazy verbs' documented "read back `get_subscriptions` to confirm" workflow is not yet available.
* The callback runs synchronously on the Rust push thread with the GVL borrowed, matching the other GLIDE clients. A slow callback stalls the messages behind it and the rest of the Ruby runtime; one that re-enters its own client deadlocks until the request timeout. A dispatcher thread was tried and reverted; the constraint is documented on the `callback:` option instead.
* A connection-time subscription can deliver before `Valkey.new` returns, so a callback must not reference the client being constructed. Pass anything it needs through `context:`. Documented, not enforced.
* The FFI handler's rescue still spans both the payload copy and the user callback, so a client defect and a user exception are indistinguishable and both vanish until a logger binding exists.

### Testing

* `bundle exec rubocop` — 126 files, no offenses.
* `bundle exec rake test:unit` — 401 tests, 903 assertions, 0 failures.
* New unit coverage in `test/unit/glide/pubsub_receiver_test.rb` (every branch of `make`, arity dispatch, close semantics) and `test/unit/glide/pubsub_test.rb`; new integration coverage in `test/integration/valkey/pubsub_test.rb`. The delivery tests read through a bounded `queue_pop` helper so a callback path that stops firing fails the suite instead of hanging it.

### Checklist


lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -77,10 +114,12 @@ def build_ffi_handler
 
       # Single delivery point, so push mode is added by branching here and
       # nothing else changes.
-      #
-      # TODO: unfinished -- callback branch:
-      #   @callback.arity == 1 ? call(msg) : call(msg, @context).
       def deliver(message)
+        if @callback
```
jamesx-improving: This runs user code in the place the comment above forbids it ("No user code, no FFI re-entry, no blocking I/O"), and for a SyncClient that thread is a `worker_threads(1)` runtime that also drives the connection (`ffi/src/lib.rs:1415-1427`), so a blocking callback starves the client itself: I measured `sub.get` from the main thread taking 3.00s behind a 3s callback, and a callback that calls back into its own client deadlocks until the request timeout and then reports `ServerUnresponsive` / "timed out".

**Suggested fix:** keep `deliver` enqueueing and dispatch the callback from one consumer thread started in `initialize`. This adds no copy — the payload is already three `read_string`s into Ruby Strings at :104-107, and has to be, since Rust frees the buffers the moment the handler returns (`ffi/src/lib.rs:1350-1363`), so handing the raw pointers to a thread would be a use-after-free; the only cost is a queue push plus a wakeup, which inline mode already pays, and one consumer keeps ordering. If it lands later, document on `callback:` that it must not block or re-enter the client.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -77,10 +114,12 @@ def build_ffi_handler
...
+          @callback.arity == 1 ? @callback.call(message) : @callback.call(message, @context)
```
jamesx-improving: `arity` is not part of the `#call` contract that `make` validates, so a plain callable object (`class Handler; def call(msg, ctx)`) is accepted and then loses every message: `NoMethodError: undefined method 'arity'` is swallowed by the handler's `rescue StandardError`, and `check_callback!` has already closed the inline path — no delivery, no queue, no error. Listing it under Limitations leaves users of a documented input shape with silent total message loss.

**Suggested fix:** resolve the arity once in the constructor with `callback.respond_to?(:arity) ? callback.arity : callback.method(:call).arity`, and validate it in `make` so a bad callable raises at connect time. A unit test with a non-proc callable would pin it.

test/unit/glide/pubsub_receiver_test.rb
```diff
@@ -56,6 +56,168 @@ def test_ffi_handler_is_retained
...
+
+    push(Kind::MESSAGE, message: "exact", channel: "news")
+    push(Kind::PMESSAGE, message: "pattern", channel: "news.tech", pattern: "news.*")
+
+    expected = [
+      ["exact", "news", nil],
+      ["pattern", "news.tech", "news.*"]
+    ]
+
+    delivered = 2.times.map { received.pop.to_a }
```
jamesx-improving: These six delivery tests (`:77, :99, :115, :128, :146, :192`) read with an unbounded `Thread::Queue#pop`, so a broken callback path hangs instead of failing — I disabled the callback branch in `deliver` and the suite (0.0006s normally) hung until I killed it at 40s on three of four seeds.

**Suggested fix:** wrap each blocking read in `Timeout.timeout(2) { ... }`, as `pubsub_test.rb:390` already does for the same reason (`Queue#pop(timeout:)` needs Ruby 3.2; the gemspec says >= 3.0).

jamesx-improving (changes requested):
The Testing section is stale — it says integration coverage is deliberately absent and that cluster was not run, but this PR adds 8 integration tests that ran and passed in cluster CI at d209ace (`TestClusterPubSub` 24/24). Worth correcting, since it changes how a reviewer weighs the change; the two blocking items are inline on the callback path.

Happy to re-review once addressed.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -51,6 +82,12 @@ def close
 
       private
 
+      def check_callback!
+        return unless callback_mode?
+
+        raise CommandError, "Pub/Sub callback was configured. Inline Pub/Sub reads are unavailable."
```
nderraugh: `CommandError` is the base for server-side command failures (`WrongTypeError`, `PermissionError`, `ExecAbortError` derive from it), so a caller who rescues `CommandError` around `get_pubsub_message` to handle server errors will also swallow "you configured a callback", which is a client-side configuration state. `Valkey::InvalidClientOptionError` already exists (it's what `Resp3RequiredError` derives from) and fits here.

`test_pop_raises_in_callback_mode` / `test_try_pop_raises_in_callback_mode` currently pin `CommandError`, so they'd flip with it. The public YARD on `get_pubsub_message` / `try_get_pubsub_message` also doesn't declare that they raise in callback mode.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -26,19 +28,48 @@ module PushKind
         MESSAGE_KINDS = [MESSAGE, PMESSAGE, SMESSAGE].freeze
       end
 
-      def initialize
+      def self.make(pubsub_configs: {})
+        configs = pubsub_configs || {}
+        callback = configs[:callback]
+        context = configs[:context]
```
nderraugh, line 34: `context:` without `callback:` is silently accepted. The plan in #135 says `context:` "Requires callback:", and a typo like `callbck:` currently yields a queue-mode client whose context is dead weight with nothing surfaced. Suggest raising `ArgumentError` here when `context` is non-nil and `callback` is nil. `test_context_without_callback_builds_queue_mode_receiver` asserts the current behavior and would need to flip.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -77,10 +114,12 @@ def build_ffi_handler
 
       # Single delivery point, so push mode is added by branching here and
       # nothing else changes.
-      #
-      # TODO: unfinished -- callback branch:
-      #   @callback.arity == 1 ? call(msg) : call(msg, @context).
       def deliver(message)
```
nderraugh: There is no delivery barrier after `close`. `#close` only closes the queue, so in callback mode `deliver` still invokes the callback afterwards (confirmed at the receiver level: `close` then `deliver` fires). On the native side `close_client` only decrements the `ClientAdapter` Arc and the push worker keeps its `callback_store`, so pushes already buffered in `push_rx` can reach the Ruby callback while `Valkey#close` is mid-flight, after `@connection` has been nil'd. A `@closed` flag set in `close` and checked here is a one-liner and gives a clean "no callbacks after `close` returns" guarantee; worth a unit test.

lib/valkey/glide/pubsub_receiver.rb
nderraugh: The single `rescue StandardError` in `build_ffi_handler` now covers both the pointer-copy step and user code, so a bug in `read_string` / `PubSubMessage.new` and a user exception are indistinguishable and both vanish. Suggest wrapping this `@callback.call` separately so the two failure classes are distinguishable once a logger lands, and so a non-`StandardError` escape from user code (`exit`, `Interrupt`) can't unwind through the FFI trampoline.

lib/valkey.rb
```diff
@@ -328,7 +328,7 @@ def initialize(options = {})
     pubsub_config = parse_pubsub_configs(options[:pubsub], protocol: options[:protocol])
     json_options.merge!(pubsub_config)
 
-    @pubsub_receiver = Glide::PubSubReceiver.new
+    @pubsub_receiver = Valkey::Glide::PubSubReceiver.make(pubsub_configs: options[:pubsub])
```
nderraugh, line 331: With `create_client_from_uri` now `blocking: true`, config-time subscriptions can deliver during connect, so the callback can fire before `Valkey.new` returns. `client = Valkey.new(pubsub: { callback: ->(m, _) { client.publish(...) } })` sees `client == nil` inside the callback and the `NoMethodError` is swallowed by the handler's rescue. Not a bug, but it should be documented on `callback:`, pointing users at `context:` for state the callback needs.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
   * `with_sort_keys:` cannot be combined with `no_content:` — the server omits the sort keys under `NOCONTENT`, so the reply cannot be parsed reliably.
   * `ft_search` is unsupported inside `pipelined`/`multi` (a queued command yields a `Valkey::Future`, not a reply) and with the `flatten_map: true` compatibility option; both raise `ArgumentError`. Use `call` in those cases.
 * Ruby: PubSub: added support for `subscribe(*channels, timeout_ms:)`, `unsubscribe(*channels, timeout_ms:)`, `publish(message, channel)`, `get_pubsub_message`, `try_get_pubsub_message`, and the connection-time `pubsub: { subscriptions: { exact: [...] } }` option. Requires `protocol: :resp3` ([#308](https://github.com/valkey-io/valkey-glide-ruby/pull/308))
+* Ruby: PubSub: added support for `psubscribe(*patterns, timeout_ms:)`, `punsubscribe(*patterns, timeout_ms:)`, `subscribe_lazy(*channels)`, `unsubscribe_lazy(*channels)`, `psubscribe_lazy(*patterns)`, `punsubscribe_lazy(*patterns)`, and the `pubsub: { callback:, context: }` connection option for callback mode delivery ([#316](https://github.com/valkey-io/valkey-glide-ruby/pull/316))
```
nderraugh, line 32: The `blocking: true` added to both `create_client*` bindings isn't mentioned here or in the PR description. It means other Ruby threads now keep running during connect, which affects every client, not just Pub/Sub. It's the right change (AGENTS.md requires it), but it deserves a line.

lib/valkey/commands/pubsub_commands.rb
```diff
@@ -96,11 +95,16 @@ def unsubscribe(*channels, timeout_ms: 0)
...
       # @raise [ArgumentError] if timeout_ms is negative
+      # @raise [Valkey::Resp3RequiredError] GLIDE Pub/Sub requires RESP3
       # @raise [Valkey::TimeoutError] if the timeout expires before the server confirms
-      # @raise [NotImplementedError] this method is not implemented yet
       #
       # @see https://valkey.io/commands/psubscribe/
-      def psubscribe(*patterns, timeout_ms: 0) = raise(NotImplementedError, "#{__method__} is not implemented yet")
+      def psubscribe(*patterns, timeout_ms: 0)
+        validate_resp3!
+        raise ArgumentError, "No channels provided for subscription" if patterns.empty?
```
nderraugh: nits: `psubscribe` / `psubscribe_lazy` (via `send_lazy_subscription` at :457) report "No channels provided for subscription" for an empty pattern list. It matches the glide-core string, but this check is client-owned; "No patterns provided" reads better for the pattern verbs.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -26,19 +28,48 @@ module PushKind
...
+      def self.make(pubsub_configs: {})
+        configs = pubsub_configs || {}
+        callback = configs[:callback]
+        context = configs[:context]
+
+        if !callback.nil? && !callback.respond_to?(:call)
+          raise ArgumentError, "Pub/Sub callback: must respond to #call, got: #{callback.class}"
+        end
+
+        Glide::PubSubReceiver.new(callback: callback, context: context)
```
nderraugh: nits: `Glide::PubSubReceiver.new` inside `PubSubReceiver.make` can be `new`.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -26,19 +28,48 @@ module PushKind
...
+        @context = context
         @message_queue = Thread::Queue.new
 
         @ffi_handler = build_ffi_handler
       end
 
       attr_reader :ffi_handler
 
+      # @return [Boolean] whether messages go to a callback rather than the queue.
+      def callback_mode?
```
nderraugh, line 70: nits: `callback_mode?` is public on an `@api private` class; fine, just noting it's now load-bearing for tests.

test/unit/glide/pubsub_receiver_test.rb
```diff
@@ -56,6 +56,168 @@ def test_ffi_handler_is_retained
...
+    @receiver = Valkey::Glide::PubSubReceiver.new(callback: callback)
+
+    push(Kind::MESSAGE, message: "first", channel: "news")
+    push(Kind::MESSAGE, message: "second", channel: "news")
+
+    assert_equal "second", received.pop.message
+    assert_empty received
+  end
+
+  def test_pop_raises_in_callback_mode
```
nderraugh, line 166: Coverage gap: nothing tests the callback-mode rejection through the public client (`Valkey#get_pubsub_message` / `#try_get_pubsub_message`); only the receiver is covered here.

test/unit/glide/pubsub_receiver_test.rb
```diff
@@ -56,6 +56,168 @@ def test_ffi_handler_is_retained
...
+
+    push(Kind::MESSAGE, message: "exact", channel: "news")
+
+    message, delivered_context = received.pop
+
+    assert_equal "exact", message.message
+    assert_equal :app_state, delivered_context
+  end
+
+  def test_raising_callback_does_not_propagate_and_does_not_kill_the_receiver
```
nderraugh, line 134: Coverage gap: no test for post-`close` behavior in callback mode (see the comment on `deliver`). Today `close` followed by `deliver` still invokes the callback.

test/integration/valkey/pubsub_test.rb
```diff
@@ -229,6 +229,152 @@ def test_subscribed_client_still_runs_ordinary_commands
...
+        r.publish("pattern-msg", channel)
+        msg = wait_for_message(subscriber)
+
+        assert_equal "pattern-msg", msg.message
+        assert_equal channel,       msg.channel
+        assert_equal pattern,       msg.pattern
+      end
+    end
+
+    def test_punsubscribe_stops_pattern_delivery
```
nderraugh, line 247: Coverage gap: no integration test for the no-argument "unsubscribe from all" forms of `punsubscribe` / `unsubscribe_lazy` / `punsubscribe_lazy`. That is the one branch in glide-core's `handle_lazy_subscription` (`channels_set = None`) the current tests never reach.

test/integration/valkey/pubsub_test.rb
```diff
@@ -229,6 +229,152 @@ def test_subscribed_client_still_runs_ordinary_commands
...
+        subscriber.psubscribe_lazy(pattern)
+        publish_until_received("before", channel, subscriber)
+
+        subscriber.punsubscribe_lazy(pattern)
+
+        assert_delivery_ceased(channel, subscriber)
+      end
+    end
+
+    def test_callback_mode_delivers_messages_end_to_end
```
nderraugh, line 383: Coverage gap: callback mode is only exercised with an exact subscription. A pattern subscription in callback mode (asserting `msg.pattern`) would close the loop on the `PMESSAGE` path through `deliver`.

test/unit/glide/pubsub_test.rb
```diff
@@ -221,6 +221,154 @@ def test_unsubscribe_rejects_a_negative_timeout
...
+  end
+
+  def test_punsubscribe_rejects_a_negative_timeout
+    client = RecordingClient.new
+
+    assert_raises(ArgumentError) { client.punsubscribe("news.*", timeout_ms: -0.5) }
+    assert_empty client.sent_commands
+  end
+
+  def test_lazy_verbs_dispatch_without_a_timeout_argument
```
nderraugh, line 306: Coverage gap: no test that a lazy verb returns `nil` and does not block; these only assert request type and args.

lib/valkey.rb
nderraugh: This one points (again) at what I think is a design flaw: we should align with other FFI consuming wrappers in how we create new clients, while still avoiding pulling in the proto deps.

Might be worth queueing up an issue on that.

lib/valkey/glide/pubsub_receiver.rb
Aryex: Addressed in [240a3c5](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/240a3c5d17a8383c6f69f49c08ee910c4a2978d0)

test/unit/glide/pubsub_receiver_test.rb
Aryex: Addressed in [240a3c5](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/240a3c5d17a8383c6f69f49c08ee910c4a2978d0)

test/integration/valkey/pubsub_test.rb
Aryex: Addressed in [240a3c5](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/240a3c5d17a8383c6f69f49c08ee910c4a2978d0)

test/integration/valkey/pubsub_test.rb
Aryex: Addressed in [240a3c5](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/240a3c5d17a8383c6f69f49c08ee910c4a2978d0)

test/unit/glide/pubsub_test.rb
Aryex: Addressed in [240a3c5](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/240a3c5d17a8383c6f69f49c08ee910c4a2978d0)

lib/valkey/glide/pubsub_receiver.rb
Aryex: Addressed in [7f96d88](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/7f96d882bf6b42e217ff40faea5372d58332390d) One caveat on the guarantee: the flag stops a callback from starting once `close` is observed, but a callback already in flight is not waited for, so user code can still be running after `close` returns. Holding a mutex across the check and the call would close that, and would deadlock against exactly the re-entrant callback in https://github.com/valkey-io/valkey-glide-ruby/pull/316#discussion_r4021293635, so the full join belongs with that change.

test/unit/glide/pubsub_receiver_test.rb
Aryex: Addressed in [7f96d88](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/7f96d882bf6b42e217ff40faea5372d58332390d)

lib/valkey/glide/pubsub_receiver.rb
Aryex: Addressed in [0a45283](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/0a4528336c82aa29872d13f81d5b741c9fd6c1b3) One deliberate difference from the suggestion: rather than resolving the arity via `callback.method(:call).arity`, `make` now requires the callback to answer `#arity` and rejects it otherwise, so a bare object defining only `#call` fails when the client is created instead of being supported. `Proc`, lambdas and `Method` objects all qualify.

lib/valkey/glide/pubsub_receiver.rb
Aryex: Addressed in [0a45283](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/0a4528336c82aa29872d13f81d5b741c9fd6c1b3)

test/unit/glide/pubsub_receiver_test.rb
Aryex: Addressed in [89e14ef](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/89e14ef3f50feb118329a65b153c69f35f7b991b). Reproduced your experiment first: with the callback branch of `deliver` disabled the file hung until killed, and now fails in 14s with 1 failure and 7 errors. All seven reads go through a `queue_pop(queue, timeout = 2)` helper, which keeps the reason in one place and leaves a single line to change when the gemspec's Ruby floor reaches 3.2 and `Queue#pop(timeout:)` becomes available.

lib/valkey/commands/pubsub_commands.rb
Aryex: Addressed in [700dd00](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/700dd00d7cf504794a791a784c78ffeb753b51c7) `send_lazy_subscription` takes the noun as an argument now, so `psubscribe` and `psubscribe_lazy` say "No patterns provided for subscription" while the channel verbs keep their wording.

CHANGELOG.md
Aryex: Addressed in [700dd00](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/700dd00d7cf504794a791a784c78ffeb753b51c7)

lib/valkey/glide/pubsub_receiver.rb
Aryex: Addressed in [700dd00](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/700dd00d7cf504794a791a784c78ffeb753b51c7)

lib/valkey.rb
Aryex: Agree on aligning with other clients. Would be good to look into how `create_client_with_uri` diverges.

That said, I'm not sure how using `context` here would fix the example provided. Isn't it more like a "don't reference things that aren't ready".

lib/valkey/glide/pubsub_receiver.rb
Aryex: updated in [b70c13b](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/b70c13be2070326da2bbeb66696c33fa4e950778)

lib/valkey/glide/pubsub_receiver.rb
Aryex: Applied in [b70c13b](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/b70c13be2070326da2bbeb66696c33fa4e950778)

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -26,31 +28,96 @@ module PushKind
...
+      # Handling user callback on a separate threads. This avoids having long running callbacks
+      # taking up the GVL and locking up the Ruby runtime.
+      def start_callback_thread
+        Thread.new do
+          while (message = @message_queue.pop)
+            break if @closed
+
+            begin
+              @callback.arity == 1 ? @callback.call(message) : @callback.call(message, @context)
+            rescue StandardError
```
jamesx-improving: Blocking, and it's a regression introduced by b70c13b rather than something pre-existing: this rescue only covers `StandardError`, so a callback raising anything else kills the consumer thread for good. The full chain, measured at b70c13b:

- the thread dies (`alive?=false`) and nothing restarts it
- the handler keeps enqueueing at `:138` with nothing draining - messages 3 and 4 never delivered, 2 left sitting in a queue that grows unbounded
- `check_callback!` has already closed the inline read path, so there's no channel and no programmatic signal (just a `report_on_exception` line on stderr)
- `close`'s bare `join` at `:94` re-raises the user's exception out of `Valkey#close`, which skips `Bindings.close_client`; `@connection` is already nil by then, so the handle is unreachable and never freed

Reachable from this gem's own API: the five `pubsub_*` introspection commands and `get_subscriptions` raise `NotImplementedError`, so a callback that introspects its own subscriptions is enough to trigger it. The same probe at 700dd00 delivered both later messages and closed normally, which is what makes this new rather than latent. CI is green because `test_callback_exception_is_contained` uses a bare `raise`, i.e. a `RuntimeError`, which is a `StandardError`.

Credit where it's due - @nderraugh named this hazard class first in https://github.com/valkey-io/valkey-glide-ruby/pull/316#discussion_r4026988082 ("so a non-`StandardError` escape from user code ... can't unwind through the FFI trampoline"). What's new is that moving the call onto a thread relocated the hazard rather than removing it, plus the three consequences above.

**Suggested fix:** rescue `Exception` here, re-raising only `SystemExit`/`Interrupt`/`SignalException`; guard the `join` so a dead thread can't raise out of `close`; and move `Bindings.close_client` into the `ensure` in `lib/valkey.rb`. A unit test with a `NotImplementedError` callback asserting a later message still arrives would pin all three.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -26,31 +28,96 @@ module PushKind
...
+      # @param context [Object, nil] second argument for a callback whose arity
+      #   is not 1.
+      def initialize(callback: nil, context: nil)
+        @callback = callback
+        @context = context
         @message_queue = Thread::Queue.new
+        @closed = false
 
         @ffi_handler = build_ffi_handler
+        @callback_thread = start_callback_thread if callback
```
jamesx-improving: This thread is only reaped by `close`, and its block closes over the receiver, so a live thread roots the receiver and the `@ffi_handler` graph - a callback-mode client dropped without `close` leaks a thread permanently. Measured: 10 created-and-dropped clients leave 11 live threads blocked at `:103` after three `GC.start` cycles. The unit suite does this too, since every `RecordingClient.new(callback:)` is never closed.

**Suggested fix:** `ObjectSpace.define_finalizer` on the receiver, closing the queue so the thread's `pop` returns nil and it exits - pass the queue in rather than capturing `self`, or the finalizer never runs. A test asserting the thread count returns to baseline after `close` would pin it.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -26,31 +28,96 @@ module PushKind
...
       end
 
       private
 
+      # Handling user callback on a separate threads. This avoids having long running callbacks
+      # taking up the GVL and locking up the Ruby runtime.
+      def start_callback_thread
+        Thread.new do
+          while (message = @message_queue.pop)
+            break if @closed
```
jamesx-improving: This drops the whole queued backlog on `close`, including the message it just popped - measured 4 of 5 discarded, with no error and no count - and it diverges from queue mode, where a closed `Thread::Queue` can still be drained via `try_pop`.

**Suggested fix:** delete the `break`. `@message_queue.close` already makes `pop` drain the backlog and *then* return nil, so the loop exits on its own, and because `close` joins the thread those callbacks still finish before `close` returns - which is exactly the guarantee @nderraugh asked for on `deliver`. If dropping the backlog is intended instead, it needs saying on `close` and in the CHANGELOG, since it makes a clean shutdown at-most-once.

jamesx-improving (changes requested):
All three of my findings are resolved, and I verified them by re-running rather than reading the diff:

- Push-worker starvation is gone: main-thread `GET` behind a 3s callback went 3.00s -> 0.00s, and the re-entrant callback went deadlock/`ServerUnresponsive` -> ok in 4.7ms. Ordering held 20/20 behind a 10ms callback, so it's one consumer rather than a thread per message.
- Two things that were easy to get wrong here and are right: the three `read_string` copies stayed on the FFI-handler side with only the materialised `Struct` crossing to the thread, so there's no use-after-free and #257's lifetime region is untouched; and `close` joins the in-flight callback (2s callback -> `close` returned at 2.01s), with a self-join guard so `close` called from inside a callback returns instead of hanging.
- That also closes @nderraugh's barrier finding (https://github.com/valkey-io/valkey-glide-ruby/pull/316#discussion_r4026988073) - the join you said "belongs with that change" now has nothing outstanding behind it.

What's blocking is a regression the fix introduced, not anything pre-existing: a callback raising a non-`StandardError` kills the consumer thread. That one is inline, with two smaller items.

Two non-code notes: the new threading semantics (dedicated thread, ordering, `close` joins, backlog dropped) deserve a CHANGELOG line, and the PR body's third Limitation is stale - it still describes the `#arity` silent-drop as live and names a `validate_pubsub_callback!` that isn't in the tree.

Aryex:
[cca467a](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/cca467aa85d4f1a5c7374f39b5161b5d7fdd816a) reverted the callback thread implementation. While it is a good improvements, there are some life cycle issues that came up as a result. Since this is an optimization, I'm reverting back to the old callback approach and will defer this improvement for the future.

lib/valkey.rb
nderraugh: Yeah, correct context doesn't fix anything. I meant a comment on the code:
```
# @note With connection-time subscriptions (pubsub: { subscriptions: ... }),
#   the callback may be invoked during connect — before Valkey.new returns.
#   Do not reference the client being constructed from inside the callback
#   (it is still nil at that point); any state the callback needs at delivery
#   time should be passed via context: instead.
```

...until we come up with a design that mitigates it.

Aryex: Updated in [d8789a4](https://github.com/valkey-io/valkey-glide-ruby/pull/316/commits/d8789a44f4fec51ce094d3b6338904f0c263861b)

nderraugh (approved):
lgtm, other than Limitation + Testing sections in the PR description which need a refresh.

lib/valkey/commands/pubsub_commands.rb
```diff
@@ -23,10 +24,28 @@ module Commands
...
+    #   valkey = Valkey.new(
+    #     protocol: :resp3,
+    #     pubsub: {
+    #       subscriptions: { exact: ["news"], pattern: ["news.*"] },
+    #       callback: ->(message, context) { context[:messages] << [message.channel, message.message] },
+    #       context: { messages: [] }
+    #     }
+    #   )
+    #
+    # @note While the callback executes it holds the GVL, blocking the Ruby runtime until it
```
jamesx-improving, line 40: This frames the cost as slowness, but a callback that calls back into its own client doesn't run slow — it fails, with `Valkey::TimeoutError` after the request timeout and `ServerUnresponsive` in the log. By the "avoid costly operations" standard a single `client.set` reads as cheap.

lib/valkey/glide/pubsub_receiver.rb
```diff
@@ -26,63 +28,104 @@ module PushKind
...
       # Builds the proc handed to the FFI.
       #
       # Runs on a Rust thread the Ruby runtime did not create, on a single push
-      # worker, with the GVL borrowed. Keep it thin: copy out, enqueue, return.
-      # No user code, no FFI re-entry, no blocking I/O -- anything slow here
-      # stalls every message behind it. The pointers are freed when it returns,
-      # so the copy has to happen synchronously.
+      # worker, with the GVL borrowed. The pointers are freed when it returns, so
+      # the copy has to happen synchronously. A callback runs here rather than on a
+      # dispatcher thread, matching the other GLIDE clients, so a slow callback
```
jamesx-improving, line 111: Small factual thing, since this comment will outlive the PR: Python sync does dispatch inline (`glide-sync/glide_sync/glide_client.py:284`), but Go copies and then hands off to `go func()` (`go/callbacks.go:152-159`), so it matches one of the two rather than both.

test/unit/glide/pubsub_receiver_test.rb
```diff
@@ -56,8 +62,205 @@ def test_ffi_handler_is_retained
...
+    expected = [
+      ["exact", "news", nil],
+      ["pattern", "news.tech", "news.*"]
+    ]
+
+    delivered = 2.times.map { queue_pop(received).to_a }
+
+    assert_equal expected, delivered
+
+    assert_raises(Valkey::InvalidClientOptionError) { @receiver.pop }
```
jamesx-improving, line 87: `test_callback_mode_leaves_the_queue_empty` went away in b70c13b when the consumer thread made it false by design, and cca467a made it true again without a test coming back — so flipping the ternary at `pubsub_receiver.rb:123` in either direction now passes the suite. That's the one gap here that isn't cosmetic.

jamesx-improving (approved):
Nothing here blocks — three inline notes on the diff, plus two things that don't anchor to a line:

The deferred dispatcher thread doesn't appear to be tracked anywhere — #135's checklist is API surface, and #309 is the `blocking: true` task.

Agreed with @nderraugh on the PR description: the third Limitation still describes the `#arity` drop as live and names a `validate_pubsub_callback!` that isn't in the tree.

PR #317, opened 2026-09-15
feat(pubsub): sharded Pub/Sub for cluster mode

Aryex:
### Summary

Adds sharded Pub/Sub support for cluster clients through `SPUBLISH`, `SSUBSCRIBE`, and `SUNSUBSCRIBE`.

### Issue link

Sub task of [#135 Pub/Sub Support](https://github.com/valkey-io/valkey-glide-ruby/issues/135).

### Changes

- Adds blocking and lazy sharded subscribe/unsubscribe methods.
- Adds `publish(message, channel, sharded: true)` and connect-time sharded subscriptions.
- Adds `Valkey#cluster_mode?` and cluster-mode validation for sharded subscriptions.
- Keeps sharded publish batchable in `pipelined` and `multi`.

### Limitations

- Requires cluster mode and Valkey 7.0+.
- Pub/Sub introspection is deferred to Part 5.

### Testing

- `bundle exec rubocop`
- `bundle exec rake test:unit`
- `bundle exec rake test:standalone`
- `bundle exec rake test:cluster`

### Checklist

- [x] This Pull Request is related to an issue.
- [x] Commit message describe your changes
- [x] Commits are signed off (`git commit -s`) per the DCO.
- [x] Tests are added or updated.
- [x] CHANGELOG.md and documentation files are updated.
- [x] Linters have been run (`bundle exec rubocop`) and pass.

PR #320, opened 2026-09-16
feat(pubsub): metadata API

Aryex:
### Summary

Adds Pub/Sub introspection and connection subscription-state APIs. This is Part 5 of the Pub/Sub implementation.

### Issue link

Sub task of [#135 Pub/Sub Support](https://github.com/valkey-io/valkey-glide-ruby/issues/135).

### Changes

- Adds `pubsub_channels`, `pubsub_numpat`, `pubsub_numsub`, `pubsub_shardchannels`, and `pubsub_shardnumsub`.
- Adds `get_subscriptions`, returning desired and server-confirmed subscriptions as `Valkey::Glide::PubSubState`.
- Supports the five `pubsub_*` introspection commands in `pipelined` and `multi`; glide-core handles cluster fan-out and aggregation.
- Enables Pub/Sub lint and integration coverage in the standalone and cluster suites.

### Limitations

- `get_subscriptions` is not supported inside `pipelined` or `multi`.
- In cluster mode, introspection inside an atomic `multi` may reflect only one node; direct and non-atomic pipelined calls aggregate across nodes.
- Non-empty sharded introspection results depend on the sharded subscription support in [#317](https://github.com/valkey-io/valkey-glide-ruby/pull/317).

### Testing

- `bundle exec rubocop`
- `bundle exec rake test:unit`
- `bundle exec rake test:standalone`
- `bundle exec rake test:cluster`

### Checklist


lib/valkey/commands/pubsub_commands.rb
```diff
@@ -385,19 +380,19 @@ def pubsub_numpat = raise(NotImplementedError, "#{__method__} is not implemented
...
-      # @return [Hash{String => Integer}] the channel names mapped to their number of subscribers
-      # @raise [NotImplementedError] this method is not implemented yet
+      # @param [Array<String>] channels the channels to query; omit to query none
+      # @return [Hash{String => Integer}, Array] subscriber counts keyed by channel;
+      #   `flatten_map: true` returns a flat `[channel, count, ...]` array
       #
       # @see https://valkey.io/commands/pubsub-numsub/
-      def pubsub_numsub(*channels) = raise(NotImplementedError, "#{__method__} is not implemented yet")
+      def pubsub_numsub(*channels)
+        send_command(RequestType::PUBSUB_NUM_SUB, channels.map(&:to_s))
```
affonsov: `pubsub_numsub` and `pubsub_shardnumsub` return the core reply unchanged. As currently encoded by the tests, a standalone RESP2 direct call can return a flat Array, while RESP3, cluster, and a pipeline on that same RESP2 connection return a Hash. This makes the public type depend on both connection configuration and invocation style, and diverges from redis-rb's stable Hash result. Please normalize the result across direct and batched calls, while preserving the explicit `flatten_map: true` behavior. If this variation is intentional, the public documentation should describe every case explicitly.

test/integration/valkey/pubsub_test.rb
```diff
@@ -461,6 +461,106 @@ def test_callback_exception_is_contained
...
+      with_client do |subscriber|
+        subscriber.subscribe(channel)
+
+        assert_includes r.pubsub_channels, channel
+        assert_includes r.pubsub_channels("#{channel}*"), channel
+        assert_numsub({ channel => 1 }, r.pubsub_numsub(channel))
+      end
+    end
+
+    def test_pipelined_pubsub_introspection
```
affonsov: Optional if you want to have more test coverage
The PR documents support for introspection inside both pipelined and multi, but the live test only exercises pipelined. Block-form multi uses the separate atomic batch path

lib/valkey/glide/pubsub_state.rb
```diff
@@ -5,14 +5,48 @@ module Glide
     # A connection's subscriptions
     #
     # @!attribute [rw] desired_subscriptions
-    #   @return [Hash{Symbol => Array<String>}] the subscriptions the client
-    #     asked for, keyed `:exact`, `:pattern` and `:sharded`. Standalone
-    #     connections omit `:sharded`.
+    #   @return [Hash{Symbol => Array<String>}] the subscriptions the client asked for, keyed
+    #     `:exact`, `:pattern` and `:sharded`. A mode is present only once the client has
+    #     subscribed in it, so this is `{}` on a fresh connection and loses a key again when the
```
affonsov: The statement that a mode “loses a key again when the last subscription of that mode is dropped” does not match the core behavior. Targeted unsubscribe leaves the mode with an empty set, while unsubscribe-all removes the mode entry. Other GLIDE clients expose these as partial maps and treat a missing mode and an empty set equivalently. Is this missing for these changes or will have a follow-up?.

test/lint/pubsub_commands.rb
```diff
@@ -3,41 +3,45 @@
...
       channels = r.pubsub_channels("test*")
+
       assert_kind_of Array, channels
+      channels.each { |channel| assert_kind_of String, channel }
     end
 
     def test_pubsub_numpat
       # Get number of pattern subscriptions
       count = r.pubsub_numpat
+
```
affonsov, line 30: `pubsub_numpat` is only checked with `count >= 0`, so the integration suite would still pass if the command always returned zero. Pattern subscriptions are already available, so please add an end-to-end test that calls `psubscribe`, verifies `pubsub_numpat` observes the subscription, and then cleans it up.

Aryex: added in [4c8dc09](https://github.com/valkey-io/valkey-glide-ruby/pull/320/commits/4c8dc09c6747d6d5d15d1634b9af0a409dcfb000)

lib/valkey/glide/pubsub_state.rb
Aryex: updated in [4c8dc09](https://github.com/valkey-io/valkey-glide-ruby/pull/320/commits/4c8dc09c6747d6d5d15d1634b9af0a409dcfb000)

test/integration/valkey/pubsub_test.rb
Aryex: Pipeline support was not intended for this PR so I have removed it in [9e6332f](https://github.com/valkey-io/valkey-glide-ruby/pull/320/commits/9e6332f0c07d0c4fd70c68e23098e4b809fa1036)

lib/valkey/commands/pubsub_commands.rb
```diff
@@ -385,19 +380,19 @@ def pubsub_numpat = raise(NotImplementedError, "#{__method__} is not implemented
       #   valkey.pubsub_numsub
       #     # => {}
       #
-      # @param [Array<String>] channels the channels to query for the number of subscribers; an empty list
-      #   returns an empty hash
-      # @return [Hash{String => Integer}] the channel names mapped to their number of subscribers
-      # @raise [NotImplementedError] this method is not implemented yet
+      # @param [Array<String>] channels the channels to query; omit to query none
+      # @return [Hash{String => Integer}, Array] subscriber counts keyed by channel;
+      #   `flatten_map: true` returns a flat `[channel, count, ...]` array
```
nderraugh: `flatten_map: true` documented but not implemented.

lib/valkey/commands/pubsub_commands.rb
```diff
@@ -426,13 +421,14 @@ def pubsub_shardchannels(pattern = nil) = raise(NotImplementedError, "#{__method
       #   valkey.pubsub_shardnumsub
       #     # => {}
       #
-      # @param [Array<String>] channels the sharded channels to query for the number of subscribers; an empty
-      #   list returns an empty hash
-      # @return [Hash{String => Integer}] the sharded channel names mapped to their number of subscribers
-      # @raise [NotImplementedError] this method is not implemented yet
+      # @param [Array<String>] channels the shard channels to query; omit to query none
+      # @return [Hash{String => Integer}, Array] subscriber counts keyed by shard channel;
+      #   `flatten_map: true` returns a flat `[channel, count, ...]` array
```
nderraugh: flatten_map: true documented but not implemented.

test/lint/pubsub_commands.rb
```diff
@@ -86,10 +94,11 @@ def test_spublish
       # SPUBLISH was introduced in Redis 7.0.
       # Skipped on Redis 6.2 and earlier versions.
       omit_version("7.0")
-      # Publish to a shard channel with no subscribers
-      result = r.spublish("test_shard", "Hello, Shard!")
-      assert_kind_of Integer, result
-      assert result >= 0
+      result = r.publish("Hello, Shard!", "lint_shard_no_subscribers", sharded: true)
```
nderraugh, line 96: test_spublish is testing publish, was testing spublish?

publish raises NotImplementedError whenever sharded: true (Part 4 not done). The test rescues that and skips. So assert_equal 0, result is dead code that can never run today, and there is now no coverage of the spublish path (the old r.spublish(...) call is gone). 

Is this a forward-looking placeholder?

Might be worth some attention or a note.

Aryex: Updated in[c0c5487](https://github.com/valkey-io/valkey-glide-ruby/pull/320/commits/c0c54875771fe5aa77b0188057496c3769325dae). 

It was overfitting because the method was not implemented. Updated now as the PR was just merged.

lib/valkey/commands/pubsub_commands.rb
```diff
@@ -385,19 +380,19 @@ def pubsub_numpat = raise(NotImplementedError, "#{__method__} is not implemented
...
-      #   returns an empty hash
-      # @return [Hash{String => Integer}] the channel names mapped to their number of subscribers
-      # @raise [NotImplementedError] this method is not implemented yet
+      # @param [Array<String>] channels the channels to query; omit to query none
+      # @return [Hash{String => Integer}, Array] subscriber counts keyed by channel;
+      #   `flatten_map: true` returns a flat `[channel, count, ...]` array
       #
       # @see https://valkey.io/commands/pubsub-numsub/
-      def pubsub_numsub(*channels) = raise(NotImplementedError, "#{__method__} is not implemented yet")
+      def pubsub_numsub(*channels)
```
nderraugh, line 404: Positional vs. keyword. In Ruby, foo(x: 1) is not a named positional argument. Keyword arguments are a separate channel, and a method only receives them if its signature declares them (x: or **opts). If it declares none, Ruby falls back to bundling the pairs into a plain Hash and passing it as the last positional argument.

The docstring is the contract a caller reads. A developer sees `flatten_map: true` documented on pubsub_numsub, uses it in good faith, and gets no error — because *channels declares no keywords, the Hash is scooped into the splat and stringified:
```ruby
pubsub_numsub("a", flatten_map: true)
# channels == ["a", { flatten_map: true }]
# .map(&:to_s) => ["a", "{:flatten_map=>true}"]   # sent as a channel name
```
The call succeeds, the reply comes back in the flat shape they expected (the server always returns that shape), and a phantom channel "{:flatten_map=>true}" appears in the result with a count of 0. Nothing tells them the option did nothing — they'll ship code believing a feature works and only notice when the extra entry surprises someone downstream. Same for pubsub_shardnumsub.

Mitigation: The option was never implemented here (it belongs to the vector-search commands). Delete the flatten_map sentence from both @return docstrings so the contract matches the signature. But _general rule:_ declare every keyword you accept, so an unexpected one raises ArgumentError: unknown keyword rather than degrading into data.

lib/valkey/commands/pubsub_commands.rb
```diff
@@ -402,19 +397,18 @@ def pubsub_numpat = raise(NotImplementedError, "#{__method__} is not implemented
       #   valkey.pubsub_numsub
       #     # => {}
       #
-      # @param [Array<String>] channels the channels to query for the number of subscribers; an empty list
-      #   returns an empty hash
-      # @return [Hash{String => Integer}] the channel names mapped to their number of subscribers
-      # @raise [NotImplementedError] this method is not implemented yet
+      # @param [Array<String>] channels the channels to query; omit to query none
+      # @return [Hash{String => Integer}, Array] subscriber counts keyed by channel;
```
nderraugh:
```suggestion
      # @return [Hash{String => Integer}, Array] subscriber counts keyed by channel
```

nderraugh (approved):

lib/valkey/commands/pubsub_commands.rb
nderraugh: outdated.

lib/valkey/pipeline.rb
```diff
@@ -193,8 +193,6 @@ def call_v(argv)
       subscribe_lazy unsubscribe_lazy psubscribe_lazy punsubscribe_lazy
       ssubscribe_lazy sunsubscribe_lazy
       get_subscriptions get_pubsub_message try_get_pubsub_message
```
jamesx-improving, line 195: Three artifacts disagree about whether `pipelined`/`multi` support is in scope here: CHANGELOG.md:34 and the PR body both advertise it, but the reply to @affonsov says it wasn't intended, and 49f2bfa removed the two pipeline unit tests plus the integration case. Dropping the five verbs from this list is what actually enables it, so whichever way you settle it this line and the CHANGELOG entry should agree — and `get_subscriptions` staying listed is currently unasserted either way.

lib/valkey/commands/pubsub_commands.rb
```diff
@@ -402,19 +397,18 @@ def pubsub_numpat = raise(NotImplementedError, "#{__method__} is not implemented
...
-      # @param [Array<String>] channels the channels to query for the number of subscribers; an empty list
-      #   returns an empty hash
-      # @return [Hash{String => Integer}] the channel names mapped to their number of subscribers
-      # @raise [NotImplementedError] this method is not implemented yet
+      # @param [Array<String>] channels the channels to query; omit to query none
+      # @return [Hash{String => Integer}, Array] subscriber counts keyed by channel
       #
       # @see https://valkey.io/commands/pubsub-numsub/
-      def pubsub_numsub(*channels) = raise(NotImplementedError, "#{__method__} is not implemented yet")
+      def pubsub_numsub(*channels)
```
jamesx-improving, line 404: `flatten_map` looks like a real client-wide constructor option rather than a vector-search-only one: `lib/valkey.rb:366` reads it, `:836` applies it to every map reply, `test/integration/valkey/flatten_map_test.rb` covers it end to end, and CHANGELOG line 30 already ships it as "the `flatten_map: true` compatibility option". Since glide-core converts NUMSUB to a map for both protocols, these two are the only map-returning commands here that pass no normalizing block — `hgetall` passes `&Utils::Hashify`, `config_get` does `Hash[*reply]` — so on a `flatten_map: true` client `pubsub_numsub("a")` hands back `["a", 0]` and indexing it raises `TypeError`. The docstring trim also left the orphaned `;` at line 441.

test/integration/valkey/pubsub_test.rb
```diff
@@ -636,6 +636,95 @@ def test_sharded_connection_config_requires_cluster_mode_in_standalone
...
+
+        assert_includes r.pubsub_channels, channel
+        assert_includes r.pubsub_channels("#{channel}*"), channel
+        assert_numsub({ channel => 1 }, r.pubsub_numsub(channel))
+      end
+    end
+
+    def test_pubsub_numpat_tracks_pattern_subscriptions
+      pattern = "#{unique_channel}*"
+      initial_count = r.pubsub_numpat
```
jamesx-improving, line 676: This assertion is order-dependent rather than vacuous — it would catch wrong metadata, but `pubsub_numpat` is server-global, so `initial_count` is a snapshot of shared state. `test_psubscribe_delivers_message_with_pattern` (line 232) and `test_psubscribe_lazy_eventually_delivers_messages` (line 293) both leave a pattern registered and rely on `with_client`'s `close` to reap it, which isn't ordered against a command on `r`; if one is reaped between this read and line 682, the `+ 1` sees `initial_count` instead.

jamesx-improving (approved):
This closes the `PUBSUB_UNSUPPORTED` pipeline defect rather than activating it, which is the part we'd flagged as latent on #311 and again on #316 — the five verbs now resolve to real implementations, `get_subscriptions` correctly stays deny-listed, and the `validate_resp3!`/`Pipeline#protocol` coupling stays unreachable because none of the five calls it.

Three inline comments below, none blocking. Two things for whoever merges: the Limitations entry pointing at #317 is stale now that #317 has merged, so `test_shard_introspection_shapes` could assert real shard state rather than shapes; and @nderraugh's approval is from 19:28Z, before `6075eed` landed at 20:48Z, so it predates the current head.

test/integration/valkey/pubsub_test.rb
Aryex: parallel test execution is not enabled so this shouldn't bee an issue.

lib/valkey/commands/pubsub_commands.rb
Aryex: Unclear what is the ask here? flatten_map is a hidden options for compatibility only. It is not advertised in our configuration list.

lib/valkey/pipeline.rb
Aryex: added back

PR #325, opened 2026-09-18
Otel improvements for shipping resource attributes

### LINKED ISSUE #323: otel(ruby): Valkey::OpenTelemetry has no way to attach process/host resource attributes
### Describe the feature

`Valkey::OpenTelemetry.init` has no way to attach `process.*` or `host.*` resource attributes to the spans/metrics it exports, and no auto-detection of the process-level attributes that `opentelemetry-sdk`'s `Resource.default` normally provides for free.

Comparing a real trace's `Process` tab for a Valkey-sourced span (`SADD`, via this gem) against a `redis-rb`-sourced span (`SADD`, via `opentelemetry-instrumentation-redis` + `opentelemetry-sdk`) in the same request:

| Resource attribute | redis-rb span | Valkey span |
|---|---|---|
| `telemetry.sdk.name/language/version` | yes | yes |
| `k8s.pod.name`/`k8s.namespace.name`/`k8s.node.name`/`k8s.pod.ip` | yes | yes |
| `service.name` | yes | no (`unknown_service:ruby`, workaroundable via `OTEL_SERVICE_NAME`) |
| `host.ip` / `host.name` | yes | **no** |
| `process.pid` | yes | **no** |
| `process.command` | yes | **no** |
| `process.runtime.name` / `.version` / `.description` | yes | **no** |

**Root cause:** `telemetry.sdk.*` and `k8s.*` match because they come from the Rust core's own SDK defaults plus its reading of the spec-standard `OTEL_RESOURCE_ATTRIBUTES` env var (confirmed via `strings` on `libglide_ffi` — it contains `OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`, `service.name`, `unknown_service:`). Everything else is missing because `lib/valkey/opentelemetry.rb` + `Bindings.init_open_telemetry` is a minimal, command-tracing-only integration with no equivalent of `Resource.default`'s auto-detection (`Process.pid`, `$PROGRAM_NAME`, `RUBY_ENGINE`, `RUBY_VERSION`, `RUBY_DESCRIPTION`) and no merge point for app-supplied attributes like `host.ip`/`host.name` (which aren't even an SDK auto-default — apps supply those explicitly).

### Use Case

Running this gem alongside `opentelemetry-ruby`-instrumented spans (e.g. HTTP, ActiveRecord) in the same request/trace. Today the two sources produce resources with different shapes, which breaks host/process-based filtering, alerting, and cross-referencing in the tracing backend for Valkey command spans specifically.

### Proposed Solution

Two independent pieces, could be shipped separately:

1. Auto-detect process attributes in Ruby (`process.pid` via `Process.pid`, `process.command` via `$PROGRAM_NAME`, `process.runtime.name`/`.version`/`.description` via `RUBY_ENGINE`/`RUBY_VERSION`/`RUBY_DESCRIPTION`) in `lib/valkey/opentelemetry.rb` and pass them through `Bindings.init_open_telemetry`'s config struct. No new Rust dependency — values are computed on the Ruby side.
2. Add a `resource_attributes:` kwarg (a plain `Hash`) to `Valkey::OpenTelemetry.init`, merged into whatever is sent to `Bindings.init_open_telemetry`. Mirrors `OpenTelemetry::SDK::Resources::Resource.create(attrs)`'s shape and lets callers supply `host.ip`/`host.name`/anything app-specific.

Note: `Valkey::OpenTelemetry.init` can only be called once per process — worth checking existing fork-safety notes in that file before deciding where this data gets gathered (e.g. `process.pid` needs to be correct per forked worker, not captured once at parent boot).

### Other Information

`service.name` alone already has an app-side workaround (setting `ENV['OTEL_SERVICE_NAME']` before `.init`, since the native core reads that one env var), but the same underlying gap (no generic resource-attribute API) blocks `host.ip`/`host.name`/`process.*` from any app-side fix. `OTEL_RESOURCE_ATTRIBUTES` could technically carry `process.pid=...` but it's a static env var set once at process start — wrong for anything that legitimately changes per-process (like `process.pid` across forked Puma workers), which is why this needs real gem support.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Client version used

valkey-glide-rb 1.0.0

### Environment details (OS name and version, Ruby version, etc.)

arm64

Sasidharan3094:
### Summary

`Valkey::OpenTelemetry.init` had no way to attach `process.*`/`host.*` resource attributes to
exported spans/metrics, unlike `opentelemetry-ruby`'s `Resource.default`, which auto-detects
`process.pid`/`process.command`/`process.runtime.*`. This showed up as a real gap when comparing
a Valkey-sourced span against a `redis-rb`-sourced span in the same trace: the Valkey span was
missing `host.ip`, `host.name`, and every `process.*` attribute.

Adds a `resource_attributes:` kwarg to `Valkey::OpenTelemetry.init` for caller-supplied attributes
(e.g. `host.ip`, `host.name`), and auto-detects `process.pid`, `process.command`, and
`process.runtime.name`/`.version`/`.description` with no config needed. No native FFI or
`glide-core` change was required: the Rust core's OpenTelemetry SDK already resolves its resource
from the `OTEL_RESOURCE_ATTRIBUTES` env var by default (confirmed - this is why `k8s.*` attributes
injected by a platform sidecar already showed up on Valkey spans before this change), so `init`
merges into that env var for the duration of the single synchronous `init_open_telemetry` FFI call,
then restores it.

### Issue link

Closes [otel(ruby): Valkey::OpenTelemetry has no way to attach process/host resource attributes](https://github.com/valkey-io/valkey-glide-ruby/issues/323)

### Changes

- `lib/valkey/opentelemetry.rb`: added `resource_attributes:` kwarg to `.init`, validated as a
  `Hash`. Auto-detects `process.pid`/`process.command`/`process.runtime.name`/`.version`/
  `.description` via `Process.pid`/`$PROGRAM_NAME`/`RUBY_ENGINE`/`RUBY_VERSION`/`RUBY_DESCRIPTION`.
  Merges auto-detected and caller-supplied attributes into `OTEL_RESOURCE_ATTRIBUTES` (caller >
  auto-detected > whatever was already set) around the `init_open_telemetry` call, restoring the
  prior value afterward even if the call raises.
- Values are sanitized before being placed in the env var: only the comma (the pair separator) is
  replaced, since `opentelemetry-rust`'s `EnvResourceDetector` does no percent-decoding - encoding
  anything else would leak literal `%2F`/`%20` sequences into the resource instead of being decoded
  back.
- `CHANGELOG.md`: added an entry under `## 1.1.0 (Pending)` → `### Changes`.

### Limitations

- `host.ip`/`host.name` still need to be supplied explicitly via `resource_attributes:` - they
  aren't auto-detectable (this matches `opentelemetry-ruby`, which doesn't auto-detect them either;
  they're app-supplied there too).
- A value containing a comma has no way to round-trip correctly through `OTEL_RESOURCE_ATTRIBUTES`
  in the current `opentelemetry-rust` (no percent-decoding support), so commas in a value are
  replaced with `_` rather than preserved.

### Testing

- Added `test/unit/opentelemetry_resource_attributes_test.rb` (12 tests): auto-detection, merge
  precedence (caller > auto-detected > pre-existing env), comma sanitization, env var
  set/restore around the FFI call (including when the block raises), and an end-to-end
  `init(resource_attributes:)` test with `Bindings.init_open_telemetry` stubbed.
- Ran `bundle exec rubocop` — no offenses.
- Ran `bundle exec rake test:unit` — all passing, 0 failures.
- Manually validated end-to-end against a real local Valkey server (via `cluster_manager.py`) using
  the file exporter: confirmed `host.ip`, `host.name`, `process.pid`, `process.command`, and
  `process.runtime.name`/`.version`/`.description` all appear correctly in the exported resource
  after a real `SET`/`GET`. This also caught and fixed a percent-encoding bug (see Changes) before
  it shipped.

### Checklist


Sasidharan3094:
@Aryex @jamesx-improving Kindly review this PR.

lib/valkey/opentelemetry.rb
```diff
@@ -203,6 +221,46 @@ def reset!
...
+
+      # Later entries win on key collision, so caller-supplied attributes override auto-detected
+      # ones, which override whatever OTEL_RESOURCE_ATTRIBUTES already held (e.g. platform-injected
+      # k8s.* attributes).
+      def build_resource_attributes_env(resource_attributes)
+        attrs = auto_detected_resource_attributes.merge(resource_attributes || {})
+        pairs = attrs.map { |k, v| "#{sanitize_otel_resource_component(k)}=#{sanitize_otel_resource_component(v)}" }
+
+        existing = ENV.fetch("OTEL_RESOURCE_ATTRIBUTES", "")
+        (existing.empty? ? pairs : [existing] + pairs).join(",")
```
jamesx-improving: Auto-detected `process.*` land after `existing` here, and later entries win in `opentelemetry-rust`, so they override an operator-set `OTEL_RESOURCE_ATTRIBUTES=process.command=checkout-worker`. `opentelemetry-ruby` goes the other way — `Resource.default` is `.merge(process).merge(telemetry_sdk)` with the env pairs folded into `telemetry_sdk`, so there the env var beats auto-detection. Weakest-first (auto, existing, caller) would preserve the "caller wins" you documented without taking operator config with it.

test/unit/opentelemetry_resource_attributes_test.rb
```diff
@@ -0,0 +1,146 @@
...
+    original_config = ::Valkey::OpenTelemetry.config
+    ::Valkey::OpenTelemetry.instance_variable_set(:@initialized, false)
+
+    seen_during_ffi_call = nil
+    stub_init = lambda do |_config|
+      seen_during_ffi_call = ENV.fetch("OTEL_RESOURCE_ATTRIBUTES", nil)
+      FFI::Pointer::NULL
+    end
+
+    Valkey::Bindings.stub(:init_open_telemetry, stub_init) do
```
jamesx-improving, line 136: None of the 12 tests would fail if the attributes stopped reaching the core — they stub the FFI call and assert only the env var's content, while the effect depends on glide-core not calling `.with_resource()` (true at pin 49f2b6d, unpinned here). A metrics-file assertion on the exported `"resource"` object would surface an upstream regression here rather than in a production trace.

jamesx-improving (approved):
Mechanism checks out end to end: glide-core never calls `.with_resource()`, `opentelemetry_sdk` 0.32's `Resource::builder()` runs `EnvResourceDetector` last and doesn't memoise it, and both providers are built synchronously inside the FFI call — so the attributes really do reach the exporter.

One correction on the validation: the file exporter can only have shown this for metrics. `span_exporter_file.rs` keeps the resource handed to `set_resource`, but `to_jsons()` never serialises it; only `metrics_exporter_file.rs` writes a `"resource"` object.

Two inline comments below, both non-blocking.

lib/valkey/opentelemetry.rb
Sasidharan3094: good catch, you're right. fixed — now merging as auto_detected.merge(existing_env).merge(caller) instead of appending existing before the auto+caller pairs, so an operator-set OTEL_RESOURCE_ATTRIBUTES value like process.command=checkout-worker won't get clobbered by our own auto-detection anymore. added test_existing_otel_resource_attributes_wins_over_auto_detected_on_key_collision + a malformed-entry test to lock it in. pushed.

test/unit/opentelemetry_resource_attributes_test.rb
Sasidharan3094: fair — stubbing the FFI call only proves our env var construction, not that it actually reaches the exporter. added test_auto_detected_resource_attributes_appear_in_exported_metrics to the integration suite: real OTel init, real file exporter, real command, asserts against the metrics file's actual "resource" object. ran it against a live server, passes. that'll catch it if glide-core ever starts calling .with_resource() explicitly.

lib/valkey/opentelemetry.rb
```diff
@@ -203,6 +221,54 @@ def reset!
...
+        end
+      end
+
+      # Weakest-first: auto-detected attributes must not clobber an operator-set
+      # OTEL_RESOURCE_ATTRIBUTES, and resource_attributes: must win over both.
+      def build_resource_attributes_env(resource_attributes)
+        existing = parse_otel_resource_attributes(ENV.fetch("OTEL_RESOURCE_ATTRIBUTES", ""))
+        attrs = auto_detected_resource_attributes.merge(existing).merge(resource_attributes || {})
+
+        attrs.map { |k, v| "#{sanitize_otel_resource_component(k)}=#{sanitize_otel_resource_component(v)}" }.join(",")
```
Aryex: `resource_attributes silently changes keys that contain `=`.

The `init` YARD docs accept arbitrary resource attributes and document only non-Hash rejection. The OpenTelemetry Resource SDK specification says `,` and `=` in keys and values must be percent-encoded.

Reproduced at this PR head with Ruby 4.0.5 and `opentelemetry_sdk` 0.32.1:

```text
Input:    {"deployment=region" => "us-west"}
Expected: {"deployment=region" => "us-west"} or ArgumentError
Actual:   {"deployment" => "region=us-west"}
```

test/integration/valkey/opentelemetry_test.rb
```diff
@@ -390,5 +410,24 @@ def test_batch_span_is_child_of_parent_context_when_provider_registered
       assert_equal trace_id, span["trace_id"]
       assert_equal span_id, span["parent_span_id"]
     end
+
+    # Test 11: auto-detected process.* attributes reach the exported metrics resource -
+    # exercising the real glide-core OpenTelemetry SDK, not a stubbed FFI call.
+    def test_auto_detected_resource_attributes_appear_in_exported_metrics
```
Aryex: `test_auto_detected_resource_attributes_appear_in_exported_metrics` can pass while caller attributes, pre-existing attributes, or their precedence fail at the native boundary.

The `resource_attributes` contract says caller values are merged with `OTEL_RESOURCE_ATTRIBUTES` and auto-detected values, with caller values taking precedence.

Verified at this PR head:

- The real native test initializes without `resource_attributes` or a controlled `OTEL_RESOURCE_ATTRIBUTES` value and asserts only `process.*`.
- The test that passes `host.name` and `host.ip` replaces `Bindings.init_open_telemetry` with a lambda, so it never checks the exported resource.

test/unit/opentelemetry_resource_attributes_test.rb
```diff
@@ -0,0 +1,164 @@
...
+
+  def test_build_resource_attributes_env_with_no_existing_env_var
+    ENV.delete("OTEL_RESOURCE_ATTRIBUTES")
+
+    value = ::Valkey::OpenTelemetry.send(:build_resource_attributes_env, nil)
+
+    refute_match(/\A,/, value)
+  end
+
+  def test_commas_are_replaced_since_they_break_the_list_format
```
Aryex: The sanitization tests exercise delimiters only in values, so the key path can remain broken while the suite passes.

The OpenTelemetry Resource SDK specification applies its `,` and `=` delimiter rules to keys and values. `build_resource_attributes_env` also processes both components.

At this PR head, the tests cover `{"custom.attr" => "a,b"}` and a value containing `=`. They never pass `custom,key` or `custom=key`. The Ruby 4.0.5 and `opentelemetry_sdk` 0.32.1 OTLP reproduction exported `custom=key => value` as `custom => key=value`.

test/unit/opentelemetry_resource_attributes_test.rb
```diff
@@ -0,0 +1,164 @@
...
+  end
+
+  def test_caller_supplied_attributes_win_over_auto_detected_on_key_collision
+    value = ::Valkey::OpenTelemetry.send(:build_resource_attributes_env, { "process.pid" => "overridden" })
+
+    assert_match(/process\.pid=overridden/, value)
+    refute_match(/process\.pid=#{Process.pid}\b/, value)
+  end
+
+  def test_existing_otel_resource_attributes_is_preserved
```
Aryex, line 47: No tests lock in three `parse_otel_resource_attributes` rules: trimming whitespace, keeping the last duplicate key, and preserving additional `=` characters in values.

This parser handles operator-provided `OTEL_RESOURCE_ATTRIBUTES` before passing them to the native SDK. Its comment says it matches the pinned Rust parser.

Verified at this PR head:

```text
Input:  "spaced.key = spaced value,dup=first,dup=last,padded=value=="
Output: {"spaced.key"=>"spaced value", "dup"=>"last", "padded"=>"value=="}
```

Current tests cover only simple unique pairs and one malformed token without `=`.

lib/valkey/opentelemetry.rb
```diff
@@ -72,12 +78,20 @@ class << self
       #   the current application span context (see {set_parent_span_context_provider}). Equivalent to
       #   calling {set_parent_span_context_provider} separately; provided here for convenience.
       #
+      # @param resource_attributes [Hash, nil] Extra OpenTelemetry resource attributes (e.g.
```
Aryex, line 86: Should the `file://` trace example say that its JSON omits resource attributes?

The new YARD text says these attributes attach to every span and metric. The endpoint docs list the trace file exporter without naming a resource-output limitation.

Reproduced at this PR head with Ruby 4.0.5 and the pinned native exporter using `opentelemetry_sdk` 0.32.1:

```text
Input:    file:// trace exporter with {"host.name" => "web-1"}
Expected: trace output contains host.name, or docs name the limitation
Actual:   no resource field, host.name, or web-1 in the trace record
```

lib/valkey/opentelemetry.rb
```diff
@@ -203,6 +221,54 @@ def reset!
...
+          "process.command" => $PROGRAM_NAME,
+          "process.runtime.name" => RUBY_ENGINE,
+          "process.runtime.version" => RUBY_VERSION,
+          "process.runtime.description" => RUBY_DESCRIPTION
+        }
+      end
+
+      # opentelemetry-rust does no percent-decoding, so only the comma (the pair separator) needs
+      # escaping - anything else would leak a literal %XX into the value.
+      def sanitize_otel_resource_component(value)
```
Aryex: Should the `resource_attributes` docs say that commas in keys and values become underscores?

The YARD text and changelog describe arbitrary resource attributes. The OpenTelemetry Resource SDK specification requires commas in environment-based keys and values to be percent-encoded, while this transport performs a lossy replacement.

Reproduced at this PR head with Ruby 4.0.5 and `opentelemetry_sdk` 0.32.1:

```text
Input:    {"region,name" => "us,west"}
Expected: {"region,name" => "us,west"} under the current docs
Actual:   {"region_name" => "us_west"}
```

lib/valkey/opentelemetry.rb
Sasidharan3094: good catch, that one's a real bug not just cosmetic. fixed — split the sanitizer into sanitize_otel_resource_key (replaces , and = with _) and sanitize_otel_resource_value (only ,, since a value's = is safe — only the first = in an entry is ever the separator). {"deployment=region" => "us-west"} now round-trips to {"deployment_region" => "us-west"} instead of {"deployment" => "region=us-west"}.

test/unit/opentelemetry_resource_attributes_test.rb
Sasidharan3094: yeah, fair gap. added test_commas_in_keys_are_replaced_since_they_break_the_list_format and test_equals_signs_in_keys_are_replaced_but_preserved_in_values, plus split the old combined test into key/value versions so the key path isn't only exercised incidentally through value cases anymore.

test/unit/opentelemetry_resource_attributes_test.rb
Sasidharan3094: added test_parse_otel_resource_attributes_trims_whitespace_keeps_last_duplicate_and_extra_equals using your exact repro string ("spaced.key = spaced value,dup=first,dup=last,padded=value==") and asserting the exact hash you verified by hand. locks in all three behaviors in one place.

lib/valkey/opentelemetry.rb
Sasidharan3094: good point, that's a real doc gap. added a note next to the resource_attributes example that the file:// trace exporter doesn't emit resource attributes in its output (only the file:// metrics exporter does) — it's a glide-core limitation on that specific exporter, not something resource_attributes: is doing wrong, and HTTP/gRPC trace exporters aren't affected.

lib/valkey/opentelemetry.rb
Sasidharan3094: agreed, added it to the resource_attributes YARD doc — a , in a key or value, or a = in a key, gets replaced with _ rather than percent-encoded, since the pinned opentelemetry-rust doesn't percent-decode OTEL_RESOURCE_ATTRIBUTES. also noted = inside a value is preserved.

Aryex (approved):

Aryex (approved):

PR #329, opened 2026-09-21
feat(ruby): enable Pub/Sub commands in pipeline and multi

Aryex:
### Summary

Enables the Python sync-supported Pub/Sub command surface inside Ruby `pipelined` and block-form `multi`, with topology-aware handling for sharded Pub/Sub commands.

### Issue link

Related to #135.

### Changes

- Pass client topology into `Valkey::Pipeline`.
- Enable `publish`, `pubsub_channels`, `pubsub_numpat`, and `pubsub_numsub` in pipelines and transactions.
- Enable sharded publish and shard metadata commands for cluster pipelines and transactions while rejecting them in standalone batches.
- Keep subscription mutation, subscription-state access, and message reads unsupported in batches.
- Port Python sync coverage for pipeline/multi, RESP2/RESP3, live delivery, empty and populated metadata replies, futures, unsupported operations, and topology restrictions.

### Limitations

- Subscribe/unsubscribe variants, `get_subscriptions`, `get_pubsub_message`, and `try_get_pubsub_message` remain unsupported inside `pipelined`/`multi`.
- Sharded Pub/Sub batch operations require cluster mode.

### Testing

- `bundle exec rubocop` — 126 files inspected, no offenses.
- `bundle exec rake test:unit` — 442 tests, 1,121 assertions, 0 failures, 0 errors.
- `SKIP_TLS_TESTS=true bundle exec rake test:standalone` — 868 tests, 4,475 assertions, 0 failures, 0 errors.
- `SKIP_TLS_TESTS=true bundle exec rake test:cluster` — 828 tests, 3,224 assertions, 0 failures, 0 errors.

### Checklist


yipin-chen (approved):
LGTM

lib/valkey/pipeline.rb
```diff
@@ -200,5 +218,17 @@ def call_v(argv)
...
       end
     end
+
+    private
+
+    def cluster_mode?
+      @cluster_mode
+    end
+
+    def validate_cluster_pubsub_batch_command!(command)
```
affonsov: This duplicates validate_cluster_mode! from PubSubCommands (same body, same "is only available in cluster mode." message), which Pipeline already inherits via include Commands, and it resolves against the cluster_mode? you define just above. You can drop this private method and have pubsub_shardchannels / pubsub_shardnumsub call validate_cluster_mode!(__method__) directly, keeping one source of truth for the check and its wording.

test/integration/valkey/pubsub_test.rb
```diff
@@ -637,6 +637,12 @@ def test_sharded_connection_config_requires_cluster_mode_in_standalone
     end
 
     SUBSCRIPTION_MODE_KEYS = %i[exact pattern sharded].freeze
+    NON_BATCHABLE_PUBSUB_COMMANDS = %i[
```
affonsov: This constant is defined but never referenced anywhere in the file. Either wire it into an integration assertion (loop over it asserting each raises ArgumentError inside pipelined / multi, mirroring the unit test) or remove it. As written it is dead code that reads like a forgotten test.

test/unit/pipeline_test.rb
```diff
@@ -42,10 +42,17 @@ def test_abort_futures_marks_only_unresolved_futures
     assert_raises(Valkey::FutureAborted) { future_b.value }
   end
 
-  def test_pubsub_commands_raise_argument_error
+  def test_pipeline_not_supported_pubsub
     pipeline = Valkey::Pipeline.new
 
-    Valkey::Pipeline::PUBSUB_UNSUPPORTED.each do |name|
+    not_supported = %i[
```
affonsov: This used to iterate Valkey::Pipeline::PUBSUB_UNSUPPORTED; it now hardcodes an identical literal list. That same list now lives in three places (the PUBSUB_UNSUPPORTED constant, this test, and NON_BATCHABLE_PUBSUB_COMMANDS in the integration test). Iterating the constant here keeps the test honest: if someone changes PUBSUB_UNSUPPORTED, the assertion follows automatically instead of drifting.

test/integration/valkey/pubsub_test.rb
```diff
@@ -733,6 +778,93 @@ def skip_unless_sharded_pubsub
...
+    end
+
+    def assert_pubsub_batch_commands(batch_method)
+      %i[resp2 resp3].each do |protocol|
+        assert_pubsub_batch_commands_for_protocol(batch_method, protocol)
+      end
+    end
+
+    def assert_pubsub_batch_commands_for_protocol(batch_method, protocol)
+      channel = unique_channel("{#{batch_method}-#{protocol}}")
```
affonsov: All channels here are hash-tagged into a single slot ("{#{batch_method}-#{protocol}}"), so an atomic multi on cluster is only ever exercised within one slot. The Java suite explicitly covers the multi-slot atomic case as a CROSSSLOT error path, and documents that a multi-channel publish transaction across slots is broken on cluster. Consider either adding a cluster-only test asserting the cross-slot atomic multi failure, or a one-line comment noting that single-slot hash-tagging is deliberate, so a future reader knows the multi-slot path was considered.

lib/valkey/pipeline.rb
```diff
@@ -186,8 +187,25 @@ def call_v(argv)
     end
     # rubocop:enable Lint/UselessMethodDefinition
 
-    # Subscriptions outlive a batch and pushes arrive out of band, so neither
-    # can be expressed as one queued reply.
+    # Python sync exposes sharded Pub/Sub operations only on ClusterBatch.
+    # Preserve that topology split even though the shared Ruby command module
+    # also permits direct sharded calls against a standalone server.
+    def publish(message, channel, sharded: false)
+      raise ArgumentError, "publish with sharded: true is only available in cluster mode." if sharded && !cluster_mode?
```
affonsov: Minor wording consistency: this raises "publish with sharded: true is only available in cluster mode." while the two shard-introspection guards raise "<command> is only available in cluster mode.". If you consolidate onto the inherited validate_cluster_mode! (comment above), decide whether publish should share that phrasing or keep the more specific one. Tests only match /cluster mode/, so either works functionally.

test/integration/valkey/pubsub_test.rb
affonsov: Since this is the last Pub/Sub PR for Ruby, I want to raise a suite-level coverage question before we call the feature done, and hear how you'd like to handle it. The core clients (Java, Python, Node, Go) all build Pub/Sub coverage from a parametrized scenario matrix: SubscriptionMethod {Config (static), Lazy, Blocking} x MessageReadMethod {Callback, Async, Sync} x topology x channel-mode, and they cover combined modes on a single connection (exact + pattern [+ sharded], and the same channel name across modes). Java's getTestScenarios() alone is 18 combinations applied to every happy-path test.

The Ruby suite covers each of these axes, but in isolated ad-hoc tests rather than as a cross-product, and it has no combined-mode coverage at all. Concretely, relative to core we are missing:
1. Combined modes on one connection (exact + pattern, exact + pattern + sharded, and overlapping channel names across modes). This is entirely absent.
2. A parametrized static-vs-dynamic (Config/Lazy/Blocking) crossed with read-method (Callback/Async/Sync), so an interaction like Blocking-subscribe plus Callback-delivery on cluster is actually exercised.
3. A many-channels scale test (core uses 256; Ruby tops out at 2 or 3).
4. A sync + async read coexistence test on one client.

Thanks to this combination I found some issues when we were implementing the node dynamic pubsub.

I mentioned this in some other PR that I don't remember XD (308/#316/#317/#320), so how you want to handle this? in this PR or do in a follow-up issue?

Aryex: This is a fair concern. I will raise  a separate PR for this and keep #135 open until then.

lib/valkey/pipeline.rb
Aryex: Should be addressed in [81d39e8](https://github.com/valkey-io/valkey-glide-ruby/pull/329/commits/81d39e8417cc1ff60cbbac0939f5c37575b71a94). Comments updated

test/integration/valkey/pubsub_test.rb
Aryex: Added a new test covering cross slot error in [81d39e8](https://github.com/valkey-io/valkey-glide-ruby/pull/329/commits/81d39e8417cc1ff60cbbac0939f5c37575b71a94)

test/unit/pipeline_test.rb
Aryex: updated in [81d39e8](https://github.com/valkey-io/valkey-glide-ruby/pull/329/commits/81d39e8417cc1ff60cbbac0939f5c37575b71a94)

test/integration/valkey/pubsub_test.rb
Aryex: removed in [81d39e8](https://github.com/valkey-io/valkey-glide-ruby/pull/329/commits/81d39e8417cc1ff60cbbac0939f5c37575b71a94)

lib/valkey/pipeline.rb
Aryex: removed in [81d39e8](https://github.com/valkey-io/valkey-glide-ruby/pull/329/commits/81d39e8417cc1ff60cbbac0939f5c37575b71a94)

affonsov (approved):

